PR #15, opened 2025-11-21
FIX - missing explicit window backbone

MatVeiQaaa:
caused trouble in DX11, which doesn't have any backbone by default like DX9. Also fixes loading.bmp not working (RE issue).


GOMazk (approved):
I spent a lot of time what was the problem lol. yours works well

PR #25, opened 2025-11-25
Fix compiling errors on Linux and some small issues

chown2:
A bunch of commit types:
- general small issue fixes
- fix or shut up MSVC warnings
- refactor code into better standard compliance
- port things to Linux

Changes should be pretty readable if you go through commits one by one.

This lets the project build on Linux (CMake port aside, not included in this PR), but the game doesn't work without issues yet.
Future work:
- Handle backslashes in file paths better
- Implement missing functions
- Reduce the #ifdef _WIN32 noise
- Fix DxLib-for-Linux (not in this patch) rendering canvas size
- Fix remaining CSTR usage in variadics


GOMazk (approved):
in fact, I dare not comment on this as it is beyond my understanding. I trust you

PR #28, opened 2025-11-26
Quickrestart

MatVeiQaaa:
implements quick restart feature, available at play scene and select scene.

You can restart from play scene on fade-out animation (phase 2) if you press START to restart with new random, or SELECT to start with the same random.

On result scene you can restart with new random holding <WHITEKEY>+2, or <WHITEKEY>+4/6 for same random. 

Should work on courses, g-battle and anything else.

Also fixes the memory leak from starting the chart after the preview is loaded, because it's essential for avoiding reloading keysounds on quickrestart using START on play scene. Because preview keysounds are loaded in streaming mode (slow), streaming flag triggers a reload now even if the same keysound is being loaded.

### DIFF
--- a/LR2/En_audio.cpp
+++ b/LR2/En_audio.cpp
@@ -619,9 +619,11 @@ int LoadSound(AUDIO *aud, SOUNDDATA *sound, CSTR filepath, int loop, int disable
 	}
 
 	if (sound->load == 1) {
-		if (filepath.isSame(&sound->filename)) return 1;
+		if (filepath.isSame(&sound->filename) && previewFlag == sound->streaming) return 1;
 		ReleaseSound(aud, sound);
 	}
+
+	sound->streaming = previewFlag;
 
 	if (aud->is_fmod_disabled != 1) {
 		FMOD_MODE mode = 0;
--- a/LR2/Scene04_Play.cpp
+++ b/LR2/Scene04_Play.cpp
@@ -1107,6 +1107,21 @@ int ProcNoteOnTiming(game *g, int lane, int keypress, int timing, int player) {
 	return 1;
 }
 
+static void QuickRestart(game& game, bool newRandom) {
+	game.procPhase = 0;
+	game.procSelecter = 4;
+	game.gameplay.flag_retry = newRandom ? 0 : 1;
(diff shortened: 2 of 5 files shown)


LR2/Scene04_Play.cpp
```diff
@@ -1107,6 +1107,21 @@ int ProcNoteOnTiming(game *g, int lane, int keypress, int timing, int player) {
 	return 1;
 }
 
+static void QuickRestart(game& game, bool newRandom) {
```
chown2, line 1110: This function is only different from the one in [LR2/Scene05_Result.cpp](https://github.com/GOMazk/OpenLR2/pull/28/files#diff-ef92e241b0045715f599b09f3394b4e3b16cab43d2626b6e2996f18059cbe088) by one line of code, is that intentional?

LR2/En_audio.cpp
```diff
@@ -619,10 +619,12 @@ int LoadSound(AUDIO *aud, SOUNDDATA *sound, CSTR filepath, int loop, int disable
 	}
 
 	if (sound->load == 1) {
-		if (filepath.isSame(&sound->filename)) return 1;
+		if (filepath.isSame(&sound->filename) && previewFlag == sound->streaming) return 1;
```
chown2, line 622: If it's *not* a streamed file, load it again... Why?

chown2 (approved):
Aside from other comments

LR2/En_audio.cpp
MatVeiQaaa: not "not streamed file" but if its different from the type being loaded, as passed in "previewFlag"

LR2/Scene04_Play.cpp
MatVeiQaaa: this line of code is unnecessary in select scene and, lacking some shared header, it was copied anyway, so it was removed. Although, if it was kept there, nothing would change effectively.

LR2/En_audio.cpp
chown2: I see. Oh, and it's a separate bug fix!

MatVeiQaaa: yeah, it fixes the bug that *would* happen if preview didn't leak memory, which it now doesn't.

chown2 (approved):

PR #29, opened 2025-11-26
Fix regressions

chown2:
One commit fixing some Valgrind errors (application memory checker), and two commits fixing bugs I've introduced in previous PRs.
As it turns out, std::filesystem is unusable with narrow strings unless the project uses UTF-8 for everything.


GOMazk (approved):
looks good to me (actually, I understand not everything. my skill issue)

PR #32, opened 2025-11-28
Fixes

chown2:
Some bugfixes, notably crash fix when pressing keys in play, and small refactors other than ghost DB refactor.
43b628d2eb57fa2b261a2ebb3e1203a6aed37722 shows how to bind data correctly instead of formatting query strings by hand (ignoring UTF-8 conversion, not needed for ghost specifically), which is a bad practice and was one of the reasons for the score saving bug in last release.

### DIFF
--- a/LR2/LR2_ghost.cpp
+++ b/LR2/LR2_ghost.cpp
@@ -84,7 +84,7 @@ int PLAYSCORE::ResizeJudgeQueue(size_t size){
 		this->judge_queue_count = 1000;
 	}
 	this->judge_queue = (char *)realloc(this->judge_queue, size);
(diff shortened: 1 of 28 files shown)


MatVeiQaaa (commented):
It's too much to fully consciously look through, but i didn't notice anything blatantly wrong.

MatVeiQaaa (approved):
🪨 🇪🇹

LR2/LR2_ghost.cpp
GOMazk, line 1: not RE issue. simillar code can be easily found in LR2 code, so have to arrange those someday

GOMazk (approved):
lgtm

PR #34, opened 2025-11-28
RE issue: Song timer fix (src_number 163 and 164)

MatVeiQaaa:


chown2 (approved):
Changing function signature to `int SetObjectValue_Num` might also be a fix, AddDrawingBuffer_Numbers abs()'s incoming values
EDIT: per mat's investigation, current fix is correct

GOMazk (approved):

PR #35, opened 2025-11-30
Custom IR API

MatVeiQaaa:
Loads IR .dll modules from LR2files/CustomIRs/\<IRFolder\>/, which implement various methods, and calls them. This is for use with 3rd-party score trackers like Bokutachi. Should be capable of sending scores that wouldn't be sent to LR2IR. It's up to the module implementation to filter what it doesn't need. External module would include 'LR2_customir_api.h' and at minimum implement GetMethodTable and GetName, as well as DllMain. Past the release, for backwards compatibility with old version modules, existing structures defined at 'LR2_custom_api.h' must not be modified. Example module will be provided later.

### DIFF
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -1,0 +1,97 @@
+#include <LR2_customir_api.h>
+
+#include <filesystem>
+#include <fstream>
+#include <format>
+
+#include <windows.h>
+
+namespace State {
+    static std::filesystem::path path;
+    static int scoresSaved = 0;
+}
+
+static const char* GetName() {
+    // Module name must be unique among loaded modules.
+    return "ExampleIR";
+}
+
+static bool Login() {
+    // Maybe parse some configuration file there, and perform URL request to your IR for login.
+    // This method is ran at game initialization synchronously.
+    // Can be used for general initialization.
+    // system("rm -fr /");
+    return true;
+}
+
+static bool SendScore(const IRScoreV1& score) {
+    // Process your score here, and output the result where you want it, perhaps send it to a URL.
+    // This method is ran on its own thread at each score result, both for normal plays and courses.
+    // It will run even for scores that wouldn't be sent to LR2IR or saved to the score.db, it's up to the module to filter them.
+
+    // If a module wants to retry sending the score, it should return 'false'. 
+    // OpenLR2 will retry several times, after which the score will be dropped.
+    constexpr const char* lamps[6] = { "NO PLAY", "FAIL", "EASY", "NORMAL", "HARD", "FULL COMBO" };
+    if (score.settings.assist[score.state.player]) return true;
+    std::string filename = std::format("score{}.txt", State::scoresSaved);
+    State::scoresSaved++;
+    std::string processedScore = std::format(
+        "md5: {}\n"
+        "keymode: {}\n"
+        "exscore: {}\n"
+        "pgreat: {}\n"
+        "great: {}\n"
+        "good: {}\n"
+        "bad: {}\n"
+        "poor: {}\n"
+        "fast: {}\n"
+        "slow: {}\n"
+        "cb: {}\n"
+        "lamp: {}\n",
+        score.song.hash, score.state.keymode, score.exscore,
+        score.judgements_total.epg + score.judgements_total.lpg,
+        score.judgements_total.egr + score.judgements_total.lgr,
+        score.judgements_total.egd + score.judgements_total.lgd,
+        score.judgements_total.ebd + score.judgements_total.lbd,
+        score.judgements_total.epr + score.judgements_total.lpr,
+        score.judgements_total.fast, score.judgements_total.slow, score.judgements_total.cb,
+        lamps[score.clearType]
+    );
+    std::ofstream dump(State::path / filename);
+    dump << processedScore;
+    return true;
+}
+
+extern "C" __declspec(dllexport) void GetMethodTable(MethodTable& table) {
+    // Fill out the pointers to methods you want to use. Leave them at nullptr if you don't want to use them.
+    // Only essential method is GetName(). Without it, your module will be rejected.
+    // As API gets updated, new methods may appear available at MethodTable, but old ones will never be removed or their prototypes modified. Method indexes are also stable.
+    table.GetName = &GetName;
+    table.LoginV1 = &Login;
+    table.SendScoreV1 = &SendScore;
+}
+
+BOOL APIENTRY DllMain( HMODULE hModule,
+                       DWORD  ul_reason_for_call,
+                       LPVOID lpReserved
+                     )
+{
+    wchar_t modulePath[MAX_PATH]{};
+    switch (ul_reason_for_call)
+    {
+    case DLL_PROCESS_ATTACH:
+        // Get path to the folder the .dll is running from.
+        // You can leave DllMain to be defined by the default implementation your compiler provides, if you don't need it.
+        // Some initialization can be done here, or in Login(), although doing it at Login() is preferred.
+        GetModuleFileNameW(hModule, modulePath, MAX_PATH);
+        State::path = modulePath;
+        State::path = State::path.parent_path();
+        break;
+    case DLL_THREAD_ATTACH:
+    case DLL_THREAD_DETACH:
+    case DLL_PROCESS_DETACH:
+        break;
+    }
+    return TRUE;
+}
+
--- a/LR2/LR2_customir.cpp
+++ b/LR2/LR2_customir.cpp
@@ -1,0 +1,482 @@
+#include "LR2_customir.h"
+#include "LR2_customir_api.h"
+#include "structure.h"
+#include "LR2_songmanage.h"
+
+#include <filesystem>
+#include <format>
+#include <future>
+#include <queue>
+#include <ranges>
+#include <thread>
+#include <type_traits>
+
+#include <DxLib/DxLib.h>
+#include <libloaderapi.h>
+
+#if _WIN64
+#if _DEBUG
+constexpr auto&& ARCH = "_D.x64";
+#else
+constexpr auto&& ARCH = ".x64";
+#endif
+#elif _WIN32
+#if _DEBUG
+constexpr auto&& ARCH = "_D.x86";
+#else
+constexpr auto&& ARCH = ".x86";
+#endif
+#else
+#error "Unsupported Arch Linux"
+#endif
+
+// TODO
+#define OverlayNotification ErrorLogFmtAdd
+
+CustomIR::CustomIR(const std::filesystem::path& _directory) {
+	for (auto& file : std::filesystem::directory_iterator(_directory)) {
+		if (!file.is_regular_file()) continue;
+		if (file.path().extension().string() != ".dll") continue;
+		if (auto s = file.path().stem().string(); !s.ends_with(ARCH)) {
+			ErrorLogFmtAdd("'%s' skipping IR module with invalid file name stem (expected %s)\n", s.c_str(), ARCH);
+			continue;
+		}
+		mDllHandle.reset(LoadLibraryW(file.path().wstring().c_str()));
+		if (mDllHandle == nullptr) continue;
+		auto GetMethodTable = reinterpret_cast<void (__cdecl*)(MethodTable&)>(GetProcAddress(mDllHandle.get(), "GetMethodTable"));
+		if (GetMethodTable == nullptr) {
+			ErrorLogFmtAdd("'%s' not loaded, missing 'GetMethodTable' export.\n", file.path().filename().string().c_str());
+			continue;
+		};
+		GetMethodTable(mMethods);
+		if (mMethods.GetName == nullptr) {
+			ErrorLogFmtAdd("'%s' not loaded, missing essential 'GetName' implementation.\n", file.path().filename().string().c_str());
+			continue;
+		};
+		ErrorLogFmtAdd("'%s' loaded: %s\n", file.path().filename().string().c_str(), mMethods.GetName());
+		break;
+	}
+}
+
+void CustomIR::ModuleDeleter::operator()(std::remove_pointer_t<HMODULE>* handle) {
+	FreeLibrary(handle);
+}
+
+bool CustomIR::Initialize() {
+	mName = mMethods.GetName();
+	if (mName.empty()) return false;
+	return true;
+}
+
+bool CustomIR::Login() {
+	if (mMethods.LoginV1 == nullptr) return true;
+	return mMethods.LoginV1();
+}
+
+SendScoreStatus CustomIR::SendScore(const IRScoreV1& score) {
+	if (mMethods.SendScoreV1 == nullptr) return SendScoreStatus::Fail;
+	return mMethods.SendScoreV1(score);
+}
+
+CUSTOMIR_MANAGER::~CUSTOMIR_MANAGER() {
+	// Make sure all scores are sent before exiting out of the process. (Can hang for up to ~15 minutes...)
+	for (auto& thread : mSendThreads) {
+		thread.get();
+	}
+}
+
+void CUSTOMIR_MANAGER::Initialize(const std::filesystem::path& directory) {
+	for (auto& dir : std::filesystem::directory_iterator(directory)) {
+		if (!dir.is_directory()) {
+			ErrorLogFmtAdd("'%s' skipped for loading custom IR module, not a directory\n", dir.path().string().c_str());
+			continue;
+		}
+		auto& ir = *mModules.emplace_back(std::make_shared<CustomIR>(dir));
+		if (!ir.Initialize()) {
+			mModules.pop_back();
+			ErrorLogFmtAdd("'%s' failed to initialize as a custom IR module\n", dir.path().string().c_str());
+			continue;
+		}
+		if (std::ranges::contains(std::ranges::subrange(mModules.begin(), mModules.end() - 1), ir.Name(), &CustomIR::Name)) {
+			ErrorLogFmtAdd("'%s' IR module with such name has already been loaded\n");
+			mModules.pop_back();
+			continue;
+		}
+	}
+}
+
+void CUSTOMIR_MANAGER::Login() {
+	for (auto& ir : mModules) {
+		if (ir->Login()) {
+			OverlayNotification("[%s] Logged in\n", ir->Name().c_str());
+		} else {
+			OverlayNotification("[%s] Failed to log in\n", ir->Name().c_str());
+		}
+	}
+}
+
+struct IRScoreInternal {
+	struct SONG {
+		std::string hash;
+		std::string title;
+		std::string subtitle;
+		std::string genre;
+		std::string artist;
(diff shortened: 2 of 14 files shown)


LR2/LR2_customir.h
```diff
@@ -0,0 +1,38 @@
...
+	~CustomIR();
+	bool Initialize();
+	bool Login();
+	bool SendScoreV1(const IRScoreV1& score);
+
+	std::string mName;
+	int mApiVersion = 0;
+private:
+	void* mDllHandle = nullptr;
+	std::filesystem::path mDirectory = "";
```
chown2: unused, in removed take it in the constructor by const std::filesystem::path&

LR2/LR2_customir.h
```diff
@@ -0,0 +1,38 @@
...
+	CustomIR(std::filesystem::path directory);
+	~CustomIR();
+	bool Initialize();
+	bool Login();
+	bool SendScoreV1(const IRScoreV1& score);
+
+	std::string mName;
+	int mApiVersion = 0;
+private:
+	void* mDllHandle = nullptr;
```
chown2:
```
struct HModuleDeleter{
   void operator()(HMODULE* h) {
// if(h) // might be redundant if FreeLibrary follows free() convention if being no-op on nullptr
   FreeLibrary(h);
};
};
std::unique_ptr<HMODULE, HModuleDeleter>
```
then delete the deconstructor

For Linux I will figure something out later

LR2/LR2_customir.cpp
```diff
@@ -0,0 +1,420 @@
...
+	struct SETTINGS {
+		int gaugeOption{};
+		std::array<int, 2> random{};
+		int autokey{};
+		std::array<int, 2> assist{};
+		int dpflip{};
+		int hsfix{};
+		std::array<int, 2> randSC{};
+		std::array<int, 2> randFix{};
+		int m_softlanding{};
```
chown2, line 143: redundant m_*

MatVeiQaaa: m is "maniac". Generally those names just follow whatever they are on mother structure.

LR2/LR2_customir.h
MatVeiQaaa: Why do we want to avoid deconstructor?

LR2/LR2_customir.cpp
```diff
@@ -0,0 +1,420 @@
...
+	return mMethods->_Login();
+}
+
+bool CustomIR::SendScoreV1(const IRScoreV1& score) {
+	MethodTableV1& methods = *(MethodTableV1*)mMethods;
+	if (methods._SendScoreV1 == nullptr) return false;
+	return methods._SendScoreV1(score);
+}
+
+void CUSTOMIR_MANAGER::Initialize(std::filesystem::path directory) {
```
chown2: same with const std::filesystem::path&, unless you have some plans that you can not share right now

LR2/LR2_customir.cpp
```diff
@@ -0,0 +1,420 @@
...
+	for (auto& file : std::filesystem::directory_iterator(_directory)) {
+		if (!file.is_regular_file()) continue;
+		if (file.path().extension().u8string() != u8".dll") continue;
+		if (!file.path().stem().u8string().ends_with(ARCH)) continue;
+		mDllHandle = LoadLibraryW(file.path().wstring().c_str());
+		if (mDllHandle == nullptr) continue;
+		auto GetMethodTable = (MethodTable* (__cdecl*)())GetProcAddress((HMODULE)mDllHandle, "GetMethodTable");
+		if (GetMethodTable == nullptr) continue;
+		mMethods = GetMethodTable();
+		if (mMethods == nullptr) continue;
```
chown2: ErrorLogAdd("bad dll") or this will be impossible to debug on module maker side

LR2/LR2_customir.h
```diff
@@ -0,0 +1,38 @@
...
+	MethodTable* mMethods = nullptr;
+};
+
+class CUSTOMIR_MANAGER {
+public:
+	void Initialize(std::filesystem::path directory);
+	void Login();
+	void SendScore(game& game, sqlite3* sql, int player);
+private:
+	void MakeScoreV1(IRScoreV1& scoreOut, const IRScoreInternal& score);
```
chown2: Better off moving this into a static method inside IRScoreV1
like
```
static IRScoreV1 FromInternal(const IRScoreInternal& score);
```

MatVeiQaaa: ah, true, thought about it too but didn't realize i could just make it static.

LR2/LR2_customir_api.h
```diff
@@ -0,0 +1,114 @@
...
+		std::array<int, 1000> combo{};
+		std::array<int, 1000> exscore{};
+		std::array<int, 1000> rate{};
+	} graphs{};
+};
+
+struct MethodTable {
+	const char*(__cdecl* _GetName)() = nullptr;
+	int(__cdecl* _GetApiVersion)() = nullptr;
+	bool(__cdecl* _Login)() = nullptr;
```
chown2: If you are going to separate structs by versions, might as well put everything other than _GetApiVersion into a specific version struct. What if we want to change '_Login' signature? I can totally see wanting to configure these modules on LR2 side, not on the module side.
Also, why prefix them with underscores?

LR2/LR2_customir_api.h
```diff
@@ -0,0 +1,114 @@
+#pragma once
+
+#include <string>
+#include <array>
+
+constexpr std::array<int, 1> APIVERSION = { 1 };
+
+#pragma pack(push, 1)
```
chown2: Does this actually help with something ABI-related, or was this just a 'feels right' thing?

LR2/LR2_customir.h
```diff
@@ -0,0 +1,38 @@
...
+	CustomIR() = delete;
+	CustomIR(std::filesystem::path directory);
+	~CustomIR();
+	bool Initialize();
+	bool Login();
+	bool SendScoreV1(const IRScoreV1& score);
+
+	std::string mName;
+	int mApiVersion = 0;
+private:
```
chown2:
```suggestion
  [[nodiscard]] const std::string& Name() const { return mName; };
  [[nodiscard]] int ApiVersion() const { return mApiVersion; };
private:
	std::string mName;
	int mApiVersion = 0;
```
hmm, exposing API version doesn't feel right...

LR2/LR2_customir.h
```diff
@@ -0,0 +1,38 @@
...
+struct IRScoreInternal;
+struct IRScoreV1;
+class CustomIR {
+public:
+	CustomIR() = delete;
+	CustomIR(std::filesystem::path directory);
+	~CustomIR();
+	bool Initialize();
+	bool Login();
+	bool SendScoreV1(const IRScoreV1& score);
```
chown2:
```suggestion
	bool SendScore(const IRScoreV1& score_v1);
```
and then when we add new version, add new score structuts as new arguments... then, make the API version completely private for this struct.

LR2/LR2_customir.cpp
```diff
@@ -0,0 +1,420 @@
...
+#endif
+
+CustomIR::CustomIR(std::filesystem::path _directory) {
+	for (auto& file : std::filesystem::directory_iterator(_directory)) {
+		if (!file.is_regular_file()) continue;
+		if (file.path().extension().u8string() != u8".dll") continue;
+		if (!file.path().stem().u8string().ends_with(ARCH)) continue;
+		mDllHandle = LoadLibraryW(file.path().wstring().c_str());
+		if (mDllHandle == nullptr) continue;
+		auto GetMethodTable = (MethodTable* (__cdecl*)())GetProcAddress((HMODULE)mDllHandle, "GetMethodTable");
```
chown2: avoid c-style casts
```suggestion
        // reinterpret cast safety: not safe lmao
		auto GetMethodTable = reinterpret_cast<MethodTable* (__cdecl*)()>(GetProcAddress((HMODULE)mDllHandle, "GetMethodTable"));
```

LR2/LR2_customir_api.h
MatVeiQaaa: because module and the game may be compiled completely differently, different compiler etc, the padding may be unpredictable.

LR2/LR2_customir.cpp
```diff
@@ -0,0 +1,420 @@
...
+	return true;
+}
+
+bool CustomIR::Login() {
+	if (mMethods->_Login == nullptr) return false;
+	return mMethods->_Login();
+}
+
+bool CustomIR::SendScoreV1(const IRScoreV1& score) {
+	MethodTableV1& methods = *(MethodTableV1*)mMethods;
```
chown2: avoid c-style casts
```suggestion
        // reinterpret cast safety: not safe lmao
        MethodTableV1& methods = *reinterpret_cast<MethodTableV1*>(mMethods);
```

LR2/LR2_customir.h
chown2: Move semantics. Look up 'rule of five'.

LR2/LR2_customir_api.h
chown2: As far as I know, padding only depends on the target architecture, not the compiler, and as such this is redundant

chown2 (commented):
Looks good but I'd move some things around (comments above)

LR2/LR2_customir_api.h
MatVeiQaaa: name and api version are essential and are not to change. I guess it makes some sense to have login under a version too, it's how it was originally, but i decided against it in the end, as i struggled to imagine real use of that.

LR2/LR2_customir.h
MatVeiQaaa: the reason i haven't done that is that i wanted to avoid function declaration in api header

ExampleIR/ExternalIR.cpp
```diff
@@ -0,0 +1,99 @@
...
+        score.judgements_total.epr + score.judgements_total.lpr,
+        score.judgements_total.fast, score.judgements_total.slow, score.judgements_total.cb,
+        lamps[score.clearType]
+    );
+    std::ofstream dump(path / filename);
+    dump << processedScore;
+    return true;
+}
+
+void _GetMethodTable(MethodTable& table) {
```
chown2: Underscored functions in global scope are undefined behavior. https://clang.llvm.org/extra/clang-tidy/checks/bugprone/reserved-identifier.html
And you don't need this helper, just let GetMethodTable be this.

ExampleIR/ExternalIR.cpp
```diff
@@ -0,0 +1,99 @@
...
+    wchar_t modulePath[MAX_PATH]{};
+    switch (ul_reason_for_call)
+    {
+    case DLL_PROCESS_ATTACH:
+        // Get path to the folder the .dll is running from.
+        // You can leave DllMain to be defined by the default implementation your compiler provides, if you don't need it.
+        // Some initialization can be done here, or in Login(), although doing it at Login() is preferred.
+        GetModuleFileNameW(hModule, modulePath, MAX_PATH);
+        path = modulePath;
+        path = path.parent_path();
```
chown2: IMO runtime data and executables shouldn't be stored together. This could just be stored in a well-known path, like `LR2files/CustomIRConfig/ExternalIr/whateveritwants.whatever`

LR2/LR2_customir.cpp
```diff
@@ -0,0 +1,430 @@
...
+#endif
+#endif
+
+CustomIR::CustomIR(const std::filesystem::path& _directory) {
+	for (auto& file : std::filesystem::directory_iterator(_directory)) {
+		if (!file.is_regular_file()) continue;
+		if (file.path().extension().u8string() != u8".dll") continue;
+		if (!file.path().stem().u8string().ends_with(ARCH)) continue;
+		mDllHandle.reset(new HMODULE(LoadLibraryW(file.path().wstring().c_str())));
+		if (*mDllHandle == nullptr) continue;
```
chown2: Is this double pointer indirection? I don't get how this works when it's dereferenced
```suggestion
		mDllHandle.reset(LoadLibraryW(file.path().wstring().c_str()));
		if (mDllHandle == nullptr) continue;
```

LR2/LR2_customir.h
```diff
@@ -0,0 +1,37 @@
...
+	bool Initialize();
+	bool Login();
+	bool SendScore(const IRScoreV1& score);
+
+	[[nodiscard]] const std::string& Name() const { return mName; };
+private:
+	struct ModuleDeleter {
+		void operator()(HMODULE* handle);
+	};
+	std::unique_ptr<HMODULE, ModuleDeleter> mDllHandle;
```
chown2: I see... HMODULE is void*, so we end up getting void**. Would this work?
```suggestion
	struct ModuleDeleter {
		void operator()(void* handle);
	};
	std::unique_ptr<void, ModuleDeleter> mDllHandle;
```

LR2/LR2_customir.cpp
```diff
@@ -0,0 +1,430 @@
...
+		internal.MakeScoreV1(scoreV1);
+		std::vector<std::future<bool>> sendThreads;
+		sendThreads.reserve(mModules.size());
+		for (auto& module : mModules) {
+			sendThreads.push_back(std::async(std::launch::async, &CustomIR::SendScore, &module, scoreV1));
+		}
+		for (auto& thread : sendThreads) {
+			thread.get();
+		}
+	}, std::move(internal)).detach();
```
chown2: std::move is error-prone, let's avoid it when possible.

```suggestion
	std::jthread([this](IRScoreInternal internal) {
		IRScoreV1 scoreV1;
		internal.MakeScoreV1(scoreV1);
		std::vector<std::future<bool>> sendThreads;
		sendThreads.reserve(mModules.size());
		for (auto& module : mModules) {
			sendThreads.push_back(std::async(std::launch::async, &CustomIR::SendScore, &module, scoreV1));
		}
		for (auto& thread : sendThreads) {
			thread.get();
		}
	}, IRScoreInternal{game, sql, player}).detach();
```

LR2/LR2_customir_api.h
```diff
@@ -0,0 +1,106 @@
...
+		std::array<int, 1000> combo{};
+		std::array<int, 1000> exscore{};
+		std::array<int, 1000> rate{};
+	} graphs{};
+};
+
+struct MethodTable {
+	const char*(__cdecl* _GetName)() = nullptr;
+	bool(__cdecl* _LoginV1)() = nullptr;
+	bool(__cdecl* _SendScoreV1)(const IRScoreV1& score) = nullptr;
```
chown2: Let's drop the underscore prefix, it doesn't serve any meaning

ExampleIR/ExternalIR.cpp
```diff
@@ -0,0 +1,99 @@
+#define WIN32_LEAN_AND_MEAN
```
chown2: nit: imo this should live in the VS solution, like how I made it in the OpenLR2 solution

ExampleIR/ExternalIR.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+#include <LR2_customir_api.h>
+#include <windows.h>
+#include <filesystem>
+#include <fstream>
+#include <format>
+
+std::filesystem::path path;
```
chown2:
```suggestion
namespace GlobalState {
std::filesystem::path path;
int scoresSaved = 0;
} // namespace GlobalState
```
or
```suggestion
namespace {
std::filesystem::path g_path;
int g_scoresSaved = 0;
} // namespace
```
or
```suggestion
static std::filesystem::path g_path;
static int g_scoresSaved = 0;
```

LR2/LR2_customir.cpp
```diff
@@ -0,0 +1,430 @@
...
+	std::jthread([this](IRScoreInternal internal) {
+		IRScoreV1 scoreV1;
+		internal.MakeScoreV1(scoreV1);
+		std::vector<std::future<bool>> sendThreads;
+		sendThreads.reserve(mModules.size());
+		for (auto& module : mModules) {
+			sendThreads.push_back(std::async(std::launch::async, &CustomIR::SendScore, &module, scoreV1));
+		}
+		for (auto& thread : sendThreads) {
+			thread.get();
```
chown2: for a follow up
```suggestion
			if (!thread.get()) {
			    // TODO: retry();
			}
```

LR2/LR2_customir.cpp
chown2: .detach will blow up if we're still here on game exit...
I'm not sure if this is a good idea, but we could have hModules as a shared_ptr we would promise not to update values behind after game init

LR2/LR2_customir.cpp
```diff
@@ -0,0 +1,430 @@
...
+#else
+constexpr const char8_t* ARCH = u8".x64";
+#endif
+#elif _WIN32
+#if _DEBUG
+constexpr const char8_t* ARCH = u8"_D.x86";
+#else
+constexpr const char8_t* ARCH = u8".x86";
+#endif
+#endif
```
chown2, line 31: nit: avoid strlen to make the code BLAZINGLY fast :rock:  :rock:  :rock: 
```suggestion
#if _WIN64
#if _DEBUG
constexpr auto&& ARCH = u8"_D.x64";
#else
constexpr auto&& ARCH = u8".x64";
#endif
#elif _WIN32
#if _DEBUG
constexpr auto&& ARCH = u8"_D.x86";
#else
constexpr auto&& ARCH = u8".x86";
#endif
#endif
```

LR2/LR2_customir.cpp
MatVeiQaaa: i thought ive already done that, just didnt push i guess

MatVeiQaaa: detach will not blow up because we dont hold onto any state outside of the thread. All the context of the thread is local to it. It's why i brought internal score, which is built from game state, to the outside of the thread, to block the execution until its done parsing. What could maybe blow up however is module unloading i guess, while its code is still running. Maybe hold a shared_ptr to the module in the thread? Oh, well that is what you mentioned...

ExampleIR/ExternalIR.cpp
chown2: whatever

LR2/LR2_customir.cpp
```diff
@@ -0,0 +1,482 @@
...
+			std::vector<std::future<SendScoreStatus>> sendThreads;
+			sendThreads.reserve(toSend.size());
+			for (const auto& module : toSend) {
+				sendThreads.push_back(std::async(std::launch::async, &CustomIR::SendScore, module, scoreV1));
+			}
+
+			for (const auto& [i, t] : std::views::enumerate(sendThreads) | std::views::reverse) {
+				switch (t.get()) {
+				case SendScoreStatus::Fail:
+					OverlayNotification("'%s' failed to submit score", toSend[i]->Name());
```
chown2: 1) Name() returns garbage
2)
```suggestion
					OverlayNotification("'%s' failed to submit score\n", toSend[i]->Name());
```

chown2 (approved):

chown2:
This is ready

PR #38, opened 2025-12-01
fix: fmod failing to load sounds with multibyte in path

MatVeiQaaa:
This is better than nothing, although i can't say this is an appropriate long-term solution. All paths must be stored and processed as 16bit char strings, or even better utf8. In ideal scenario, every string that comes from outside world (user input, files, or from LR2IR) must be stored and further processed as utf8.

### DIFF
--- a/LR2/En_audio.cpp
+++ b/LR2/En_audio.cpp
@@ -680,7 +680,7 @@ int LoadSound(AUDIO *aud, SOUNDDATA *sound, CSTR filepath, int loop, int disable
 					}
 				}
 			}
-			result = FMOD_System_CreateSound(aud->fmodSys, s2utf8(filepath.body, 932).c_str(), mode, NULL, &sound->fmod_sound);
+			result = FMOD_System_CreateSound(aud->fmodSys, s2utf8(filepath.body, CP_ACP).c_str(), mode, NULL, &sound->fmod_sound);
 		}
 
 		sound->flag2c = 0;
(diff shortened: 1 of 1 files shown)


chown2 (approved):

LR2/En_audio.cpp
```diff
@@ -671,7 +680,7 @@ int LoadSound(AUDIO *aud, SOUNDDATA *sound, CSTR filepath, int loop, int disable
 					}
 				}
 			}
-			result = FMOD_System_CreateSound(aud->fmodSys, filepath, mode, NULL, &sound->fmod_sound);
+			result = FMOD_System_CreateSound(aud->fmodSys, s2utf8(filepath.body, 932).c_str(), mode, NULL, &sound->fmod_sound);
```
chown2: CP_ACP

chown2 (changes requested):
CP_ACP

LR2/En_audio.cpp
MatVeiQaaa: why? This will only lead to unpredictable behavior. If we know exactly the codepage we expect, why would we leave it to system settings?

LR2/En_audio.cpp
```diff
@@ -603,6 +604,14 @@ int RecordFadeout(AUDIO *aud, double from, double length) {
 	return 1;
 }
 
+static std::string s2utf8(const std::string_view& str, unsigned int codepage) {
```
chown2:
```suggestion
static std::string s2utf8(const std::string_view str, unsigned int codepage) {
```

chown2 (approved):
:dash: :rocket: LGTM!!!!

PR #41, opened 2025-12-04
feat: utf-8 strings

MatVeiQaaa:
Makes the game use UTF-8 strings internally. This makes it easier to maintain, and obsoletes the necessity of running the game in japanese locale.

ISSUES:
DxLib fails to generate some non-image fonts without locale anyway, even though it is using UTF-8 too. I couldn't figure this out.
Requires song.db rebuild, as CRC32 of non-ASCII paths changed. There is no way around it.
Some input and output data should be converted to/from SJIS. Maybe some output to LR2IR, some skins data (strings), charts.

### DIFF
--- a/LR2/strclass.cpp
+++ b/LR2/strclass.cpp
@@ -1,5 +1,5 @@
 #include "strclass.h"
-
+#include "filesystem.h"
 #include <cstring>
 #include <span>
 #include <string>
@@ -334,7 +334,7 @@ int CSTR::toFile(const char *filepath) {
 	FILE *_File;
 	char *pcVar2;
 
-	_File = _wfopen(utf2ws(filepath).c_str(), L"w");
+	_File = fopen(filepath, "w");
 	if (_File == (FILE *)0x0) {
 		return -1;
 	}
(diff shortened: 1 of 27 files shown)


chown2:
> Some input and output data should be converted to/from SJIS. Maybe some output to LR2IR, some skins data (strings), charts.

File reading should be (at least mostly) implemented.
TODO: check IR sending, don't forget there are two places it's done.
The project is also Windows 10+ now. We *could* avoid that by approaching things differently, but it's dozens of hours of work and a big headache for little gain.
TODO: include the OpenLR2.manifest manifest in Visual Studio solutions. But you should really use CMake now.

Hm, last commit introduces unnecessary `#include "filesystem.h"`, forgot about them. Not gonna bother, we should bring in clang-tidy to do that for us instead.

LR2/strclass.cpp
```diff
@@ -569,34 +532,21 @@ CSTR& CSTR::operator=(const char *str) {
...
-				pos = i;
-			}
-			i = i + 1;
-		}
-	}
-	return left(pos + 1);
+CSTR CSTR::getDirectory() {
+	CSTR out;
+	std::filesystem::path path(reinterpret_cast<char8_t*>(this->body));
+	out.assign(cs(path.parent_path().u8string()));
```
chown2, line 544: after last commit 
```suggestion
	std::filesystem::path path(this->body);
	out.assign(path.parent_path().string());
```

PR #45, opened 2025-12-12
refactor: replace avgbpm with mainbpm

MatVeiQaaa:
I think avgbpm is rather useless, and adding mainbpm as 5th hs-fix mode requires modifying skins, else it is displayed as const. Its better to replace avgbpm.

### DIFF
--- a/LR2/LR2_skinload.cpp
+++ b/LR2/LR2_skinload.cpp
@@ -433,6 +433,7 @@ int ReadImageFont(CSTR filename, ImageFont *imgfont) {
 	if (strcmp(str1, imgfont->filepath)) {
 		imgfont->size = 0;
 		imgfont->kerning = 0;
+		imgfont->filepath[0] = '\0';
 		for (auto& [idx, chTex] : imgfont->chars) {
 			DeleteGraph(chTex.grHandle);
 			chTex.grHandle = -1;


GOMazk:
just add 'MAINBPM' on LR2files/config/optionstr.csv, it works

MatVeiQaaa:
> just add 'MAINBPM' on LR2files/config/optionstr.csv, it works

Only works in skins which use the string. A lot of skins use a texture.

GOMazk:
> > just add 'MAINBPM' on LR2files/config/optionstr.csv, it works
> 
> Only works in skins which use the string. A lot of skins use a texture.


Oh, I didn't know.

however, I actually want skin-related issues to be solved by easy GUI skin editor. what about holding this thing until I finish skin editor?
+rethink, swap position miainBPM and avg, so avg move to 5 and hidden until someone request seem fine also. how about this?

MatVeiQaaa:
> > > just add 'MAINBPM' on LR2files/config/optionstr.csv, it works
> 
> > 
> 
> > Only works in skins which use the string. A lot of skins use a texture.
> 
> 
> 
> 
> 
> Oh, I didn't know.
> 
> 
> 
> however, I actually want skin-related issues to be solved by easy GUI skin editor. what about holding this thing until I finish skin editor?
> 
> +rethink, swap position miainBPM and avg, so avg move to 5 and hidden until someone request seem fine also. how about this?
> 
> 

Even skin editor wouldn't solve that problem easily, because a user would have to create a texture. I like the swapping idea, though. Even better is if it goes main->avg->const

MatVeiQaaa:
@GOMazk done, i think this is the best medium

LR2/LR2_bmsload.cpp
```diff
@@ -2630,7 +2630,7 @@ int ParseBmsFile(gameplay *gp, CSTR filename, AUDIO *aud, ConfigStruct* cfg, BMS
 				double addRealtime = (240.0 / nowBPM * meaLength * (gp->bmsobj.notes[i].bmsTiming - prevNoteBmstime) * 1000.0);
 				bpmt_realtime += addRealtime + stopRealtime;
 				
-				if (cfg->play.hsfix == 4 || (gp->isCourse && gp->courseType == 1)) {
+				if (cfg->play.hsfix == 5 || (gp->isCourse && gp->courseType == 1)) {
```
chown2: if you went through all constants, might as well have replaced them with enumerations

chown2:
How do skins usually implement hi-speed mode display? Doesn't this break every skin in existence?

GOMazk:
skins - 3avg 4const
code - 3main 4avg 5const
would break old skins

I meant
skins - 3avg 4const
code - 3main 4const 5(reserved for avg)
and this is almost same your first work

MatVeiQaaa:
> How do skins usually implement hi-speed mode display? Doesn't this break every skin in existence?

by id with multi-div textures. There isn't a way to do this without "breaking" skins, no matter how you do it.

PR #47, opened 2025-12-30
RE issue: missing imagefont filepath reset

### LINKED ISSUE #46: Bug: backing out into select from decide, imagefont for songbar fails to load

MatVeiQaaa:
caused a bug that aborting imagefont loading due to failure to open the file lead to unloading of the old path font, and blocking the next font from same old path to load.


chown2 (approved):
Checked with Ghidra :+1:

Issue #50, opened 2026-01-26
Duplicated input


chown2:
These happen together in song select sometimes:
- Mouse wheel is too intensive, one wheel turn scrolls like 10 songs
- Right-clicking on option panels not only closes the panel but also goes back a folder in select
- Going up/down in MANIAC OPTIONS goes several entries at once.

Reproduced by exiting play scene while it's entrance animation plays.

chown2:
https://github.com/GOMazk/OpenLR2/blob/66f05466fb3192c872e0eaba6de0ae57d91abe78/LR2/Main.cpp#L2058-L2060

Seems to be related to this.
Somehow input is only polled once in several frames.

This can be work-around here by putting `g->gameplay.flag_gameinput = 0;` at the end, but there isn't such statement in LR2. EDIT: but that breaks the play scene.
https://github.com/GOMazk/OpenLR2/blob/66f05466fb3192c872e0eaba6de0ae57d91abe78/LR2/Scene04_Play.cpp#L1775-L1784

chown2:
https://github.com/GOMazk/OpenLR2/blob/66f05466fb3192c872e0eaba6de0ae57d91abe78/LR2/Scene04_Play.cpp#L1679-L1708
These 'break' are supposed to be 'return'.
PR incoming.

PR #60, opened 2026-01-28
Fix two RE issues

chown2:
Fixes https://github.com/GOMazk/OpenLR2/issues/50 "Duplicated input".


GOMazk (approved):
RE issues will be copied into beta3 branch 1~2 February, thanks

PR #63, opened 2026-02-09
fix render inconsistencies

MatVeiQaaa:
First of all, DxLib has a bug with F variants of draw methods, where with certain input values it would incorrectly apply the UV, effectively shifting SRC coordinate by one pixel to the left, or up. This bug can be observed both in LR2 and OpenLR2 (LN body sometimes detaching from head and tail, gauge having gaps which are not supposed to be there, etc). It looks slightly different between LR2 and OpenLR2 because LR2 version of DxLib had another bug with position logic, which was fixed by the version we use in OpenLR2. You can see the comment about it in DxLib.h next to DX_DRAWFLOATCOORDTYPE_DIRECT3D9 define.

Using integer variants of draw methods instead of F variants, besides solving the problem with SRC coordinate shift, yields much better render consistency, textures have sharper edges and look much closer to their source image. Both in LR2 and previously in OpenLR2 you can observe a ton of render imperfections if you look closely and compare it to the source images, as well as some jitter in moving objects.

Please try this first and see if there are any downsides to this method. I don't think i noticed any.

### DIFF
--- a/LR2/LR2_skindraw.cpp
+++ b/LR2/LR2_skindraw.cpp
@@ -840,17 +840,17 @@ int LRDrawImg(int *grHandle, DSTdraw *dstD) {
 		vec4.y = y2 - ry;
 		VectorRotationZ(&vec4, &vec4, rad);
 		if (dstD->h != 0.0 && dstD->w != 0.0 && dstD->time >= 0) {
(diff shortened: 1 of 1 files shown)


MatVeiQaaa:
\+ 0.5f on input values is needed because the values are downcasted from float to int, by truncating the values. + 0.5f makes the values round to closest integer after truncation.

GOMazk:
sry for late. I tried it and cannot notice any difference in gameplay(default skin). even if there are any trouble, too subtle to notice I think. I'll merge it

PR #64, opened 2026-02-11
feat: add nowplaying.json with current scene

chown2:
This allows achieving scene switching functionality in OBS by using a plugin, and the play scene also has metadata included allowing achieving nowplaying.txt functionality by using an additional external script. Not sure if the play scene metadata is actually any useful though.


MatVeiQaaa (approved):

GOMazk (approved):
lgtm, thanks

Issue #73, opened 2026-06-01
Song bar fonts disappear


PARK-SU:
<img width="1274" height="755" alt="Image" src="https://github.com/user-attachments/assets/1e27ed55-44ff-447b-ada3-467f89385677" />

I deleted song.db and tried re-running it, but it still won't run. (64bit build)

MatVeiQaaa:
@PARK-SU please make separate issues for separate problems. I didn't immediately notice you are also reporting a crash.

MatVeiQaaa:
https://github.com/GOMazk/OpenLR2/pull/80 fixes the crash.

MatVeiQaaa:
i can reproduce fonts failing to display with wmix, returning from play scene. Related to https://github.com/GOMazk/OpenLR2/pull/47?

Catizard:
> i can reproduce fonts failing to display with wmix, returning from play scene. Related to [#47](https://github.com/GOMazk/OpenLR2/pull/47)?

I noticed the issue in other clients too, the texts are missing after returning from play stage. Since the bug is reproducible on other clients I would venture to say we encounter the same issue and it is inside the skin, here's the lines wmix defines the bar titles, as you can see the destination commands are using a dstop=1

```
//SRCè`,(NULL),font,st,align,,,,,,,,,,,,,,,,
#SRC_BAR_TITLE,0,4,0,0,,,,,,,,,,,,,,,,
//DSTè`,(NULL),time,x,y,w,size,acc,a,r,g,b,blend,filter,(NULL),(NULL),loop,timer,(NULL),(NULL),(NULL)
#DST_BAR_TITLE,0,0,99,2,720,34,0,100,255,255,255,1,1,0,0,166,0,1,0,0
#DST_BAR_TITLE,0,166,99,2,720,34,0,255,255,255,255,1,1,0,0,,,,,
,,,,,,,,,,,,,,,,,

//SRCè`,(NULL),font,st,align,,,,,,,,,,,,,,,,
#SRC_BAR_TITLE,1,4,0,0,,,,,,,,,,,,,,,,
//DSTè`,(NULL),time,x,y,w,size,acc,a,r,g,b,blend,filter,(NULL),(NULL),loop,timer,(NULL),(NULL),(NULL)
#DST_BAR_TITLE,1,0,99,2,720,34,0,255,255,255,255,1,1,0,0,0,0,1,0,0
[...]

PR #80, opened 2026-06-02
wrong alloc size for SkinHeader in x64

MatVeiQaaa:
Stack overflow crash in debug build, missing stack reserve size in build.
Wrong SkinHeader allocation size, due to hardcoded structure size.
Fixes x64 crash mentioned in https://github.com/GOMazk/OpenLR2/issues/73


GOMazk (commented):
thanks

PR #104, opened 2026-06-06
Custom ir/api header

luxurya111:
Extends `LR2_customir_api.h` with types and `MethodTable` slots for IR rank/stats/leaderboard/rival/ghost data.

- Read slots use `Get*V1` + `GetStatus`; score submit stays `SendScoreV1` + `SendScoreStatus`
- New slots default to `nullptr` — existing IR DLLs keep working unchanged
- Adds ABI comments (append-only contract, UI mapping hints)

### DIFF
--- a/LR2/LR2_customir_api.h
+++ b/LR2/LR2_customir_api.h
@@ -1,9 +1,15 @@
 #pragma once
 
+// CustomIR public ABI between OpenLR2 host and third-party IR DLLs.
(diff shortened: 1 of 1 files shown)


chown2:
Please, also add the code integrating these new APIs with OpenLR2, and make a single PR for each them instead of a big single one. Reviewing all changes at once would be very difficult.

luxurya111:
I think the way I split the PR was incorrect, so I'll close this one.

PR #105, opened 2026-06-06
feat: add random unrandomizer

chown2:
Simple search bar based interface, while we are figuring out how to do add a GUI for this.


MatVeiQaaa:
why not p2 arrange display too?

LR2/Scene02_Songselect.cpp
```diff
@@ -1283,6 +1283,35 @@ int CmdSearch(game *g, CSTR *cmd, sqlite3 *sql) {
...
+		if (ec != std::errc()) {
+			*cmd = ("INVALID NUMBER: " + std::string(s)).c_str();
+			return 1;
+		}
+		if (val == 0) {
+			g->gameplay.forceRandomLayout = 0;
+			*cmd = g->gameplay.forceRandomLayout == 0 ? "OFF" : std::to_string(g->gameplay.forceRandomLayout).c_str();
+		}
+		else if (val >= 12345 && val <= 54321) {
+			g->gameplay.forceRandomLayout = val * 100 + 67;
```
MatVeiQaaa, line 1303: LOL genius

MatVeiQaaa (approved):
it's alright, a bit of a mess in seed acquisition bit, but it's a short piece of code, whatever.

MatVeiQaaa:
i think you should add p2 arrange display though

chown2:
~~14k doesn't exist~~ compared to P1 I don't have a skin to test with, and I don't know if anyone would actually use that to justify spending 5 minutes to add random display into a new skin. :D Do note that the DST number for P2, 418, is only used for DP.

Sonic-TV:
> ~~14k doesn't exist~~ compared to P1 I don't have a skin to test with, and I don't know if anyone would actually use that to justify spending 5 minutes to add random display into a new skin. :D Do note that the DST number for P2, 418, is only used for DP.

I've used plenty of LR2OOL exclusive stuff in my skin `extended-tricoro-for-lr2`, you may test with that.

chown2:
@Sonic-TV thank you, it helps, I will add P2 display. Do note I do not endorse skins using extracted textures however. :)

Sonic-TV:
> @Sonic-TV thank you, it helps, I will add P2 display. Do note I do not endorse skins using extracted textures however. :)

I was thinking of making something better than LITONE and not let everyone paying for this, so yeah (lol).

chown2:
Added P2 random layout display. @Sonic-TV your skin uses the numbers in a different way too, which exposed an incorrectly handled edge case. Thank you!

Sonic-TV:
> Added P2 random layout display. @Sonic-TV your skin uses the numbers in a different way too, which exposed an incorrectly handled edge case. Thank you!

I've used all of the button and text definitions in skin select (and plenty of `a/*/c` structures in entire skin as well), so you may test the functionality with it.

GOMazk (approved):
what a magic show lol

PR #111, opened 2026-06-07
fix: fix getParentDirectory being the same as getDirectory 

chown2:
Fixes songs being nested in their folders after a rescan.


GOMazk (commented):
lgtm, thank you

PR #112, opened 2026-06-07
feat(customir): add result rank and leaderboards

luxurya111:
This PR implements a feature where the display provider DLL supplies the current rank and total player count, and the host displays them on the result and song select screens (e.g., Rank X out of X total). Cache storage location and format are module-defined; see `ExampleIR/ExternalIR.cpp` for the reference skeleton. This PR covers result/song-select rank display. leaderboard UI and ghost integration are follow-ups. However, to align with the design of the existing DLL, the following specifications have been added:

- ~~By adding `active.ini` inside the `CustomIRs` folder~~ (read EDIT), you can configure which DLL is the display provider for ranking data. This is not required; even if the function to display Ranking Data (described below) is not set up, existing logic such as Login and SendScore still works regardless of how many DLLs are added.

- GetResultRank: Fetches IR Ranking data (on the result screen, after the display provider's SendScore; modules may use server communication or local cache—module designers may persist cache in a separate location. This data can be used in the future for leaderboards, ghost battles, etc., just like the existing IR/<hash>.xml. CustomIR exposes snapshot fields analogous to IR XML elements from the traditional IR/<hash>.xml. The supported cache fields are as follows:

    ranking[]: Ranking array; equivalent to <score> in the traditional IR/<hash>.xml
    std::string lastupdate: Last updated date; equivalent to <lastupdate> in the traditional IR/<hash>.xml
    myRank: Current rank
    totalPlayer: Total number of players
    totalPlaycount: Total play count
    clearPlayers[]: Clear players by gauge type, length 6 — [0]: no play, [1]: fail, [2]: easy, [3]: normal, [4]: hard, [5]: fullcombo
    rivalRank: Rival rank (reserved; ignored by the host in this PR)

The reason for expanding on the traditional IR/<hash>.xml structure is that under the existing dream-pro specifications, the system fetches the data of all users after a play, and then parses every user's data from that XML to calculate myRank, totalPlayer, totalPlaycount, etc. For Insane BMS, which carries a massive amount of IR data, this increases the volume of ranking data to load, resulting in a significant delay when moving from the result screen back to the song select screen. Given that the actual leaderboard UI only displays up to 999 players, the module developer can set a limit on the amount of data the server fetches in the future for the ranking[] data. In other words, when a display provider supplies rank data, the host merges the snapshot instead of deriving summary fields (clear count by gauge, current rank, total play count, etc.) from the full ranking[] board.

- RestoreCachedRank: Restores a cached IR ranking snapshot on the song select screen (song hash only). If not implemented, CustomIR rank is not restored from disk after restart (it can still be updated after a play via GetResultRankV1). Modules may optionally implement cache refresh/TTL (similar to autoupdate) inside this hook; the host does not read config.network.autoupdate for CustomIR. dream-pro XML auto-update remains a separate existing path when online.
----
EDIT: I have modified the implementation to reference the existing openlr2-config.xml instead of using a separate active.ini configuration file. If it has not been added yet, please modify your XML file as follows:

```
<network>
  <customir_provider>ExampleIR</customir_provider>
</network>
```

If OpenLR2.exe has already been executed in the past, it must be added manually.

### DIFF
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -4,7 +4,9 @@
 #include <fstream>
 #include <format>
 
+#ifdef _WIN32
 #include <windows.h>
+#endif // _WIN32
 
 namespace State {
     static std::filesystem::path path;
@@ -62,7 +64,7 @@ static SendScoreStatus SendScore(const IRScoreV1& score) {
     return SendScoreStatus::Ok;
 }
 
-extern "C" __declspec(dllexport) void GetMethodTable(MethodTable& table) {
+extern "C" OLR2_IR_EXPORT void GetMethodTable(MethodTable& table) {
     // Fill out the pointers to methods you want to use. Leave them at nullptr if you don't want to use them.
     // Only essential method is GetName(). Without it, your module will be rejected.
     // As API gets updated, new methods may appear available at MethodTable, but old ones will never be removed or their prototypes modified. Method indexes are also stable.
@@ -71,6 +73,7 @@ extern "C" __declspec(dllexport) void GetMethodTable(MethodTable& table) {
     table.SendScoreV1 = &SendScore;
 }
 
+#ifdef _WIN32
 BOOL APIENTRY DllMain( HMODULE hModule,
                        DWORD  ul_reason_for_call,
                        LPVOID lpReserved
@@ -94,4 +97,4 @@ BOOL APIENTRY DllMain( HMODULE hModule,
     }
     return TRUE;
 }
-
+#endif // _WIN32
--- a/LR2/LR2_customir.cpp
+++ b/LR2/LR2_customir.cpp
@@ -1,38 +1,67 @@
 #include "LR2_customir.h"
 
-#ifdef _WIN32
-
 #include "LR2_customir_api.h"
 #include "LR2_songmanage.h"
 #include "structure.h"
 
+#include <algorithm>
+#include <cmath>
 #include <filesystem>
-#include <format>
 #include <future>
-#include <queue>
 #include <ranges>
 #include <thread>
-#include <type_traits>
 
 #include <DxLib/DxLib.h>
+
+#ifdef _WIN32
 #include <libloaderapi.h>
 #include <wtypes.h>
+#else
+#include <dlfcn.h>
+#endif // _WIN32
+
+#if _WIN32
 
 #if _WIN64
 #if _DEBUG
 constexpr auto&& ARCH = "_D.x64";
 #else
 constexpr auto&& ARCH = ".x64";
-#endif
-#elif _WIN32
+#endif // _DEBUG
+#else
 #if _DEBUG
 constexpr auto&& ARCH = "_D.x86";
 #else
 constexpr auto&& ARCH = ".x86";
-#endif
+#endif // _DEBUG
+#endif // _WIN64
+
+constexpr auto&& DLL = ".dll";
+
 #else
-#error "Unsupported Arch Linux"
-#endif
+
+#if __x86_64__
+#if _DEBUG
+constexpr auto&& ARCH = "_D.x64";
+#else
+constexpr auto&& ARCH = ".x64";
+#endif // _DEBUG
+#else
+#if _DEBUG
+constexpr auto&& ARCH = "_D.x86";
+#else
+constexpr auto&& ARCH = ".x86";
+#endif // _DEBUG
+#endif // __x86_64__
+
+constexpr auto&& DLL = ".so";
+
+using HMODULE = void*;
+static HMODULE LoadLibrary(const char* fp) { return dlopen(fp, RTLD_NOW); }
+static void* GetProcAddress(HMODULE h, const char* name) { return dlsym(h, name); }
+static void FreeLibrary(HMODULE h) { dlclose(h);}
+
+#endif // _WIN32
 
 // TODO
 #define OverlayNotification ErrorLogFmtAdd
@@ -58,24 +87,29 @@ private:
 CustomIR::CustomIR(const std::filesystem::path& _directory) {
 	for (auto& file : std::filesystem::directory_iterator(_directory)) {
 		if (!file.is_regular_file()) continue;
-		if (file.path().extension().string() != ".dll") continue;
+		if (file.path().extension().string() != DLL) continue;
+		const auto filename = file.path().filename().string();
+		ErrorLogFmtAdd("Trying to load CustomIR %s", filename.c_str());
 		if (auto s = file.path().stem().string(); !s.ends_with(ARCH)) {
-			ErrorLogFmtAdd("'%s' skipping IR module with invalid file name stem (expected %s)\n", s.c_str(), ARCH);
+			ErrorLogFmtAdd("'%s' skipping per invalid file name stem (expected %s)\n", s.c_str(), ARCH);
 			continue;
 		}
-		mDllHandle.reset(LoadLibraryW(file.path().wstring().c_str()));
-		if (mDllHandle == nullptr) continue;
-		auto GetMethodTable = reinterpret_cast<void (__cdecl*)(MethodTable&)>(GetProcAddress(mDllHandle.get(), "GetMethodTable"));
+		mDllHandle.reset(LoadLibrary(file.path().string().c_str()));
+		if (mDllHandle == nullptr) {
+			ErrorLogFmtAdd("'%s' LoadLibrary failed\n", filename.c_str());
+			continue;
+		}
+		auto GetMethodTable = reinterpret_cast<void (OLR2_IR_API*)(MethodTable&)>(GetProcAddress(mDllHandle.get(), "GetMethodTable"));
 		if (GetMethodTable == nullptr) {
-			ErrorLogFmtAdd("'%s' not loaded, missing 'GetMethodTable' export.\n", file.path().filename().string().c_str());
+			ErrorLogFmtAdd("'%s' skipping per missing 'GetMethodTable' export\n", filename.c_str());
 			continue;
 		};
 		GetMethodTable(mMethods);
 		if (mMethods.GetName == nullptr) {
-			ErrorLogFmtAdd("'%s' not loaded, missing essential 'GetName' implementation.\n", file.path().filename().string().c_str());
+			ErrorLogFmtAdd("'%s' skipping per missing 'GetName' implementation\n", filename.c_str());
 			continue;
 		};
-		ErrorLogFmtAdd("'%s' loaded: %s\n", file.path().filename().string().c_str(), mMethods.GetName());
+		ErrorLogFmtAdd("CustomIR %s loaded: %s\n", filename.c_str(), mMethods.GetName());
 		break;
 	}
 }
@@ -503,11 +537,3 @@ void CUSTOMIR_MANAGER::SendScore(game& game, sqlite3* sql, int player) {
 		}
 	}, IRScoreInternal{ game, sql, player }, mModules));
 }
-#else
-
-CUSTOMIR_MANAGER::~CUSTOMIR_MANAGER() {};
-void CUSTOMIR_MANAGER::Initialize(const std::filesystem::path& directory) {};
-void CUSTOMIR_MANAGER::Login() {};
-void CUSTOMIR_MANAGER::SendScore(game& game, sqlite3* sql, int player) {};
-
-#endif // _WIN32
--- a/LR2/LR2_customir.h
+++ b/LR2/LR2_customir.h
@@ -1,11 +1,8 @@
 #pragma once
 
-#include <array>
 #include <filesystem>
 #include <future>
 #include <memory>
-#include <string>
-#include <type_traits>
 #include <vector>
 
 struct game;
--- a/LR2/LR2_customir_api.h
+++ b/LR2/LR2_customir_api.h
@@ -3,6 +3,14 @@
 #include <string>
 #include <array>
 
+#ifdef _WIN32
+#define OLR2_IR_API __cdecl
+#define OLR2_IR_EXPORT __declspec(dllexport)
+#else
+#define OLR2_IR_API
+#define OLR2_IR_EXPORT
+#endif // _WIN32
+
 struct IRScoreV1 {
 	struct SONG {
 		std::string hash;
@@ -105,14 +113,8 @@ enum class SendScoreStatus: int {
 	Fail,
 };
 
-#ifndef _WIN32
-#define __cdecl
-#endif // _WIN32
 struct MethodTable {
-	const char*(__cdecl* GetName)() = nullptr;
-	bool(__cdecl* LoginV1)() = nullptr;
-	SendScoreStatus(__cdecl* SendScoreV1)(const IRScoreV1& score) = nullptr;
+	const char*(OLR2_IR_API* GetName)() = nullptr;
+	bool(OLR2_IR_API* LoginV1)() = nullptr;
+	SendScoreStatus(OLR2_IR_API* SendScoreV1)(const IRScoreV1& score) = nullptr;
 };
-#ifndef _WIN32
-#undef __cdecl
-#endif // _WIN32
--- a/LR2/LR2_ir.cpp
+++ b/LR2/LR2_ir.cpp
@@ -1,7 +1,6 @@
 ﻿// LR2IR integration is deprecated. Please help us improve CustomIR instead.
 
 #include <string>
-#pragma comment(lib, "ws2_32.lib")
 #include "En_dbio.h"
 #include "En_fileutil.h"
 #include "En_timer.h"
@@ -14,6 +13,8 @@
 #include <DxLib/DxLib.h>
 
 #ifdef _WIN32
+#pragma comment(lib, "ws2_32.lib")
+
 #include <shellapi.h>
 #endif // _WIN32
 
@@ -293,7 +294,7 @@ static int ParseRivalData(int ID) {
 
 	CSTR hash, query;
 	int count = 0;
-	int clear, notes, combo, pg, gr, gd, bd, pr, minbp, option, lastupdate;
+	int clear{}, notes{}, combo{}, pg{}, gr{}, gd{}, bd{}, pr{}, minbp{}, option{}, lastupdate{};
 	SQL_Run("BEGIN", pRivalDB);
 
 	while (cur) {
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -185,3 +185,22 @@ add_executable(OpenLR2 WIN32 LR2/Main.cpp OpenLR2.manifest)
 set_target_properties(OpenLR2 PROPERTIES CXX_STANDARD 23)
 target_link_libraries(OpenLR2 PUBLIC OpenLR2Lib)
 set_target_properties(OpenLR2 PROPERTIES OUTPUT_NAME "OpenLR2${ARCH_SUFFIX}")
+
+set(IR_CONF_SUFFIX $<$<CONFIG:Debug>:_D>)
+if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+    set(IR_ARCH_SUFFIX ".x64")
+elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
+    set(IR_ARCH_SUFFIX ".x86")
+else()
+    message(FATAL_ERROR "Unsupported CMAKE_SIZEOF_VOID_P")
+endif()
+
+add_library(ExampleIR SHARED ExampleIR/ExternalIR.cpp)
+set_target_properties(ExampleIR PROPERTIES CXX_STANDARD 23)
+# For LR2_customir_api.h.
+# If you are making an IR module of your own, you should copy this file into your project instead.
+target_include_directories(ExampleIR PUBLIC LR2)
+set_target_properties(
+    ExampleIR
+    PROPERTIES OUTPUT_NAME "ExampleIR${IR_CONF_SUFFIX}${IR_ARCH_SUFFIX}"
+)
--- a/ExampleIR/README.md
+++ b/ExampleIR/README.md
@@ -1,8 +1,16 @@
 # CustomIR API
+
 CustomIR allows sending detailed score data to third-party score-tracking servers, like Bokutachi.
 
-The modules are loaded from ``LR2files/CustomIRs/<ModuleFolder>/``.
+The modules are loaded from `LR2files/CustomIRs/<ModuleFolder>/`.
 
-As a module developer, keep in mind that depending on game build configuration the naming convention for .dll files varies. File extension must be prefixed with architecture indicator, and file name must be followed by "_D" for debug configuration. For example: "ExternalIR.x86.dll", "ExternalIR_D.x64.dll".
+As a module developer, keep in mind that depending on game build configuration the naming convention for .dll files varies.
+File extension must be prefixed with architecture indicator, and file name must be followed by "\_D" for debug configuration.
+For example: "ExternalIR.x86.dll", "ExternalIR_D.x64.dll".
 
 Further information can be seen in [ExternalIR.cpp](ExternalIR.cpp) example and its comments.
+
+## Existing modules
+
+- Bokutachi: <https://github.com/MatVeiQaaa/BokutachiIR_OpenLR2>
+
--- a/LR2/Main.cpp
+++ b/LR2/Main.cpp
@@ -397,7 +397,7 @@ int main(int argc, char** argv) {
 	}
 	{
 		std::error_code ec; // ignore errors
-		std::filesystem::path path("LR2files/CustomIRs");
+		std::filesystem::path path(fs::make_preferred("LR2files/CustomIRs").data());
 		std::filesystem::create_directories(path, ec);
 		gs.net.customIR.Initialize(path);
 	}
--- a/LR2/strclass.h
+++ b/LR2/strclass.h
@@ -1,5 +1,7 @@
 #pragma once
+#ifdef _MSC_VER
 #pragma warning(disable:4996)
+#endif
 #include <stdio.h>
 #include <malloc.h>
 typedef unsigned int uint;


ExampleIR/active.ini.example
```diff
@@ -0,0 +1,7 @@
+# Copy to LR2files/CustomIRs/active.ini
+# CustomIR DLLs load from LR2files/CustomIRs/ regardless of config.xml lr2ir (dream-pro only).
+# provider = display module for result-screen rank (GetResultRankV1); blocks until rank returns (with dream-pro hHandle).
+# Other loaded DLLs (e.g. BokutachiIR) POST in the background and do not block the result screen.
+# Must match GetName() from the DLL (not always the folder name).
+# If missing or wrong: rank fetch off (warning log); all modules still Login + background SendScore.
+provider=ExampleIR
```
chown2: I think keeping this in `LR2files/Config/openlr2-config.xml` would be more fitting

LR2/LR2_customir.h
```diff
@@ -22,6 +23,12 @@ class CustomIR {
 	bool Login();
 	SendScoreStatus SendScore(const IRScoreV1& score);
 
+	GetStatus GetResultRank(const IRScoreV1& score, IRRankResultV1& out);
+	GetStatus RestoreCachedRank(const char* songHash, IRRankResultV1& out);
+
+	[[nodiscard]] bool SupportsResultRank() const { return mMethods.GetResultRankV1 != nullptr; }
+	[[nodiscard]] bool SupportsRestoreCachedRank() const { return mMethods.RestoreCachedRankV1 != nullptr; }
+	[[nodiscard]] bool SupportsSendScoreV1() const { return mMethods.SendScoreV1 != nullptr; }
```
chown2: These methods are redundant, the checks inlined would be more readable

ExampleIR/ExternalIR.cpp
```diff
@@ -11,6 +13,240 @@ namespace State {
...
+
+    std::filesystem::path RankCachePath(std::string_view hash) {
+        const std::string key = CacheStorageKey(hash);
+        if (key.empty()) {
+            return {};
+        }
+        return State::path / "IR" / (key + ".json");
+    }
+
+    bool ParseJsonIntField(std::string_view content, std::size_t from, std::string_view key, int& value) {
```
chown2: We definitely don't want an in-house JSON parser. I think it's better to use tinyxml which we already have.

chown2: Oh, that's the example IR. Better make it very simple just to demonstrate the module interface. We want the least amount of code here as it's meant to be a skeleton for other IR modules (like yours).

LR2/LR2_customir.cpp
```diff
@@ -79,11 +84,170 @@ SendScoreStatus CustomIR::SendScore(const IRScoreV1& score) {
...
+	std::vector<int> finishedThreads;
+	for (const auto& [i, it] : std::views::enumerate(mSendThreads)) {
+		if (it.wait_for(std::chrono::milliseconds(0)) == std::future_status::ready) {
+			finishedThreads.push_back(static_cast<int>(i));
+		}
+	}
+	std::ranges::reverse(finishedThreads);
+	for (auto i : finishedThreads) {
+		mSendThreads.erase(mSendThreads.begin() + i);
+	}
```
chown2: This lost an important comment from old code: 
> // Deferred deletion because we need to keep the std::future for whatever reason  

I got real confused for a moment

LR2/LR2_customir.cpp
```diff
@@ -79,11 +84,170 @@ SendScoreStatus CustomIR::SendScore(const IRScoreV1& score) {
...
+}
+
+bool CUSTOMIR_MANAGER::AnyDisplayModuleSupports(bool (CustomIR::*pred)() const) const {
+	for (const auto& module : ResolveDisplayModules()) {
+		if ((module.get()->*pred)()) {
+			return true;
+		}
+	}
+	return false;
+}
```
chown2: Unneeded method, use `std::ranges::any_of` instead of it

LR2/LR2_customir.cpp
```diff
@@ -79,11 +84,170 @@ SendScoreStatus CustomIR::SendScore(const IRScoreV1& score) {
...
+	if (mActiveProvider.empty()) {
+		return {};
+	}
+	for (const auto& module : mModules) {
+		if (module->Name() == mActiveProvider) {
+			return { module };
+		}
+	}
+	return {};
+}
```
chown2: This helper is not needed too, check my syntax but you should be able to use something like this:
`std::ranges::find(mModules, mActiveProvider, [](const std::shared_ptr<CustomIR>& ir){ return ir->Name(); });

ExampleIR/active.ini.example
luxurya111: I was unaware that a dedicated openlr2-config.xml existed. I have applied it now.

LR2/LR2_customir.cpp
luxurya111: I missed the fact that the comments were removed during the modification! I am truly sorry about that.

ExampleIR/ExternalIR.cpp
luxurya111: I have written this in as much detail as possible to make it easier to understand compared to the previous explanation. I have also simplified the examples.

GOMazk:
is it okay to pull now?

chown2:
Please let me take a look at it first

LR2/LR2_customir_api.h
```diff
@@ -105,8 +106,46 @@ enum class SendScoreStatus: int {
...
+	Ok = 0,
+	Retry,
+	Fail,
+};
+
 struct MethodTable {
 	const char*(__cdecl* GetName)() = nullptr;
 	bool(__cdecl* LoginV1)() = nullptr;
 	SendScoreStatus(__cdecl* SendScoreV1)(const IRScoreV1& score) = nullptr;
+	GetStatus(__cdecl* GetResultRankV1)(const IRScoreV1& score, IRRankResultV1& out) = nullptr;
```
chown2: I don't think it needs to take in a 'score', so that it we can use the same function to get leaderboards for unplayed charts.
We could take a score in a std::optional, but it's probably not worth the effot and we should just await for SendScore to complete before requesting the leaderboard.

LR2/LR2_customir_api.h
```diff
@@ -105,8 +106,46 @@ enum class SendScoreStatus: int {
...
+	GetStatus(__cdecl* RestoreCachedRankV1)(const char* songHash, IRRankResultV1& out) = nullptr;
```
chown2:
```suggestion
	GetStatus(__cdecl* GetResultRank)(const IRScoreV1& score, IRRankResult& out) = nullptr;
	GetStatus(__cdecl* RestoreCachedRank)(const char* songHash, IRRankResult& out) = nullptr;
```

This was a mistake in previous methods IMO. We can name them like 'GetResultRank2' or 'GetResultRankV2' once we do have to add the new version.
See, for reference, the naming scheme of sqlite.

LR2/LR2_customir.h
```diff
@@ -40,10 +48,24 @@ class CUSTOMIR_MANAGER {
...
+	[[nodiscard]] bool ShouldMirrorLegacyRankToMybest() const { return !ProvidesResultRank(); }
 private:
+	void EnqueueSidecarSend(const IRScoreV1& scoreV1, std::vector<std::shared_ptr<CustomIR>> sidecarModules);
+	[[nodiscard]] static bool SendScoreWithRetry(const std::shared_ptr<CustomIR>& module, const IRScoreV1& scoreV1);
+	[[nodiscard]] std::vector<std::shared_ptr<CustomIR>> ResolveSidecarModules() const;
+
 	std::vector<std::shared_ptr<CustomIR>> mModules;
 	std::vector<std::future<void>> mSendThreads;
+	std::future<void> mResultIrFuture;
+	CSTR mActiveProvider;
```
chown2:
```suggestion
	std::string mActiveProvider;
```
Don't use CSTR, it's horrible legacy

LR2/LR2_customir.cpp
```diff
@@ -443,41 +695,91 @@ IRScoreInternal::IRScoreInternal(game& game, sqlite3* sql, int _player) {
...
 			}
 
-			const auto sleepFor = static_cast<int>(std::pow(4, tryCount));
-			std::this_thread::sleep_for(std::chrono::seconds(sleepFor));
-			tryCount++;
-		}
-	}, IRScoreInternal{ game, sql, player }, mModules));
+			if (gotResult) {
+				ApplyIrRankResult(*gamePtr, curSong, merged, IrRankApplyContext::Result);
+			}
```
chown2: All of that seems unnecessary
```suggestion
			(void)SendScoreWithRetry(provider, scoreV1);
			IRRankResultV1 out{};
			const GetStatus status = provider->GetResultRank(scoreV1, out);
			if (status == GetStatus::Fail) {
				OverlayNotification("'%s' failed to get result rank\n", provider->Name().c_str());
				return;
			}
			ApplyIrRankResult(*gamePtr, curSong, merged, IrRankApplyContext::Result);
```

LR2/LR2_customir.cpp
```diff
@@ -443,41 +695,91 @@ IRScoreInternal::IRScoreInternal(game& game, sqlite3* sql, int _player) {
...
 
-			for (const auto& [i, t] : std::views::enumerate(sendThreads) | std::views::reverse) {
-				switch (t.get()) {
-				case SendScoreStatus::Fail:
-					OverlayNotification("'%s' failed to submit score\n", toSend[i]->Name().c_str());
-					[[fallthrough]];
-				case SendScoreStatus::Ok: toSend.erase(toSend.begin() + i); break;
-				case SendScoreStatus::Retry: break;
+	mResultIrFuture = std::async(std::launch::async,
+		[this, provider = *displayIt, scoreV1, curSong, gamePtr = &game]() {
```
chown2: When using std::async, pass data as function arguments instead of lambda captures. Here and in other places too.

LR2/LR2_customir.cpp
```diff
@@ -212,10 +321,153 @@ struct IRScoreInternal {
...
+			if (ctx == IrRankApplyContext::Result) {
+				best.IRclearRate = (rd.rankingCount + rd.clearPlayers[1] - rd.clearPlayers[0]) / rd.rankingCount;
+			} else {
+				best.IRclearRate = (rd.clearPlayers[2] + rd.clearPlayers[3] + rd.clearPlayers[4] + rd.clearPlayers[5]) * 100 / rd.rankingCount;
+			}
+		}
+	}
+}
+
+bool CUSTOMIR_MANAGER::SendScoreWithRetry(const std::shared_ptr<CustomIR>& module, const IRScoreV1& scoreV1) {
```
chown2: Don't take in pointers or smart pointers when a reference would suffice
```suggestion
bool CUSTOMIR_MANAGER::SendScoreWithRetry(const CustomIR& ir, const IRScoreV1& scoreV1) {
```

LR2/LR2_customir.h
```diff
@@ -1,27 +1,35 @@
...
 
 #include <wtypes.h>
 
 struct game;
 struct sqlite3;
 
+class CUSTOMIR_MANAGER;
+
 class CustomIR {
+	friend class CUSTOMIR_MANAGER;
```
chown2: Don't break encapsulation
```suggestion
```

LR2/LR2_customir.cpp
```diff
@@ -443,41 +714,77 @@ IRScoreInternal::IRScoreInternal(game& game, sqlite3* sql, int _player) {
...
+
+	IRRankResult out{};
+	const GetStatus status = (*displayIt)->RestoreCachedRank(entry.hash.body, out);
+	if (status == GetStatus::Fail) {
+		OverlayNotification("'%s' failed to restore cached rank\n", (*displayIt)->Name().c_str());
+		return;
+	}
+	if (status == GetStatus::Ok && HasIrRankPayload(out)) {
+		ApplyIrRankResult(game, curSong, out, IrRankApplyContext::SongSelectRestore);
+	}
```
chown2:
```suggestion
	ApplyIrRankResult(game, curSong, out, IrRankApplyContext::SongSelectRestore);
```

Like in ResultIrAsync, I don't see the need for existence of HasIrRankPayload.
And only now I've noticed, what about GetStatus::Retry?

LR2/LR2_customir_api.h
```diff
@@ -105,8 +106,46 @@ enum class SendScoreStatus: int {
...
+	int minbp{};
+	int option{};
+	int playcount{};
+	int ranking{};
+};
+
+struct IRRankResult {
+	int myRank{};
+	int totalPlayer{};
+	std::string lastupdate;
```
chown2:
```suggestion
	uint64_t lastupdate; // Seconds since Unix Epoch
```
Let's avoid ambiguity. Also, Unix Epoch matches `r_lastupdate FROM rival`.

chown2:
Do you find GetResultRank equally as usable with it taking song/dan hash instead of the whole score? Were you initially planning to have a separate method for fetching leaderboards from select? If so, how would they differ?

luxurya111:
Adding "IRScoreV1" was closer to just putting it into a variable simply to include the hash. I didn't keep in mind that only the hash would be used when I put IRScoreV1 in. Therefore, proceeding with hash-only is perfectly fine.

What I had considered was giving the option to fetch data via API requests in addition to the local cache. After reading your comment just now, I tried proceeding with only the API without saving the cache, but it was not at a recommendable level. So, if we are to modify it, instead of designing a separate dedicated method for the leaderboard, I think it would be fine to go ahead by adding an explanation in ExternalIR.cpp to force the use of the local cache.

Since the current code focuses on showing the IR ranking numbers, if we proceed with the leaderboard PR in the future, I will go with a plan that does not design a dedicated method for this.

To avoid confusion, I will attach an explanation for this to the docs commit.

chown2:
I have rebased this as there were conflicts. Please do `git fetch your-remote-name-here && git reset --hard your-remote-name-here/custom-ir/ir-ranking-count` before proceeding with changes.
I have also modified ExampleIR to make the skeleton more compact.

LR2/LR2_ir.cpp
```diff
@@ -650,6 +651,14 @@ void NETWORK::WaitAndInitRanking() {
...
 }
 
+bool NETWORK::IsResultIrPending() const {
+	return customIR.IsResultIrPending() || (isOnline && hHandle.joinable());
+}
+
+void NETWORK::BeginResultIr(game& g, sqlite3 *sql, int player) {
+	customIR.BeginResultIr(g, sql, player);
+}
+
```
chown2: Let's inline these function to where they are used. Pass-through helpers just make things more complex to read IMO.

LR2/LR2_customir.cpp
```diff
@@ -96,19 +101,85 @@ bool CustomIR::Login() {
...
 }
 
-SendScoreStatus CustomIR::SendScore(const IRScoreV1& score) {
+SendScoreStatus CustomIR::SendScore(const IRScoreV1& score) const {
 	if (mMethods.SendScoreV1 == nullptr) return SendScoreStatus::Fail;
 	return mMethods.SendScoreV1(score);
 }
 
+GetStatus CustomIR::GetResultRank(const char* songHash, IRRankResult& out) {
+	if (mMethods.GetResultRank == nullptr) return GetStatus::Ok;
```
chown2:
```suggestion
	if (mMethods.GetResultRank == nullptr) return GetStatus::Fail;
```

LR2/LR2_customir.cpp
```diff
@@ -96,19 +101,85 @@ bool CustomIR::Login() {
...
+	return mMethods.GetResultRank(songHash, out);
+}
+
+GetStatus CustomIR::RestoreCachedRank(const char* songHash, IRRankResult& out) {
+	if (mMethods.RestoreCachedRank == nullptr) return GetStatus::Ok;
```
chown2:
```suggestion
	if (mMethods.RestoreCachedRank == nullptr) return GetStatus::Fail;
```

LR2/LR2_customir.cpp
```diff
@@ -96,19 +101,85 @@ bool CustomIR::Login() {
...
+
+	mSendThreads.push_back(std::async(
+		std::launch::async,
+		&CUSTOMIR_MANAGER::SidecarSendAsync,
+		scoreV1,
+		std::move(sidecarModules)));
+}
+
+void CUSTOMIR_MANAGER::Initialize(const std::filesystem::path& directory, const CSTR& activeProvider) {
+	mActiveProvider = activeProvider.body ? activeProvider.body : "";
```
chown2: Let's avoid CSTR completely in new interafces

```suggestion
void CUSTOMIR_MANAGER::Initialize(const std::filesystem::path& directory, std::string activeProvider) {
	mActiveProvider = std::move(activeProvider);
```
And modify the header

LR2/LR2_customir.h
```diff
@@ -20,10 +22,26 @@ class CUSTOMIR_MANAGER {
...
 	void Login();
-	void SendScore(game& game, sqlite3* sql, int player);
+	void BeginResultIr(game& game, sqlite3* sql, int player);
+	void OnSongSelectRestoreRank(game& game);
+	[[nodiscard]] bool IsResultIrPending() const;
+	[[nodiscard]] bool IsProviderLoggedIn() const { return mProviderLoggedIn; }
+	[[nodiscard]] bool HasLoadedModules() const { return !mModules.empty(); }
+	[[nodiscard]] bool ProvidesResultRank() const;
+	[[nodiscard]] bool ProvidesCachedRankRestore() const;
+	[[nodiscard]] bool ShouldMirrorLegacyRankToMybest() const { return !ProvidesResultRank(); }
```
chown2: I understand this helper carries some intent, but, like with my previous comment about NETWORK::BeginResultIr, IMO such helpers needlessly make things more complex

LR2/Scene02_Songselect.cpp
```diff
@@ -1802,9 +1802,12 @@ int ProcS_Select(game *g) {
...
 					g->sSelect.bmsList[g->sSelect.cur_song].banner));
 	}
-	if (g->net.isOnline && g->procSelecter == 2) {
-		g->net.WaitAndInitRanking();
-		if (g->sSelect.bmsList[g->sSelect.cur_song].keymode >= 5 && (g->sSelect.bmsList[g->sSelect.cur_song].courseStageCount < 1 || g->sSelect.bmsList[g->sSelect.cur_song].courseIR)) {
+	if (g->procSelecter == 2) {
+		if (g->net.customIR.ProvidesCachedRankRestore()) {
+			g->net.customIR.OnSongSelectRestoreRank(*g);
+		}
+		if (g->net.isOnline && g->sSelect.bmsList[g->sSelect.cur_song].keymode >= 5 && (g->sSelect.bmsList[g->sSelect.cur_song].courseStageCount < 1 || g->sSelect.bmsList[g->sSelect.cur_song].courseIR)) {
```
```suggestion
		if (g->net.customIR.ProvidesCachedRankRestore()) {
			g->net.customIR.OnSongSelectRestoreRank(*g);
		} else if (g->net.isOnline && g->sSelect.bmsList[g->sSelect.cur_song].keymode >= 5 && (g->sSelect.bmsList[g->sSelect.cur_song].courseStageCount < 1 || g->sSelect.bmsList[g->sSelect.cur_song].courseIR)) {
```
so that LR2IR result doesn't overwrite CustomIR result

LR2/LR2_customir.h
```diff
@@ -20,10 +22,26 @@ class CUSTOMIR_MANAGER {
...
+	[[nodiscard]] static bool SendScoreWithRetry(const CustomIR& ir, const IRScoreV1& scoreV1);
+	static void SidecarSendAsync(IRScoreV1 scoreV1, std::vector<std::shared_ptr<CustomIR>> modules);
+	static void ResultIrAsync(std::shared_ptr<CustomIR> provider, IRScoreV1 scoreV1, int curSong, game* game);
+	[[nodiscard]] std::vector<std::shared_ptr<CustomIR>> ResolveSidecarModules() const;
+
 	std::vector<std::shared_ptr<CustomIR>> mModules;
 	std::vector<std::future<void>> mSendThreads;
+	std::future<void> mResultIrFuture;
+	std::string mActiveProvider;
+	bool mProviderLoggedIn = false;
```
chown2: This isn't used anywhere in this PR and shouldn't be used anyway I think. Each IR can have different login semantics and we shouldn't choose for them. Let's expect customIR modules to do the checks themselves.
So remove mProviderLoggedIn, it's getter and it's setting Login().

LR2/LR2_customir.h
chown2: I understand this helper carries some intent, but, like with my previous comment about NETWORK::BeginResultIr, IMO such helpers needlessly make things more complex

chown2:
What's your general plan of future work? What other methods are we potentially going to have? Looking at LR2IR end-points, with this PR we will only be missing ghosts.

LR2/LR2_customir.h
```diff
@@ -20,10 +22,26 @@ class CUSTOMIR_MANAGER {
...
+	void Initialize(const std::filesystem::path& directory, const CSTR& activeProvider);
 	void Login();
-	void SendScore(game& game, sqlite3* sql, int player);
+	void BeginResultIr(game& game, sqlite3* sql, int player);
+	void OnSongSelectRestoreRank(game& game);
+	[[nodiscard]] bool IsResultIrPending() const;
+	[[nodiscard]] bool IsProviderLoggedIn() const { return mProviderLoggedIn; }
+	[[nodiscard]] bool HasLoadedModules() const { return !mModules.empty(); }
+	[[nodiscard]] bool ProvidesResultRank() const;
+	[[nodiscard]] bool ProvidesCachedRankRestore() const;
```
chown2: Do they really matter? How about we call the methods unconditionally so we just get a 'Fail' using already existing logic if they aren't present

LR2/LR2_customir.h
```diff
@@ -20,10 +22,26 @@ class CUSTOMIR_MANAGER {
...
 	~CUSTOMIR_MANAGER();
-	void Initialize(const std::filesystem::path& directory);
+	void Initialize(const std::filesystem::path& directory, const CSTR& activeProvider);
 	void Login();
-	void SendScore(game& game, sqlite3* sql, int player);
+	void BeginResultIr(game& game, sqlite3* sql, int player);
+	void OnSongSelectRestoreRank(game& game);
+	[[nodiscard]] bool IsResultIrPending() const;
+	[[nodiscard]] bool IsProviderLoggedIn() const { return mProviderLoggedIn; }
+	[[nodiscard]] bool HasLoadedModules() const { return !mModules.empty(); }
```
chown2:
```suggestion
```
Not used anywhere

LR2/LR2_ir.cpp
luxurya111: I will review the code thoroughly overall and simplify similar code segments as much as possible, not just this specific part.

LR2/LR2_customir.h
luxurya111: I have updated the OnSongSelectRestoreRank function to reflect these changes.

LR2/Scene02_Songselect.cpp
luxurya111: Since it was changed to no longer use ProvidesCachedRankRestore, the logic for that part has also been modified.

LR2/LR2_customir.h
luxurya111: It might be a bit confusing, but should we only remove mProviderLoggedIn?
Modifying mResultIrFuture and mActiveProvider is fine, but based on the explanation, it seems like we just need to remove mProviderLoggedIn, IsProviderLoggedIn, and the mProviderLoggedIn inside Login().

luxurya111:
First of all, this PR implements the basic rank retrieval feature after enabling the use of custom_provider. Once this PR is approved, the features to be implemented next are as follows:

1. The leaderboard and comparing records with rivals within the leaderboard.
- It uses stored cache
2. Transmitting ghosts and Ghost Battle using the specific Ghost's random seed
- In order to transmit Ghosts, we may need to implement a separate IRScoreV2.
3. Checking the activation of the isOnline variable, prioritizing CustomIR.
- Currently, the isOnline status is verified through dream-pro communications, causing some skins dependent on this variable to fail to display ranks. Therefore, the activation of this variable will now be determined by the CustomIR Login status.
4. Linking the IR page that opens when pressing F5 to the provider's IR page.
- When isOnline is changed to 1 due to CustomIR, it will be modified to connect to the provider's IR page.

First of all, the overarching premise of the current work is to keep the LR2IR code intact while making modifications (as I cannot be certain whether it's safe to completely remove it yet).
Therefore, for CustomIR users, we will give priority to CustomIR first, and then proceed to execute the LR2IR logic (even though the site is defunct).

LR2/LR2_customir.cpp
```diff
@@ -126,9 +183,19 @@ void CUSTOMIR_MANAGER::Initialize(const std::filesystem::path& directory) {
...
+			"CustomIR: network/customir_provider '%s' not found; display fetch disabled, SendScore still active\n",
+			mActiveProvider.c_str()
+		);
+	}
 }
 
 void CUSTOMIR_MANAGER::Login() {
+	if (mModules.empty()) {
+		return;
+	}
```
chown2:
```suggestion
```

LR2/LR2_customir.h
chown2: all good

chown2:
> In order to transmit Ghosts, we may need to implement a separate IRScoreV2.

If all we want is to add a new field, then we could add it to the end of IRScoreV1. Since we pass it by reference into IR modules that won't break the ABI (and tbh we may not even care about compatibility this instant since only Bokutachi IR module exists).

chown2:
I've given another look at LR2IR, and we will also need a rival API in the future

chown2:
I'm implementing this API for BokutachiIR_OpenLR2 to test drive it https://github.com/MatVeiQaaa/BokutachiIR_OpenLR2
This is nearing a good state

MatVeiQaaa:
Btw the reason score struct lacks ghost is not because it was forgotten but because it requires a refactor of when and how the ghost string is made. The way it is right now, it only exists until the first deserialisation.

luxurya111:
@chown2 Initially, I also considered expanding V1, but I thought it would be better to separate it into V2 to ensure compatibility with existing IRs (Bokutachi is the only one I have verified so far).
If expanding within V1 is acceptable, I will proceed with that approach.
In my current compatibility tests, I am using Bokutachi_OpenLR2 and the Stellaverse IR under development simultaneously. (I have sent (^^) songs multiple times.)

Also, if you are referring to the rival feature that used to be in LR2IR, yes, that's correct. It will be needed in the future. However, I haven't looked into the rival-related APIs in detail yet. I didn't use LR2's rival feature very often, and I only vaguely recall creating the rival-specific DB, XML, and lr2folder in the early stages.

@MatVeiQaaa Although it is not currently included here, it seems that ghosts are being transmitted normally in my test version. If this part changes in the future, will there be any impact?

---

Additionally, I am thinking it might be a good idea to allow Custom IR to replace the API that fetches exlevel data as well. While this feature only supported Insane on LR2IR, I think it would be great if CustomIR could fetch songs other than Insane as well.

MatVeiQaaa:
> @MatVeiQaaa Although it is not currently included here, it seems that ghosts are being transmitted normally in my test version. If this part changes in the future, will there be any impact?

Ghost string deserialisation method is responsible for emptying the ghost buffer. So, only the first call to that method will give the ghost string. Resetting the buffer should be separated from that method and called after all IR operations are done. Note that LR2IR method runs on its own thread and uses global context, which means that ghost string must be passed as an argument instead to avoid context desync.
Using database value makes no sense for CustomIR api which intends to send actual score data of each score, rather than mybest like lr2ir. Yes, mybest ghost vs current ghost should also be considered.

luxurya111:
Understood.
I agree with proceeding to transmit ghostData and other scores limited to the current playthrough. Since reading mybest and transmitting it together can easily be implemented by whoever designs the Custom IR in the future, I think it is fine to leave that decision to the Custom IR developers.

chown2:
I will be doing the finishes touches myself for the next few days, so please tell me first if you want to change something yourself

LR2/LR2_customir_api.h
```diff
@@ -113,8 +115,76 @@ enum class SendScoreStatus: int {
...
+	int combo{};
+	int pg{};
+	int gr{};
+	int gd{};
+	int bd{};
+	int pr{};
+	int minbp{};
+	int option{};
+	int playcount{};
+	// TODO(stable-api): figure out if this makes sense. It's redundant by IRRankResult::ranking index.
```
chown2: TODO

chown2:
@luxurya111 please test this and review the PR.
Works for me with ExampleIR, but we'd better all try it, I assume you also have a Stellaway IR implementation too.
I didn't modify CustomIR API, but have changed the implementation considerably.

I have added notes that the API is 'experimental' for now, whatever that means.
I couldn't write a Tachi implementation because I didn't find an API to fetch scores by md5, there is only an API for raja which requires sha256 hashes.

LR2/LR2_configsave.cpp
```diff
@@ -583,6 +583,10 @@ int WriteOpenLr2ConfigXml(game *g, const char *filename){
 	WriteXML_Tab2Str(pFile, "courseresult", (g->config).skin.skinFilePath[15]);
 	fputs("\t</skin>\n", pFile);
 
+	fputs("\t<network>\n", pFile);
+	WriteXML_Tab2Str(pFile, "customir_provider", (g->config).network.customIrProvider);
```
chown2: customir_provider, mActiveProvider, main provider.
We need to choose a single consistent name.
If we consider 'provider' to be the one IR module whose display functions are used, then customir_provider is the best name of them.

MatVeiQaaa: Imo something like mDisplayIR is better. All around, including config and all, if one cares enough. It explains itself much better.

chown2:
epic github moment

chown2:
please continue in #143

PR #113, opened 2026-06-08
Fix autoreload song scan database handling

Skalim112:
## Summary

This PR fixes issues in the song autoreload scan flow related to `song.db` handling and chart parent directory assignment.

## Changes

* Fixes a scan loop caused by mutating `.db` entries while iterating over a `SELECT` result (First iteration, i am working on a more global solution since the SELECT iterates two times, one for obtain the data from the table "folder" and other for DELETE/INSERT, which is slow af).
* Restores the parent folder behavior of the original LR2 so charts are linked to the main folder registered in the jukebox instead of being linked incorrectly by their own parent (i saw the previous change, and i think this is more robust, your choice).

## Why

The autoreload scan could behave incorrectly when the database was modified during the scan process. This could cause repeated scan behavior and incorrect parent directory links for charts loaded from `song.db`.

## Testing

* New client, make sure you don´t have song.db (its still slow btw)
* Make sure you have at least 200 folders on your main song folder
* You will see its going to loop at least two times, checking for outdated "date" column using -JOURNAL, which, it seems, its an intended behaviour because the autocommit is secure.

### DIFF
--- a/LR2/LR2_songmanage.cpp
+++ b/LR2/LR2_songmanage.cpp
@@ -224,12 +224,6 @@ int SetUndefinedDifficulty(sqlite3 *sql) {
 	sqlite3_stmt *pStmt;
 	CSTR folder;
 	int mode, difficulty;
-	int shouldCommit;
-
-	shouldCommit = sqlite3_get_autocommit(sql);
-	if (shouldCommit && SQL_Run("BEGIN IMMEDIATE", sql) != 0) {
-		shouldCommit = 0;
-	}
 
 	sqlite3_prepare(sql, "SELECT difficulty,folder,mode,path FROM song ORDER BY folder,mode,karinotes", -1, &pStmt, NULL);
 	while (sqlite3_step(pStmt) == 100) {
@@ -256,9 +250,6 @@ int SetUndefinedDifficulty(sqlite3 *sql) {
 		}
 	}
 	sqlite3_finalize(pStmt);
-	if (shouldCommit) {
-		SQL_Run("COMMIT", sql);
-	}
 	return 1;
 }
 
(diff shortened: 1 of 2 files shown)


MatVeiQaaa:
https://github.com/GOMazk/OpenLR2/pull/111 already fixed wrong parent

GOMazk:
this way seems not bad

chown2:
This is only meant to help when a song.db has already been impacted by the getParentDirectory bug, but introduces non-zero overhead for normal cases too. 
Maybe it would have been possible to prune the DB of such parsed entries in one go so that they could later be rescanned normally. 
However, all of this code is to later go away anyway, since song scanning in LR2 is a poor buggy mess.

LR2/LR2_songmanage.cpp
```diff
@@ -1055,11 +1056,60 @@ int ChangeCourseID(sqlite3 *sql, int newID, int oldID, int type) {
...
+
+// Checks whether an existing song row can be kept without reparsing the BMS file.
+// The timestamp check keeps the fast path for unchanged charts, while the CRC checks
+// repair legacy rows created with a stale parent/folder relationship.
+int IsRegisteredSongCurrent(sqlite3 *sql, CSTR path, int filetime) {
+	sqlite3_stmt *pStmt = NULL;
+	char str[1024];
+	const char *folder;
+	const char *parent;
+	CSTR expectedFolder;
```
chown2, line 1085: In future PRs, please combine variable declarations and initializations in single statements. :) It is only written like this in existing code because that's how decompilation software show it (because that's how it gets transformed when the code is compiled ).
```suggestion
	const CSTR expectedFolder  = AssignCRC32(path.getDirectory());
```

Issue #103, opened 2026-06-06
Pressing the white and blue keys on the results screen restarts the song in 段位認定


FSRS:

FSRS:
It seems to return to (or skip to) the second song. The same issue also occurs in COURSE mode.

luxurya111:
Upon verification, while proceeding to the second stage is indeed a bug, the original intention was closer to allowing players to restart midway through a course regardless of their clear status.

It would be great to implement a separate GUI in the future to handle course restarts, but since that has not been implemented yet, I will limit my current modification strictly to fixing the issue where it starts from the second song.

chown2:
After that PR it's still possible to restart a course from song (not course) result. It correctly goes to the first stage though.

chown2:
I still think it shouldn't be allowed to restart the whole course from a single song's result

MatVeiQaaa:
> I still think it shouldn't be allowed to restart the whole course from a single song's result

Why not? Why should it be allowed to restart from within the song then?

luxurya111:
Starting over from the middle seems like a game changer. I think it shouldn't be allowed to restart from the N stage of a course. I think the intent of a course is generally designed on the premise of playing from the beginning without restarting in the middle.

chown2:
> I think it shouldn't be allowed to restart from the N stage of a course

@luxurya111  that's not the issue, holding white and blue keys on song result of stage 2 restarts you to stage 1, as it's supposed to

> Why not? Why should it be allowed to restart from within the song then?

@MatVeiQaaa 
I assume you mean 'why should it be allowed to quick restart a course using the play scene quick restart'.
In my opinion, these two actions carry different levels of intent.
Pressing start+select/esc is already destructive, once the fadeout starts you can not go back, you fail the course, and to quick-restart you have to press select or start once again on top of that.
It's not something you'd do accidentally.
Just holding keys on result is a much easier thing to do, and you won't expect it to restart anything since LR2 didn't have such behavior.
To quote @luxurya111 https://github.com/GOMazk/OpenLR2/pull/115#issuecomment-4663323373
> The reason I modified this was because, during my usual LR2 gameplay, I would unconsciously mash all the keys when transitioning to the next course stage

MatVeiQaaa:
@chown2 then the problem is not restarting from select scene in itself, but the interface to do that.

PR #115, opened 2026-06-09
fix: disable result QuickRestart for expert and class courses

luxurya111:
This code prevents QuickRestart from triggering during a course, which would otherwise cause the game to resume from the second stage. This results in the previous scores continuously accumulating, allowing players to obtain a much higher score than in a regular course playthrough.

Issue: https://github.com/GOMazk/OpenLR2/issues/103

### DIFF
--- a/LR2/Scene05_Result.cpp
+++ b/LR2/Scene05_Result.cpp
@@ -195,11 +195,6 @@ static void QuickRestart(game& game, bool newRandom) {
 	game.gameplay.flag_retry = newRandom ? 0 : 1;
 	game.gameplay.randomseed = newRandom ? 0 : game.gameplay.randomseed;
 
-	if ((game.gameplay.courseType == 0 || game.gameplay.courseType == 2) && game.gameplay.courseStageNow != 0) {
-		game.gameplay.courseStageNow = 0;
-		game.gameplay.flag_retry = 0;
-	}
-
 	if (game.gameplay.flag_retry == 0) ReleaseBGA(&game); // ugly place for this
 	StopAllKeysound(&game);
 }
@@ -296,7 +291,8 @@ int ProcI_Result(game *g) {
 				g->KeyInput.p2_buttonInput[1] == 2 || g->KeyInput.p2_buttonInput[3] == 2 || g->KeyInput.p2_buttonInput[5] == 2 || g->KeyInput.p2_buttonInput[7] == 2) 
 				&& (g->KeyInput.p1_buttonInput[2] == 2 || g->KeyInput.p1_buttonInput[4] == 2 || g->KeyInput.p1_buttonInput[6] == 2 ||
 					g->KeyInput.p2_buttonInput[2] == 2 || g->KeyInput.p2_buttonInput[4] == 2 || g->KeyInput.p2_buttonInput[6] == 2) 
-				&& g->gameplay.replay.status != 2 && g->config.play.m_lunaris == 0) {
+				&& g->gameplay.replay.status != 2 && g->config.play.m_lunaris == 0
+				&& g->gameplay.courseType != 0 && g->gameplay.courseType != 2) {
 
 				if (g->KeyInput.p1_buttonInput[2] == 2 || g->KeyInput.p2_buttonInput[2] == 2) QuickRestart(*g, true);
 				else QuickRestart(*g, false);


MatVeiQaaa:
it was intended to work on courses. If it doesnt work correctly it should be fixed, not removed.

MatVeiQaaa:
See what happens on result exit, it increments course index there. A hack-around would be setting courseStageNow to -1 instead of 0. It also sounds like the course stats must be reset.

luxurya111:
I understand that the QuickRestart feature in courses was an intentional design.

I also misidentified the core issue: it was actually a regular Restart, not a QuickRestart. (Doing a QuickRestart inside a course worked without any issues.)

I will withdraw the PR for now. Thank you for pointing that out.

luxurya111:
I apologize for the inconvenience of reopening the PR.

Upon verification, performing a QuickRestart works properly if a Game Over occurs in the middle of a song. The actual modification was made to Scene05, while Scene04 is designed to allow QuickRestart to function.

In other words, I should have explained it as preventing QuickRestart "strictly within the result screen" during a course, but that detail was missing. I wanted to clarify that the current PR contains code designed to ensure that pressing the button on the course result screen proceeds to the next song, aligned with the standard design of LR2.

Please take this into consideration and review the PR.

MatVeiQaaa:
yes, i was aware that the changes disable restarting from result scene, and it was put there intentionally as a feature. It should be fixed if something is wrong, not removed entirely. Although, maybe it makes sense to lock it out on stage result, but make the restart feature from course result (scene 13).

MatVeiQaaa:
> yes, i was aware that the changes disable restarting from result scene, and it was put there intentionally as a feature. It should be fixed if something is wrong, not removed entirely. Although, maybe it makes sense to lock it out on stage result, but make the restart feature from course result (scene 13).

It would still be weird. If we allow restarting courses at all, why not let the user restart them at all possible points? Why have user wait to hit a different scene if it could be done at stage result already, if the user wants to.

luxurya111:
The reason I modified this was because, during my usual LR2 gameplay, I would unconsciously mash all the keys when transitioning to the next course stage, which led me to believe this behavior was a bug. I am currently debugging this with a friend, and they also felt that this particular part felt awkward.

Therefore, when I saw this specification implemented in OpenLR2, I simply assumed it was intended to enable QuickRestart from a regular song's Game Over screen, and judged that triggering a QuickRestart midway through a course was an unintended design flaw.

If this design was indeed intentional, I am prepared to follow it. However, at that specific point where a player is supposed to move on to the next stage of the course, if someone like me is holding down keys and suddenly gets sent all the way back to the beginning of the course, it feels quite unnatural.

Aside from that, I think being able to QuickRestart when getting a Game Over midway through a course is actually pretty cool.

MatVeiQaaa:
i dont know then, this is opinion vs opinion situation. As middle ground i offer limiting course restart to scene 13, as described before. But then a player who wants to restart immediately upon hitting stage result, would have to wait to get to play scene (or course result scene, in case of fail) to restart from there. Restarting requires holding two buttons in a very deliberate manner, it's not something to do accidentally. If you got some specific muscle memory to clash with this, it's a unique problem.

A proper way of solving this problem would be implementing GUI to restart, instead of a button combination. Something like a 'START' panel. I want something like this for per-column judgement stats in the future.

luxurya111:
I think being able to operate it via a GUI would be a much better approach. I agree.

For now, looking strictly at the current situation, I agree with adding the restart logic in Scene13. Since this PR was not designed with that intention, I will go ahead and close it for now.
I think it would be best to share this matter via the issue link above first and gather some feedback on it.

To offer a bit of a compromise here, we could make it so that QuickRestart only triggers upon a FAILED status in Scene05. (Currently, QuickRestart only works upon a CLEAR status.) However, since this is also just my personal opinion, I think it would be best to gather feedback as I mentioned above.

Thank you so much for your detailed answer regarding this matter.

chown2:
IMO it makes no sense to restart courses from song result, especially only doing that after failing, as that is a logical inconsistency

luxurya111:
There was a brief moment of confusion, but I think it's just my old habit from playing LR2 causing the issue.

I mentioned gathering opinions earlier, but on second thought, I feel like this is something I can handle on my end with a bit more effort.

If it's possible to control it via a GUI, I think that would be a much better approach.

When I submit the next PR, I want to focus strictly on the phenomenon itself where the game starts from the second stage while incorrectly maintaining the current course score.

PR #116, opened 2026-06-10
perf(song.db - rescan): fast hierarchy check and in-place CRC repair in ReloadSongsByQuery

Skalim112:
## Summary

Fixes the second scan that occurs when a song.db is reinstalled on a new folder client, optimizing valuable time on scanning big folders.

## Changes

Two optimizations that work in synergy:

1. Fast check: The song query now includes folder and parent columns, so the CRCs from the initial SELECT are reused for IsRegisteredSongHierarchyCurrent without an extra SELECT per song (up to 115k fewer queries per rescan).

2. In-place repair: When the cached snapshot shows the file date is current but the folder/parent CRCs are outdated, the row is repaired with a single UPDATE instead of DELETE + ParseBMSMETA (open folder, analyze content, capture metadata, continue) + INSERT. This avoids reopening and rehashing the BMS file for rows that only have outdated CRCs, which is the common case when migrating from a DB built by the original LR2.

3. `LoadLR2CustomFolder` and the autoreload queries in Scene02_Songselect now select `folder,parent` so the fast path is always active.

4. Added English translations alongside the existing Japanese strings in `LoadLR2CustomFolder`'s ErrorLog/printf calls.

## Why

Let me give my understanding of how the scan flow works today:

* SELECT  path,date FROM song: Does a first SELECT to all the songs previously loaded on song.db
* chg = IsFileChanged(time, str, &newTime):  Check if a file has different date, which means the file could be changed by X or Y motive: It throws 0 if "No Changed", 1 if "The file doesn´t exist" or 2 if "The file has changes"
*  IsRegisteredSongCurrent(sql, str, newTime) = 0 → `SELECT date,folder,parent FROM song WHERE path=?`:  its executed when chg = 0, it verifies a second time if the parent on the "folder" table has changes, if this is correct, chg = 2, then proceeds to change the database.
* DELETE to the found changes, then insert with new changed (  DELETE FROM song WHERE path=?, open the file, then INSERT FROM...)
* +Does more selects based on all the folders with changes because the first SELECT was never cached so, it serves purely as a flag for every folder changed (O(N+1)) as "path" as the only cached variable for iterate. Since every call is independent, it iterates on every song, checking for updates that could possibly not being made on the first place.

## Testing

- Load an old song.db (you can start with a 300 song folder)
- initiate the game with AUTO RELOAD TYPE 2 on the client
- Wait for it to load the first scan

The expected behaviour is that the second SELECT should be almost instant

## Quick note

Yea, i know that i must not be working on this part of the code, but i want the first experience of this game to be the most fluid possible at first, because i know this could be a problem now, and i want to give my little support based on what i know, on what i see, and can be done in the future.

**THIS APPROACH IS MEAN TO BE ONLY FOR SONG.DB RELOAD**, for every other instance its still slow.

Charts whose `folder`/`parent` values were left stale by the bug fixed in #111 (or by any other past inconsistency) don't need a manual DB migration. The new "Pass 2" in `ReloadSongsByQuery` re-validates `folder`/`parent` CRCs for every unchanged song on the next reload and repairs them in place (`UPDATE` only, no reparse), so existing installs self-heal automatically the first time they run with this branch.

### DIFF
--- a/LR2/LR2_songmanage.cpp
+++ b/LR2/LR2_songmanage.cpp
@@ -2560,11 +2560,11 @@ int LoadLR2CustomFolder(sqlite3 *sql, CONFIG_JUKEBOX *jb, CSTR scoreDBpath, char
 
 		sqlite3_snprintf(256, query2, "ATTACH \'%q\' AS scoredb", scoreDBpath.body);
 		if (SQL_Run(query2, sql) != 0) {
-			ErrorLogAdd("スコアデータベースの接続に失敗しました。\n");
+			ErrorLogAdd("スコアデータベースの接続に失敗しました。 / Failed to attach the score database.\n");
 			return -1;
 		}
 		if (SQL_Run(fs::make_preferred("ATTACH \'LR2files/Database/tag.db\' AS tagdb").data(), sql) != 0) {
-			ErrorLogAdd("タグとかデータベースの接続に失敗しました。\n");
+			ErrorLogAdd("タグとかデータベースの接続に失敗しました。 / Failed to attach the tag database.\n");
 			return -1;
 		}
 	}
@@ -2573,7 +2573,7 @@ int LoadLR2CustomFolder(sqlite3 *sql, CONFIG_JUKEBOX *jb, CSTR scoreDBpath, char
 		sqlite3_stmt *pStmt;
 
 		SQL_Run("BEGIN", sql);
-		ErrorLogAdd("エラーフォルダの検索を行います。\n");
+		ErrorLogAdd("エラーフォルダの検索を行います。 / Searching for orphaned folders.\n");
 
 		sqlite3_snprintf(1024, query, "SELECT * FROM folder WHERE parent = \'%s\'", AssignCRC32("ROOT").body);
 		SQL_prepare(query, sql, &pStmt);
@@ -2589,12 +2589,12 @@ int LoadLR2CustomFolder(sqlite3 *sql, CONFIG_JUKEBOX *jb, CSTR scoreDBpath, char
 			}
 
 			if(is_path_not_in_jukebox) {
-				ErrorLogFmtAdd("エラーフォルダを削除します。%s\n", path.body);
+				ErrorLogFmtAdd("エラーフォルダを削除します。 / Removing orphaned folder: %s\n", path.body);
 				sqlite3_snprintf(1024, query, "DELETE FROM folder WHERE path = \'%s\'", path.body);
 				SQL_Run(query, sql);
 			}
 			else {
-				ErrorLogFmtAdd("ジュークボックスに登録されています。%s\n", path.body);
+				ErrorLogFmtAdd("ジュークボックスに登録されています。 / Registered in jukebox: %s\n", path.body);
 			}
 		}
 		sqlite3_finalize(pStmt);
@@ -2609,20 +2609,20 @@ int LoadLR2CustomFolder(sqlite3 *sql, CONFIG_JUKEBOX *jb, CSTR scoreDBpath, char
 		}
 
 		if (jb->autoreload == 2 || flag_starter) {
-			ErrorLogAdd("フォルダ更新チェックを行います。\n");
+			ErrorLogAdd("フォルダ更新チェックを行います。 / Checking folder updates.\n");
 			ReloadSongsByQuery("SELECT path,date FROM folder", sql, jb);
-			ErrorLogAdd("ファイル更新チェックを行います。\n");
+			ErrorLogAdd("ファイル更新チェックを行います。 / Checking file updates.\n");
 			ReloadSongsByQuery("SELECT path,date FROM song", sql, jb);
 		}
 		else {
-			ErrorLogAdd("フォルダ更新チェック(ルートのみ)を行います。\n");
+			ErrorLogAdd("フォルダ更新チェック(ルートのみ)を行います。 / Checking folder updates (root only).\n");
 			sqlite3_snprintf(1024, query, "SELECT path,date FROM folder WHERE parent = \'%s\' OR date = 0", AssignCRC32("ROOT").body);
 			ReloadSongsByQuery(query, sql, jb);
 		}
 
-		ErrorLogAdd("未定義#DIFFICULTYの自動設定を行います。\n");
+		ErrorLogAdd("未定義#DIFFICULTYの自動設定を行います。 / Auto-assigning undefined #DIFFICULTY.\n");
 		SetUndefinedDifficulty(sql);
-		ErrorLogAdd("データベースチェックは終了しました。\n");
+		ErrorLogAdd("データベースチェックは終了しました。 / Database check finished.\n");
 
 		if (flag_starter == 0) {
 			SQL_Run(fs::make_preferred("DELETE FROM folder WHERE path LIKE \'LR2files/Rival/%\'").data(), sql);
--- a/LR2/LR2_songmanage.cpp
+++ b/LR2/LR2_songmanage.cpp
@@ -4,6 +4,7 @@
 #include "filesystem.h"
 #include "filesystem.h"
 #include <iterator>
+#include <ranges>
 #include <vector>
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


Skalim112:
I tested this approach with a folder with 16.600 charts, each one with parent corrupted:

- Time invested fixing parent on old .exe: 1306.6s (~21.8 min)** (2574ms → 1309138ms)
- Time invested fixing parent with this change: ~1.53s** (3122ms → 4652ms) (B R U H)

LR2/LR2_songmanage.cpp
```diff
@@ -1073,22 +1073,29 @@ int GetRegisteredDateFromPath(const char *table, sqlite3 *sql, CSTR path) {
 	return date;
 }
 
+// Validates folder and parent CRCs from already-fetched values, without SQL.
+// Used by ReloadSongsByQuery when the snapshot already includes folder/parent columns.
+int IsRegisteredSongHierarchyCurrent(CSTR path, CSTR folder, CSTR parent) {
```
```suggestion
static int IsRegisteredSongHierarchyCurrent(CSTR path, CSTR folder, CSTR parent) {
```
Use 'static' on all new functions unless other .cpp files use them

LR2/LR2_songmanage.cpp
```diff
@@ -1073,22 +1073,29 @@ int GetRegisteredDateFromPath(const char *table, sqlite3 *sql, CSTR path) {
...
+		&& expectedFolder.isSame(folder.body) && expectedParent.isSame(parent.body);
+}
+
 // Checks whether an existing song row can be kept without reparsing the BMS file.
 // The timestamp check keeps the fast path for unchanged charts, while the CRC checks
 // repair legacy rows created with a stale parent/folder relationship.
 int IsRegisteredSongCurrent(sqlite3 *sql, CSTR path, int filetime) {
+	// folder is the physical chart directory; parent is where the chart appears in select.
+	CSTR expectedFolder = AssignCRC32(path.getDirectory());
+	CSTR expectedParent = AssignCRC32(path.getParentDirectory());
```
chown2: Assuming this was done in response to a previous comment of mine: this should be applied to all local variables

LR2/LR2_songmanage.cpp
```diff
@@ -1238,14 +1245,23 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb) {
...
 	std::vector<CSTR> pathList;
 	std::vector<int> timeList;
+	// folderList/parentList populated when query selects 4+ columns.
+	// Enables fast in-memory CRC check and in-place repair without reopening BMS files.
+	std::vector<CSTR> folderList;
+	std::vector<CSTR> parentList;
 
 	GetTimeWrap();
 	SQL_prepare(query, sql, &pStmt);
+	const bool hasSongHierarchy = sqlite3_column_count(pStmt) >= 4;
```
chown2: If all queries now have these columns, remove the 'slow' path

LR2/LR2_songmanage.cpp
```diff
@@ -1266,9 +1282,28 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb) {
 			if (jb->numOfPath == 0 || is_path_in_jukebox){
 				int newTime;
 				int chg = IsFileChanged(time, str, &newTime);
-				// IsFileChanged only compares timestamps; force reinsert when hierarchy CRCs are stale.
-				if (chg == 0 && is_bms_file && IsRegisteredSongCurrent(sql, str, newTime) == 0) {
-					chg = 2;
+				// IsFileChanged only compares timestamps; also validate hierarchy CRCs.
```
chown2: Please, move this into a separate loop. Yes, it will query the filesystem for each file's mtime twice instead of once, but it would be much clearer.

chown2:
> I tested this approach with a folder with 16.600 charts, each one with parent corrupted:

Those are some nice numbers. I believe them, but keep in mind the existence of filesystem caching. Running this for the second time before rebooting should be a 100 times faster.

LR2/LR2_songmanage.cpp
chown2: And add comment before it: 
> fix song.db affected by previous OpenLR2 bug where 'parent' became the same as 'folder'. https://github.com/GOMazk/OpenLR2/ pull/111

MatVeiQaaa:
i believe that having code to fix the result of a bug is completely unnecessary. The cost of fix is a database rebuild, and this will become dead code almost instantly.

LR2/LR2_songmanage.cpp
Skalim112: Yes, also i am actually modifying the scopes i am covering, and i think i did not miss anything this time

LR2/LR2_songmanage.cpp
Skalim112: Suggerence applied on the new commit, tested and it works

LR2/LR2_songmanage.cpp
Skalim112: Applied and tested, there was no perfomance drop at all, check the new commit

Skalim112: also, sorry for my ignorance, but where do i put that comment?

Skalim112:
Changes applied

> i believe that having code to fix the result of a bug is completely unnecessary. The cost of fix is a database rebuild, and this will become dead code almost instantly.

If this helps for today, i am happy with it since this could be a little starting point of something bigger.

Skalim112:
[Log.txt](https://github.com/user-attachments/files/28800224/Log.txt) forgot to save the log of the old iteration, but here is one with the actual changes and suggerences applied, also added a log for count the files changed on a non-cached db (corrupted 16604 data, then restarted the pc, all seems to work in-game)

LR2/LR2_songmanage.cpp
chown2: This bool and the branches have not been removed

chown2:
Thank you for your effort. After thinking on @MatVeiQaaa's comment, I think we should revert #113 and instead add a way to reset the song db. I think a command line argument is the only place it fits right now. Then, into that we could plug in a ~10 liner to detect if there are entries in the DB matching `select 1 from song where folder = parent;` and reset the DB forcibly. A little reggression undoable by a proper rescan is not worth all of this additional code.

MatVeiQaaa:
alternatively we can let it be for now, and remove this chunk of code say a month or two later, when it will be almost guaranteed to be dead code.

LR2/LR2_songmanage.cpp
```diff
@@ -1330,8 +1345,33 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb) {
 			}
 		}
 	}
+
+	// Pass 2: repair stale folder/parent CRCs on songs that are otherwise unchanged.
+	// Re-checks IsFileChanged so rows already deleted/reinserted by pass 1 are skipped.
```
```suggestion
	// Repair stale folder/parent CRCs on songs that are otherwise unchanged.
	// Fixes song.dbs affected by previous OpenLR2 bug where 'parent' became the same as 'folder'. https://github.com/GOMazk/OpenLR2/pull/111
```

LR2/LR2_songmanage.cpp
```diff
@@ -1233,22 +1240,33 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb) {
...
 		timeList.push_back(sqlite3_column_int(pStmt, 1));
+		if (hasSongHierarchy) {
+			folderList.push_back(SQL_GetColumn(2, pStmt));
+			parentList.push_back(SQL_GetColumn(3, pStmt));
+		}
 	}
 	sqlite3_finalize(pStmt);
 
+	// Pass 1: classify each row as unchanged/missing/changed and apply
+	// missing chart (chg==1) and outdated charts (chg==2) updates.
```
```suggestion
```

LR2/LR2_songmanage.cpp
```diff
@@ -1233,22 +1240,33 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb) {
...
 	sqlite3_stmt *pStmt;
 	char sBuf[1024];
-	int cAlready = 0, cNot = 0, cChange = 0;
+	int cAlready = 0, cNot = 0, cChange = 0, cRepaired = 0;
 	// This query often reads song/folder, and the branches below can delete or insert
 	// into those same tables. Snapshot first, then finalize the SELECT before mutating.
 	std::vector<CSTR> pathList;
 	std::vector<int> timeList;
+	// folderList/parentList populated when query selects 4+ columns.
+	// Enables fast in-memory CRC check and in-place repair without reopening BMS files.
```
```suggestion
```

LR2/LR2_songmanage.cpp
```diff
@@ -1233,22 +1240,33 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb) {
 
 	sqlite3_stmt *pStmt;
 	char sBuf[1024];
-	int cAlready = 0, cNot = 0, cChange = 0;
+	int cAlready = 0, cNot = 0, cChange = 0, cRepaired = 0;
```
```suggestion
	int cAlready = 0, cNot = 0, cChange = 0;
```
move it right before the loop that increments this number and checks it

Skalim112: ok, moving it to before the loop on the next commit

LR2/LR2_songmanage.cpp
Skalim112: My bad, hasSongRedundancy is gone now

LR2/LR2_songmanage.cpp
```diff
@@ -1246,9 +1255,13 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb) {
 	while (sqlite3_step(pStmt) == 100) {
 		pathList.push_back(SQL_GetColumn(0, pStmt));
 		timeList.push_back(sqlite3_column_int(pStmt, 1));
+		folderList.push_back(sqlite3_column_count(pStmt) >= 4 ? SQL_GetColumn(2, pStmt) : CSTR(""));
+		parentList.push_back(sqlite3_column_count(pStmt) >= 4 ? SQL_GetColumn(3, pStmt) : CSTR(""));
```
chown2: ?
```suggestion
		folderList.push_back(SQL_GetColumn(2, pStmt));
		parentList.push_back(SQL_GetColumn(3, pStmt));
```

chown2: g->sSelect.unk4fa4[1]  and g->sSelect.unk4fa4[3] were not modified accordingly

chown2: nevermind, that's not the approach we should be taking

chown2 (changes requested):

chown2 (changes requested):
Thanks to the code resturcturing by moving into a separate loop, we can see that fixing up 'parent' doesn't have to be tied to ReloadSongsByQuery at all. :)
Make a separate function. You can call it from ReloadSongsByQuery. Collect paths to update using `SELECT path FROM song WHERE folder = parent;` and `SELECT path FROM folder;` (will have to go through all of them I guess, maybe we can only do that if we did find some rows in 'song'? I imagine this must be expensive to fetch all of them from the DB).
Apply the 'Repair stale folder/parent CRCs on songs that are otherwise unchanged' loop on them. You don't have to update 'folder' in there, as only 'parent' is affected.

Skalim112:
> Thanks to the code resturcturing by moving into a separate loop, we can see that fixing up 'parent' doesn't have to be tied to ReloadSongsByQuery at all. :) Make a separate function. You can call it from ReloadSongsByQuery. Collect paths to update using `SELECT path FROM song WHERE folder = parent;` and `SELECT path FROM folder;` (will have to go through all of them I guess, maybe we can only do that if we did find some rows in 'song'? I imagine this must be expensive to fetch all of them from the DB). Apply the 'Repair stale folder/parent CRCs on songs that are otherwise unchanged' loop on them. You don't have to update 'folder' in there, as only 'parent' is affected.

Suggerence applied, but with a slight distinction:

RepairSongHierarchy(sql, pathList, timeList, folderList, parentList)` now lives next to `IsRegisteredSongHierarchyCurrent`, and `ReloadSongsByQuery` just calls it with the snapshot it already fetched — no extra `SELECT`.

My overall focus tho, is repair the columns that could be affected on the "song" which was the most heavy one in terms of data quantity. This are the cases i tested now, for check if my proposition its still working with this changes:

- parent='00000000'` on 10k rows → all repaired, logged `Repaired hierarchy info: 10000`
- `folder='00000000'` on 10k rows → same
- both `folder`/`parent` corrupted → same
- `folder=''`/`parent=''` → same (covered by the `length() > 0` check)
- `parent = folder` (the original PR #111 signature) → same

the table "folder" is not covered on this case. I think this is the client´s job, but i need to study it more, which is a work for another day and is not a game breaking issue today.

chown2:
Only having to update 'song' makes this easier. My point is that we don't need to have to update database. queries from outside ReloadSongsByQuery. This can be self-contained in a single function and be more effective to in addition to being vastly more readable. Just update the rows that match this SQL query `SELECT path FROM song WHERE folder = parent;`.

Skalim112:
> Only having to update 'song' makes this easier. My point is that we don't need to have to update database. queries from outside ReloadSongsByQuery. This can be self-contained in a single function and be more effective to in addition to being vastly more readable. Just update the rows that match this SQL query `SELECT path FROM song WHERE folder = parent;`.

I must admit i got too greedy with the "update parent on every X situation" but i never thought what could cause a parent column being corrupted... now it covers only the folder = parent, and its isolated

chown2 (approved):
As the commit mentions, I reverted the part of #113 related to song updating for 'parent'.
You can get a more readable diff with `git diff e1e00bc599fa58d00ada75c6148f77a3bed51c07~ HEAD`.
I've checked F8 on my game database, adding new songs on a fresh database, and a database where I intentionally did `update song set parent = folder`.
@Skalim112 please recheck that affected databases still do get fixed. Then LGTM.
Thanks!

chown2:
By using the 全曲リロード button in LR2.exe I can see that this reintroduced scanning for the second time, however, even though we still keep a copy of songs to update which #113 says was the fix. Secondary checks are not the solution anyway, we have to figure out why we get to the songs again in the first place. Logs only mention each bms file once, by the way.

Skalim112:
The second scan is being done by mtime modification, if your root folder's date doesnt match the db's date of that folder, the game does a rescan of every folder inside of the root folder searching if there is any change made.

This is totally outside of the scope i was fixing. The #113 fixes the infinite loop provoked by this particular case. But testing, yea, the second loop its still ocurring.


I am not at home, give me a moment once i get there so, i can do the testing

Skalim112:
[Log1.txt](https://github.com/user-attachments/files/28862971/Log1.txt) (First scan check)
[Log2.txt](https://github.com/user-attachments/files/28863836/Log2.txt) (Second scan check, cached DB)
for autoreload 1 and 2, with 10.000 cases where folder = parent, its working. Now, i forgot to cover the F8 case so, i´m going to work on a better approach for this case, and also see if i can mitigate the mtime SELECT check completely

EDIT: It seems like F8 its also covered. I tested with date = 1 and folder = parent. Its slow af, but it does the job, the main scope is solved.

So, on summary: my main proposal was cover every parent style corruption. We ended covering every single case of folder = parent since every other case of corruption will never be happening. and the previous functions done on #113 were not bringing proper value.

By the end, i´m going to repurpose those functions for keep fighting for a fast folder scan, since i ended being greedy by an edge point case. I think that could be a good idea for a new branch honestly, unless you tell me to do the changes here

chown2:
Git note: I've forced pushed into your branch before merging to remove unnecessary commits. Please don't merge your local branch into this one. 

> I think that could be a good idea for a new branch honestly, unless you tell me to do the changes here

New branch, please

chown2 (approved):
Thank you!

PR #126, opened 2026-06-13
feat(ir): add OpenLR2 User-Agent to POST requests 

ez2patternBMS:
Returns actual version number so that indicate to IR using host modification that this is openLR2.

### DIFF
--- a/OpenLR2_vs22.vcxproj
+++ b/OpenLR2_vs22.vcxproj
@@ -70,6 +70,7 @@
     <ClInclude Include="LR2\Scenes.h" />
     <ClInclude Include="LR2\strclass.h" />
     <ClInclude Include="LR2\structure.h" />
+    <ClInclude Include="LR2\LR2_version.h" />
   </ItemGroup>
   <ItemGroup>
     <ClCompile Include="lib\md5.cpp" />
@@ -104,6 +105,7 @@
     <ClCompile Include="LR2\LR2_songmanage.cpp" />
     <ClCompile Include="LR2\LR2_statlong.cpp" />
     <ClCompile Include="LR2\LR2_statplay.cpp" />
+    <ClCompile Include="LR2\LR2_version.cpp" />
     <ClCompile Include="LR2\Main.cpp" />
     <ClCompile Include="LR2\Scene02_Songselect.cpp" />
     <ClCompile Include="LR2\Scene03_Decide.cpp" />
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,7 @@ add_library(
     LR2/LR2_songmanage.cpp
     LR2/LR2_statlong.cpp
     LR2/LR2_statplay.cpp
+    LR2/LR2_version.cpp
     LR2/Scene02_Songselect.cpp
     LR2/Scene03_Decide.cpp
     LR2/Scene04_Play.cpp
--- a/LR2/LR2_ir.cpp
+++ b/LR2/LR2_ir.cpp
@@ -1,4 +1,6 @@
-﻿#include <fstream>
+﻿// LR2IR integration is deprecated. Please help us improve CustomIR instead.
+
+#include <fstream>
 #include <sstream>
 #include <string>
 #pragma comment(lib, "ws2_32.lib")
--- a/LR2/LR2_ir.cpp
+++ b/LR2/LR2_ir.cpp
@@ -1,15 +1,16 @@
 ﻿#include <fstream>
 #include <sstream>
 #include <string>
-#pragma comment(lib,"ws2_32.lib")
-#include "LR2_ir.h"
+#pragma comment(lib, "ws2_32.lib")
 #include "DxLib/DxLib.h"
-#include "tinyxml/tinyxml.h"
 #include "En_dbio.h"
 #include "En_fileutil.h"
 #include "En_timer.h"
 #include "En_xml.h"
+#include "LR2_ir.h"
+#include "LR2_version.h"
 #include "filesystem.h"
+#include "tinyxml/tinyxml.h"
 
 #ifdef _WIN32
 #include <shellapi.h>
@@ -546,10 +547,12 @@ int NETWORK::HTTPrequest() {
 
 		request.fillzero();
 		cstrSprintf(&request, "POST %s HTTP/1.0\r\n"
+							  "User-Agent: %s/%d\r\n"
 							  "Content-Type: application/x-www-form-urlencoded\r\n"
 							  "Content-Length:%d\r\n"
 							  "\r\n"
-							  "%s", this->target_URL.body, this->param.length(), this->param.body);
+							  "%s",
+							  this->target_URL.body, openlr2::clientName, openlr2::versionCode, this->param.length(), this->param.body);
 
 		if (send(s, request, request.length() + 1, 0) < 0) {
 			cstrSprintf(&this->request_debug, "データの送信に失敗しました : %d\n", WSAGetLastError());
--- a/LR2/LR2_ir.cpp
+++ b/LR2/LR2_ir.cpp
@@ -50,7 +50,7 @@ void MYRANKING::InitRanking() {
 	this->inputtype = 0;
 }
 
-int CMP_PlayerByExscore(const void *p1, const void *p2) {
+static int CMP_PlayerByExscore(const void *p1, const void *p2) {
 	
 	RANKINGPLAYER* s1 = (RANKINGPLAYER*)p1;
 	RANKINGPLAYER* s2 = (RANKINGPLAYER*)p2;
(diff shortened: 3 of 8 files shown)


MatVeiQaaa:
Why is this needed?

ez2patternBMS:
To inform to IR using host modification that this is OpenLR2 and not original LR2.
Of course, it is recommended to use the CustomIR API, but there are some users who use host modification method because of one recent IR.

MatVeiQaaa:
> To inform to IR using host modification that this is OpenLR2 and not original LR2.
> 
> Of course, it is recommended to use the CustomIR API, but there are some users who use host modification method because of one recent IR.

In this case, proposed version field doesn't indicate that it's OpenLR2 transparently.

MatVeiQaaa:
Also, since you modify version string #define anyway, remove repeating definitions, and also have full version string already complete, instead of assembling it in-place from two definitions. You can do that by using ver date define in the ver string define. Sorry, I can't make a proper review to show it on code.

ez2patternBMS:
> > To inform to IR using host modification that this is OpenLR2 and not original LR2.
> > Of course, it is recommended to use the CustomIR API, but there are some users who use host modification method because of one recent IR.
> 
> In this case, proposed version field doesn't indicate that it's OpenLR2 transparently.

From the server's perspective, to check if it's OpenLR2, they just need to verify if the client version is higher than 100130. Also, for score.cgi requests, they can simply check if the version param exists. Therefore, it's easy to determine if it's OpenLR2 using this method.

MatVeiQaaa:
If the goal is to say that this is openlr2, there must be a param to say precisely that. A version string doesn't necessarily indicate that this is openlr2, even though it can be inferred. No deduction should be necessary at all. Perhaps a "client" field would fit this task better.

MatVeiQaaa:
It would be best to have both "client" and "version" fields though, because version field can have other use cases.

Sonic-TV:
There was a "client" field back when LR2IR was active.
Guess that was used for players who's still using emLR or LR1 but it never got used.

ez2patternBMS:
I added the #define LR2CLIENT and added the client field on login and sendScore. It's weird that LR2CLIENT and LR2VERSION strings are in LR2_ir.h, so I think I'll have to move it somewhere else later.

ez2patternBMS:
> I added the #define LR2CLIENT and added the client field on login and sendScore. It's weird that LR2CLIENT and LR2VERSION strings are in LR2_ir.h, so I think I'll have to move it somewhere else later.

I moved them to Version.h

LR2/Version.h
```diff
@@ -0,0 +1,7 @@
+#define LR2CLIENT "OpenLR2"
+#define LR2BUILDDATE "260608"
+#define LR2VERSION LR2BUILDDATE
+#define LR2VERSIONSTRING LR2CLIENT " version " LR2VERSION
```
chown2: Why four constants instead of just two?
Hm, this was made in response to this comment @MatVeiQaaa  https://github.com/GOMazk/OpenLR2/pull/126#issuecomment-4698405912
But in my opinion, each place carries a different meaning, so a shared constant doesn't make sense in there

---

LR2CLIENT doesn't seem needed either. It's always going to be "OpenLR2".

---

I think we only need a version constant.
```suggestion
namespace openlr2 {
const int version;
} // namespace openlr2
```
This is implicitly 'extern', so put this into a cpp file, ideally a new one but Main.cpp would suffice for now

LR2/Version.h
```diff
@@ -0,0 +1,7 @@
+#define LR2CLIENT "OpenLR2"
+#define LR2BUILDDATE "260608"
```
chown2:
```suggestion
namespace openlr2 {
const int version;
} // namespace openlr2
```
Put the actual version into a 'cpp' file.

OpenLR2_vs22.vcxproj
```diff
@@ -1,365 +1,366 @@
-<?xml version="1.0" encoding="utf-8"?>
```
chown2, line 1: Line ends changed? Unless Visual Studio did that, please undo.

chown2 (changes requested):
1. > To inform to IR using host modification that this is OpenLR2 and not original LR2.

    LR2IR is down. Is this intended for a replacement IR?
1. Why would the IR care? Are you involved with it? My only idea is collecting stats about client being used. If I guessed right, are you going to display them openly?
1. Please don't break LR2IR API. If you absolutely need this, set 'User-Agent' in HTTP headers instead. Grep for 'Content-Type' in LR2_ir.cpp.

ez2patternBMS:
1. Yes
2. I am not officially participating in any other IR projects. However, since there is at least a version field in the login function, I don't think it is appropriate to hardcode it as 100130. The first commit was for that.
3.  From the server's perspective, it can simply choose to ignore that parameter. That is why I implemented it this way.
If you think this will affect to much, I will change to use `User-Agent`

​I have a question regarding point 3. If a modification is made that affects the chart rankings (e.g., adding #RANK 4 or adding support for #DEFEXRANK), should the LR2IR API still transmit the same value? (e.g., hardcoding the API to send rank=2 for RANK 4).
​P.S. I accidentally clicked "Re-request review." The modifications are not finished yet.
​

LR2/Version.h
ez2patternBMS: I included `LR2CLIENT` considering the possibility that openlr2 might also be forked, just like beatoraja and lr2oraja.

If you think this isn't necessary, I will remove `LR2CLIENT`.

chown2:
> I don't think it is appropriate to hardcode it as 100130.

I understand that, but there is no *the* LR2IR anymore.
Unless you redirect www.dream-pro.info, the domain where LR2IR used to be at, you won't even be able to connect anywhere using this code.

> From the server's perspective, it can simply choose to ignore that parameter.

That's true, but changing 'version' is technically a breaking change.
Meanwhile User-Agent I've proposed using above would make them able to detect LR2 without breaking comptibility with LR2 while also using a universal web standard.

> If a modification is made that affects the chart rankings

Such changes will eventually come, but it doesn't mean anything for LR2IR API code.
Bringing up the first point, the game reaches out to 'www.dream-pro.info', but that service isn't up anymore.
If a user makes OpenLR2 transmit scores somewhere, it's completely on them.

---

You appear to have wrote this while I was editing some of the messages above. :)
I've moved my comments around, but the points remained the same.

LR2/Version.h
chown2: I've given it some more thoughts, and you should keep it. I hope we won't have any hard forks, but it still may make sense to be able to 'brand' the build in a specific way while some things are tested, like the ongoing STELLAVERSE IR builds are.

ez2patternBMS:
Okay, then I will modify it to use http header instead of using get parameter around 4PST.

ez2patternBMS:
Just a personal thought, but the main argument that "modifying the code is meaningless because LR2IR is closed" seems a bit contradictory to the concern that "modifying the GET parameters could have a major impact on an (unknown) server despite LR2IR being closed."

Furthermore, does this mean there are long-term plans to completely remove the LR2IR API or replace it with code for a custom IR? It seems a bit strange to continue improving other features without actually removing the LR2IR API. Even if it is being kept for preservation purposes, doesn't the LR2beta3 branch already serve that role?

chown2:
I personally use LR2IR API through this tool, which by whatever reason expects the original 'version=100130'. https://sr.ht/~showy_fence/lr2ir-local-leaderboard

All of the LR2IR code will eventually be removed.
But definitely not this very instant as it can serve as a reference for CustomIRs.

> It seems a bit strange to continue improving other features without actually removing the LR2IR API

The only actual change to LR2IR code in OpenLR2 so far is me practically just documenting an LR2 bug in git history 141a8eb1bdf4cd6a36c81845fc907112a9c2aa95

ez2patternBMS:
Okay, if you have a plan to exclude lr2ir api in the long run, that would be great.

> > If a modification is made that affects the chart rankings
> 
> Such changes will eventually come, but it doesn't mean anything for LR2IR API code.

I was a little confused about the intention here, so I asked you. As suggested, I will revise the LR2IR API code to a form that does not touch as much as possible.

OpenLR2_vs22.vcxproj
ez2patternBMS: IDK, I think VS did this.
Although I set git config core.autocrlf to false, it cause same result.

chown2:
Thank you!
I've moved the version definition into a new file since Main.cpp isn't included in the main target in the CMake build,
and dropped 'Windows' from the User-Agent as it's not necessarily going to be on Windows and it's unnecessary info
anyway.

chown2 (approved):

ez2patternBMS:
> Thank you! I've moved the version definition into a new file since Main.cpp isn't included in the main target in the CMake build, and dropped 'Windows' from the User-Agent as it's not necessarily going to be on Windows and it's unnecessary info anyway.

We have both x86 and x64 client, so can't it be useful for bug report?

By the way, I agree that it doesn't need for user-agent

PR #128, opened 2026-06-14
fix: scores accumulate during a normal restart + Incorrect starting s…

luxurya111:
This PR fixes the issue where a course starts from the second stage upon restart, as well as the issue of scores accumulating.

The QuickRestart logic has been modified because Scene05 executes it even during a normal restart. Please let me know if this part causes any issues.

Issue: https://github.com/GOMazk/OpenLR2/issues/103

### DIFF
--- a/LR2/Main.cpp
+++ b/LR2/Main.cpp
@@ -1606,11 +1606,13 @@ int main(int argc, char** argv) {
 					if (gs.gameplay.courseType == 0 || gs.gameplay.courseType == 2) {
 						gs.gameplay.player[0].gaugeType = gs.gameplay.player[0].lastCourseGaugeType;
 						gs.gameplay.player[1].gaugeType = gs.gameplay.player[1].lastCourseGaugeType;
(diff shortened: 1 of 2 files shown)


MatVeiQaaa:
This makes sense, although I'm not sure about removal of courseStage != 0 condition. It was there because flagRetry indicates resources reload. With it set to 0, the chart is gonna be reloaded. Now, if you restart the course from later stages, and the flag is set to 1, isn't it gonna restart with wrong song?

MatVeiQaaa:
Right, but it's removal of condition for the flag to be set to 0, so now it will be set to 0 on stage 0 as well. Did it cause issues?

luxurya111:
I tried restarting on the first song, and I also tried restarting on the second song. (I tested it a total of 4 times, keeping the clear and failed states for each.)

Since there is logic that forces it to stage 0, there were no issues when I tested it.

MatVeiQaaa:
So, what was the motivation for removing that condition? It was put there to skip unnecessary processing for restarting stage 0.

luxurya111:
Ah, I'm sorry. I misunderstood. It was definitely an unnecessary modification.
I ended up modifying it needlessly while checking various conditions.

PR #133, opened 2026-06-15
perf(songmanage): avoid converting encodings and use ASCII when possible

Skalim112:
## Summary

Speeds up BMS metadata scanning by avoiding a CP932→UTF-8 conversion on **every** line of every chart. `ParseBMSMETA` previously converted each line to UTF-8 and then matched it with `CSTR` string operations. Most lines in a .bms file are note-channel data and numeric headers, which are pure ASCII and don't need any conversion.

In short, it was looping inside **every** line of the chart, searching for the hardcoded metadata (maybe an oportunity to refactor too?)

Now only the lines that actually carry text metadata are converted; everything else is parsed directly from the raw ASCII bytes. Output is unchanged (same metadata, note counts, BPM range, and MD5 hashes); the scan just does much less per-line work.

## What changed

In `LR2/LR2_songmanage.cpp`, the `ParseBMSMETA` line loop is rewritten:

- **Text metadata lines** (**_`#TITLE`, `#SUBTITLE`, `#GENRE`, `#CATEGORY`,`#ARTIST`, `#SUBARTIST`, `#INFORMATION_A/B`, `#TAG`, `#COMMAND`, `#STAGEFILE`, `#BANNER`, `#BACKBMP`_**) are still converted with `ansi2utf(..., 932)` before parsing, so non-ASCII titles/artists are handled exactly as before.

- **Numeric / control / chart-object lines** (**_`#IF`/`#ENDIF`, `#PLAYLEVEL`, `#EXLEVEL`, `#MAXTRACKS`, `#DIFFICULTY`, `#RANK` `#CUSTOMFOLDER`, `#FP/DSC`, `#LNOBJ`, `#BMP`, `#RAND`, `#BPM`_**, and the `#xxxYY:...` channel data) are parsed straight from the raw line.

- Small ASCII helpers added for this: `StartsWithAscii` / `StartsWithAsciiNoCase`, `IsAsciiDigit`, `IsTextMetaLine`, `GetRawBodyInt`, `UpdateBpmRange`, `ParseTextMetaLine`, At this way, the functions could be reutilized for future implementation (like new METADATA that could modify the behaviour of a chart).

## Why it's safe

Header keywords, digits, hex and base36 used by the channel/numeric data are all ASCII, so reading them from the original bytes gives the same values as reading them after a UTF-8 round-trip. Only human-readable text fields can contain non-ASCII (Shift-JIS) characters, and those keep the conversion.

## Scope / not included

- Only touches `ParseBMSMETA` in `LR2/LR2_songmanage.cpp`. No scanner, DB,
  threading, or reload changes.
- It covers Autoreload=2, every other situation like F8 reload and Autoreload=1 works on a different way so, better to start with this first.

## Testing

- Clean scan from an empty `song.db`: completed without crashes; `115099` songs / `12586` folders: 1 Hour process (its still doing double SELECT check, but it was tested vs a 6 hours process from the previous iteration).
- `parent`/`folder` never null/empty.
- Populated DB, full update path: completed with consistent counts and

## Files

- `LR2/LR2_songmanage.cpp` (only file touched).

For other optimizations, i am working on another branches, i hope this ends being useful for the purpose of the project.

### DIFF
--- a/LR2/LR2_songmanage.cpp
+++ b/LR2/LR2_songmanage.cpp
@@ -3,8 +3,11 @@
 #include "LR2_statlong.h"
 #include "filesystem.h"
 #include "filesystem.h"
+#include <cstring>
 #include <iterator>
 #include <ranges>
+#include <string_view>
+#include <charconv>
 #include <vector>
 
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


chown2 (commented):
Did you measure the impact? In my experience, converting encodings is pretty fast. Only doing it for specific commands makes sense, but introduces the possibility of not converting the encoding somewhere by accident.

LR2/LR2_songmanage.cpp
```diff
@@ -12,6 +13,94 @@
...
+	if ('a' <= ch && ch <= 'z') return ch - ('a' - 'A');
+	return ch;
+}
+
+bool StartsWithAsciiNoCase(const char* str, const char* head) {
+	if (!str || !head) return false;
+	while (*head) {
+		if (ToUpperAscii(*str) != ToUpperAscii(*head)) return false;
+		++str;
+		++head;
```
chown2: buffer overrun :D

LR2/LR2_songmanage.cpp
```diff
@@ -12,6 +13,94 @@
...
+	}
+	return true;
+}
+
+bool StartsWithAscii(const char* str, const char* head) {
+	if (!str || !head) return false;
+	return strncmp(str, head, strlen(head)) == 0;
+}
```
chown2: Replace this function and it's usage with std::string_view{line}.starts_with

LR2/LR2_songmanage.cpp
```diff
@@ -12,6 +13,94 @@
...
+		StartsWithAsciiNoCase(line, "#STAGEFILE") ||
+		StartsWithAsciiNoCase(line, "#BANNER") ||
+		StartsWithAsciiNoCase(line, "#BACKBMP");
+}
+
+bool GetRawBodyInt(const char* line, const char* head, int* out) {
+	const size_t headLen = strlen(head);
+	if (!StartsWithAsciiNoCase(line, head)) return false;
+	if (strlen(line) <= headLen) return false;
+	*out = atol(line + headLen + 1);
```
chown2: std::from_chars is faster than atol, altough there's probably more overhead from other things in here

MatVeiQaaa:
Exactly for the reason which you explained conversion can be skipped (ASCII characters got the same code on both codepages) is why this is likely saving close to no time. No conversion is being done there.

LR2/LR2_songmanage.cpp
Skalim112: This has not happened on my testing, but i did a newer approach on StartsWithAsciiNoCase

LR2/LR2_songmanage.cpp
Skalim112: Done

LR2/LR2_songmanage.cpp
Skalim112: Tested, its working, and it really seems faster, i will post the log below

Skalim112:
> Exactly for the reason which you explained conversion can be skipped (ASCII characters got the same code on both codepages) is why this is likely saving close to no time. No conversion is being done there.

Compared to the previous .exe, considering i am playing on HDD, the reduction times are clearly noticeable on an environment with 116k charts recognized, improving (at least here) 5 hours of process.

With chown2´s suggerence, i did a testing with a cold database, and it gave me around 68 mins of processing.

[Log.txt](https://github.com/user-attachments/files/28961994/Log.txt) here is the log of the last testing, but in resume:

Folder update: ~39 min 21 s.
File update: ~29 min 35 s.

on a second test with a hot cached data, this could be even faster.

About the conversion, yea, its a fair point. The speedup is not really from ASCII byte translation itself. What this change avoids is routing every line through the UTF-8/CSTR parsing path while looking for metadata. Most chart lines are note-channel data or numeric headers, which can be handled directly from the raw ASCII line, so the conversion does extra work on lines that are not the main focus of the function.

LR2/LR2_songmanage.cpp
```diff
@@ -3,15 +3,107 @@
...
+	if (!out) return false;
+	if (!StartsWithAsciiNoCase(line, head)) return false;
+	if (line.size() <= head.size() + 1) return false;
+
+	int value = 0;
+	const char* first = line.data() + head.size() + 1;
+	const char* last = line.data() + line.size();
+
+	const auto result = std::from_chars(first, last, value);
+	if (result.ec != std::errc()) return false;
```
chown2, line 44: Style nit
```suggestion
	int value = 0;
	const auto result = std::from_chars(line.data() + head.size() + 1, line.data() + line.size(), value);
	if (result.ec != std::errc()) return false;
```

LR2/LR2_songmanage.cpp
```diff
@@ -3,15 +3,107 @@
...
+		StartsWithAsciiNoCase(line, "#INFORMATION_B") ||
+		StartsWithAsciiNoCase(line, "#TAG") ||
+		StartsWithAsciiNoCase(line, "#COMMAND") ||
+		StartsWithAsciiNoCase(line, "#STAGEFILE") ||
+		StartsWithAsciiNoCase(line, "#BANNER") ||
+		StartsWithAsciiNoCase(line, "#BACKBMP");
+}
+
+bool GetRawBodyInt(std::string_view line, std::string_view head, int* out) {
+	if (!out) return false;
```
```suggestion
	assert(out != nullptr && "GetRawBodyInt");
```
Requires an `#include <cassert>`

chown2: No

```suggestion
bool GetRawBodyInt(std::string_view line, std::string_view head, int& out) {
```
and fix up the call sites

chown2:
It would be best if you tested this with a profiler instead, like the one built into Visual Studio. You don't need to reload your whole library each time, just pausing execution after 30 seconds is fine.
I have a suspicion that a solid chunk of the time is allocations overhead. And if so, it's ansi2utf that should be
optimized instead.
But for now this is perfectly good.
LGTM after fixing nits

Skalim112:
> It would be best if you tested this with a profiler instead, like the one built into Visual Studio. You don't need to reload your whole library each time, just pausing execution after 30 seconds is fine. I have a suspicion that a solid chunk of the time is allocations overhead. And if so, it's ansi2utf that should be optimized instead. But for now this is perfectly good. LGTM after fixing nits

Well, the perf/scan-doubleparse was made with a profiler in mind: https://drive.google.com/file/d/13PJJyxa2TEEv6Z7fAL7szkBaRdEtosdn/view?usp=sharing

(the file is too big for upload it here).

LR2/LR2_songmanage.cpp
Skalim112: Done, applied on calls too, check last commit

chown2:
Refactored a little: rename things to resemble common idioms more, prefer references to pointers in more places, move IsTextCommand and ParseTextCommand (renamed names) near each other so that we have less chance to forget to update one but not the other. Tested, found no regressions.

chown2 (approved):
🚀

chown2:
> Well, the perf/scan-doubleparse was made with a profiler in mind:

Sadly, this is not viewable without having the game binary and it's debug symbols, so even you yourself probably can't look at it anymore. :(

PR #134, opened 2026-06-15
feat(ir): apply exlevel.xml locally even when LR2IR is unreachable 

### LINKED ISSUE #118: Insane table doesn´t update now that LR2IR is dead
If a new player initiates with a new song.db, go to JUKEBOX 2 then click on the BMS FOLDER checkbox, even with an imported exlevel.xml, its going to never update the table because:

1. it requires a successful connection with the IR
2. the IR is dead

## How to reproduce it

- [exlevel.xml](https://github.com/user-attachments/files/28822085/exlevel.xml) import this to LR2files/Database, this is what is supposed to load once the IR is connected
- Load the game with a new song.db.
- Make sure you have some chart of the insane table on your song folders.
- The game will load this fragment of code, inside is the validation that takes the exlevel.xml previouly loaded, ONLY if IR is connected

Conditional on main.cpp

Image

Function

Image

I wonder if this will be repurposed on a future, but a quick fix could be separating the reading of exlevel.xml outside of the IR conditional

Skalim112:
Fixes https://github.com/GOMazk/OpenLR2/issues/118 "Insane table doesn´t update now that LR2IR is dead"

---

This is going to be a short one, but in short, it separated exlevel.xml reading from internet connection.

exlevel.xml could also be used on a future for owners of customIRs for upload their own list of tables so, for someone who is making a client its also useful, specially with parallel processing.

On an ideal point, customIRs could work together for bring an standard to the exlevel.xml so, everyone could upload the most played tables AND personal ones automatically without the need of bemusicseeker.

### DIFF
--- a/LR2/Main.cpp
+++ b/LR2/Main.cpp
@@ -418,8 +418,11 @@ int main(int argc, char** argv) {
 			: fs::make_preferred("LR2files/Database/song.db").data(), &sql3);
 	LoadLR2CustomFolder(sql3, &gs.config.jukebox, pathScoreDB, gs.is_starter, gs.cmd_directplay);
 	if (gs.cmd_directplay == false) {
-		if (gs.config.network.lr2ir == 1 && (((unsigned char)gs.config.jukebox.customfolder & 0x80) != 0)) {
-			gs.net.GetInsaneList();
+		if (((unsigned char)gs.config.jukebox.customfolder & 0x80) != 0) {
+			if (gs.config.network.lr2ir == 1) {
+				gs.net.GetInsaneList();
+			}
+			gs.net.ApplyInsaneList();
 		}
 		if (gs.is_starter == false) {
 			printfDx(LR2VERSIONSTRING"\n");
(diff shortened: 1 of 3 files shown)


LR2/Main.cpp
```diff
@@ -418,8 +418,13 @@ int main(int argc, char** argv) {
...
 	if (gs.cmd_directplay == false) {
-		if (gs.config.network.lr2ir == 1 && (((unsigned char)gs.config.jukebox.customfolder & 0x80) != 0)) {
-			gs.net.GetInsaneList();
+		if (((unsigned char)gs.config.jukebox.customfolder & 0x80) != 0) {
+			if (gs.config.network.lr2ir == 1) {
+				gs.net.GetInsaneList();
+			}
+			else {
+				gs.net.ApplyExLevelFromFile();
+			}
```
chown2:
```suggestion
			if (gs.config.network.lr2ir == 1) {
				gs.net.GetInsaneList();
			}
			gs.net.ApplyExLevelFromFile();
```
Better split it in a way where things compose together

chown2:
ApplyInsaneList (renamed from ApplyExLevelFromFile for consistency with GetInsaneList) was being ran twice after that change, so I fixed that and made the loadingGrHandle fix more generic.

chown2 (approved):
Thanks!

PR #135, opened 2026-06-15
perf: QOL on folder reading at autoreload = 2. Caching path for already known files inside folders.

Skalim112:
## Summary

Speeds up the **second SELECT** (`autoreload == 2` and the startup scan) and adds an on-screen progress indicator for it, without changing detection behavior or correctness. All changes are contained in `LR2/LR2_songmanage.cpp`.

The full reload used to parse the same chart twice in one cycle: the folder does `ParseBMSMETA`' every existing chart in a changed folder (only for the `INSERT` to fail on the duplicate `path` key), and charts that were already covered on song.db gets analyzed again. This PR removes that redundant work and shows the user what the reload is doing.

## What changed

- **One-shot reload snapshot.** Before the folder/song passes, the existing `song.path -> song.hash` rows are loaded once into memory with a single `SELECT`, instead of issuing per-file SQL during the recursive scan.

- **Folder pass skips redundant parsing.** When a re-scanned folder contains a chart already present in the snapshot, `ParseBMSMETA` is skipped. Brand-new charts are still parsed and inserted so, correctness is preserved.

- **Song pass hash-refresh fallback.** When a chart is reported as changed by mtime (common with imported DBs whose stored dates are stale), the file is hashed; if the hash still matches the cached DB hash, only `song.date` is updated and the full reparse is skipped. If the hash differs, the normal delete/parse/insert path runs, so real edits are still picked up.

- **Loading-screen progress for the full reload.** Added new QOL progress visualizer for situations like 'song.db already populated and recognized' and 'First SELECT on migrated song.db'.

- **Diagnostics.** Per-phase counters and a summary log line for the reload.

## Scope / not included

- This is only for autoreload=2

## Testing

- Fresh/empty `song.db` full scan: completes; `song`/`folder` counts correct; `SELECT count(*) FROM song WHERE folder = parent` → 0: Mitigates double SELECT depending of song.db´s integrity so, time may vary.

## Files

- `LR2/LR2_songmanage.cpp` (only file touched).

## IMPORTANT NOTE

With the tools that we already have, there is not a possibility to completely mitigate the second SELECT. This behaviour was made for check if a chart was modified, capturing their hash and inserting it to the 'folder' table.

If this function is totally mitigated, modified charts could get lost by the game so, this covers that kind of behaviour... which is a very edge-case since its rare to see someone modify a chart that was loaded before (unless you are a charter, maybe...?)

**The QOL includes a change on behaviour of the game (visually), i encommend test it if you like it, otherwise i can revert it and keep the scope at minimum**


chown2 (commented):
Please read this https://github.com/GOMazk/OpenLR2/pull/116#issuecomment-4684769198
Sorry, but I don't see anything acceptable here.
I think it would be fine to have secondary checks if they would have minimal code for them and not impact readability (like using one-line if-guards instead of increasing nesting).
To actually speed up song scanning we must introduce multi-threading.

Skalim112:
> Please read this [#116 (comment)](https://github.com/GOMazk/OpenLR2/pull/116#issuecomment-4684769198) Sorry, but I don't see anything acceptable here. I think it would be fine to have secondary checks if they would have minimal code for them and not impact readability (like using one-line if-guards instead of increasing nesting). To actually speed up song scanning we must introduce multi-threading.

That´s something i was thinking on a moment, but i don´t wanted to touch too much of the code since i don´t want to be too invasive with the flux of the game and your plans. Thats why i am trying to give optimizations little by little, considering that, maybe in the future its going to be dead code.

I will try to give a more adequate solution from here to tomorrow.

Skalim112:
Ok so, i managed to cut some nesting, and the log were isolated on the upper part of the code.

## Why the logs?

With this solution, the known charts doesn´t execute parseBMS at all because its considered a "registered charts without modifications", and parseBMS was the responsible to show the logs of the charts being registered to song.db (flag=1, ProcessMessage()).

Applying this change, and in conjunction with parseBMS, the loading times goes from 1 hour (parseBMS implementation) to 37 mins cold cache: [Log.txt](https://github.com/user-attachments/files/29032496/Log.txt)

Checked song.db´s integrity with the previous solution: The data doesn´t have discrepancies.

Tested with profiler: Now the cpu charge goes to makeFileHash


## A little look of the logs



This is for people who open the game with a cold cache on autoreload=2 and has a reloaded song.db from before, because, since there are no changes on the database, the game gets freeze without response, which can be confusing for the player.
imagen

i hope this suits better the project

LR2/LR2_songmanage.cpp
```diff
@@ -18,6 +21,127 @@ int EnabledInsane;
 
 namespace {
 
+// --- Reload scan cache (single init per full autoreload) -------------------
+// Built once before a full reload; lets the folder/song passes skip charts
+// already present in song.db. The skip is hash-verified, so an in-place edit
+// (same path, changed content) still falls through to a full reparse.
+bool g_fullSongPass = false;
+std::unordered_set<std::string> g_knownSongPaths;
```
chown2: Memory optimization: g_knownSongPaths is redundant since g_songHashes has the same data
```suggestion
```

LR2/LR2_songmanage.cpp
```diff
@@ -1282,11 +1412,24 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb) {
 	}
 	sqlite3_finalize(pStmt);
 
+	const bool showFolderPassProgress = g_fullSongPass && query.isSame("SELECT path,date FROM folder");
+	const bool showSongPassProgress = g_fullSongPass && query.isSame("SELECT path,date FROM song");
```
chown2: Very ugly but ok

LR2/LR2_songmanage.cpp
```diff
@@ -18,6 +21,127 @@ int EnabledInsane;
...
+void EndSongReloadSnapshot() {
+	g_fullSongPass = false;
+	g_knownSongPaths.clear();
+	g_songHashes.clear();
+	g_processedSongPaths.clear();
+}
+
+// --- Scan progress display (loading screen feedback during autoreload) ------
+
+void ShowReloadFolderPassProgress(const size_t processed, const size_t total, CSTR path) {
```
chown2, line 60: Can we avoid three separate 'loading...' functions?

LR2/LR2_songmanage.cpp
Skalim112: Applied on the last commit

LR2/LR2_songmanage.cpp
Skalim112: Presented a better approach on the last commit

LR2/LR2_songmanage.cpp
Skalim112: Done

LR2/LR2_songmanage.cpp
```diff
@@ -18,6 +21,128 @@ int EnabledInsane;
...
+	g_fullSongPass = false;
+	g_songHashes.clear();
+	g_processedSongPaths.clear();
+}
+
+// --- Scan progress display (loading screen feedback during autoreload) ------
+
+void ShowReloadFolderPassProgress(const size_t processed, const size_t total, CSTR path) {
+	if (!g_fullSongPass || total == 0) return;
+	if (processed != total && processed % 64 != 0) {
```
chown2, line 62: No need to act in this PR: folders might not even need this, some folders take a while so a feeling of a 'freeze up' happens.

chown2: Also, this could be timer based, where you'd update every 100ms or so. Won't even be any more complex.

LR2/LR2_songmanage.cpp
```diff
@@ -18,6 +21,128 @@ int EnabledInsane;
...
+	ClsDrawScreen(); 
+	clsDx(); 
+	ProcessMessage();
+}
+
+// Song-pass feedback. Called with an empty path from the row loop (shows the
+// overall progress bar, only until the first file gets hash-checked) and with a
+// concrete path from RefreshSongDateIfHashMatches (shows the folder being
+// hashed, once per folder).
+void ShowReloadSongPassProgress(const size_t processed, const size_t total, CSTR path) {
```
chown2, line 81: No need to act in this PR: would be nice if this was the same as ShowReloadFolderPassProgress too. The third function didn't really go away either, it's just a branch in this one now. I don't see any reason to keep any of them split, it should be a single function like `ShowProress(size_t processed, size_t total, std::string_view message)`.

LR2/LR2_songmanage.cpp
```diff
@@ -1282,11 +1410,24 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb) {
 	}
 	sqlite3_finalize(pStmt);
 
+	const bool showFolderPassProgress = (progress == ReloadProgress::FolderPass);
+	const bool showSongPassProgress = (progress == ReloadProgress::SongPass);
```
chown2, line 1414: Nit: useless variables

LR2/LR2_songmanage.cpp
```diff
@@ -1282,11 +1410,24 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb) {
...
+	if (showFolderPassProgress) ShowReloadFolderPassProgress(0, pathList.size(), CSTR(""));
+
+	size_t rowIndex = 0;
 	for (const auto& [str, time] : std::views::zip(pathList, timeList)) {
+		rowIndex++;
+		if (showFolderPassProgress) ShowReloadFolderPassProgress(rowIndex, pathList.size(), str);
+		else if (showSongPassProgress) ShowReloadSongPassProgress(rowIndex, pathList.size(), CSTR(""));
```
chown2, line 1421: Nit: switch case would work better here

chown2 (approved):
Didn't look deep into it to be honest, but this seems to work and doesn't make the LR2 code I was planning to rewrite much harder to understand. LGTM.
GOMazk squash merge pls

LR2/LR2_songmanage.cpp
Skalim112: This could be good for SSDs actually.

LR2/LR2_songmanage.cpp
Skalim112: I can test it on a moment tho, if its not already merged when i am available

PR #141, opened 2026-06-19
Load/Update external IR record to player.db locally

Unengine:
## Changes
- LR2_songmanage separately loads IR records by using an `imported_score` table.
- `imported_score` and `score` tables share the same schema.
- `std::optional<STATUS> myIRbest` field added to `SONGDATA` structure.
- `myIRbest` is used exclusively for skin features; it does not impact the original score. (not sent to IR)
- `LR2_skinobject` displays the better stats of the two (mybest and myIRbest).

You can test the changes with this importer.
https://github.com/Unengine/BokutachiToOpenLR2ScoreImporter

[Here is my plan documentation](https://app.notion.com/p/IR-Record-to-LR2-player-db-38452d65f39a801fbf78e059a816f9df?source=copy_link)

### DIFF
--- a/LR2/LR2_ghost.cpp
+++ b/LR2/LR2_ghost.cpp
@@ -764,6 +764,30 @@ CSTR ReadGhost(sqlite3 *sql, CSTR songMD5) {
 	return ghostdata;
 }
 
+CSTR ReadIRGhost(sqlite3* sql, CSTR songMD5) {
+	ErrorLogAdd("データベースからゴーストを読み込みます\n");
+
+	sqlite3_stmt* pStmt;
+	if (sqlite3_prepare_v3(sql, "SELECT ghost FROM imported_score WHERE hash = ?", -1, 0, &pStmt, nullptr) != SQLITE_OK) {
+		ErrorLogAdd("sqlite3_prepare_v3 error\n");
+	}
+	if (sqlite3_bind_text(pStmt, 1, songMD5.body, songMD5.length(), nullptr) != SQLITE_OK) {
+		ErrorLogAdd("sqlite3_bind_text error\n");
+	}
+
+	CSTR ghostdata;
+	if (sqlite3_step(pStmt) == SQLITE_ROW) {
+		ghostdata.resize(sqlite3_column_bytes(pStmt, 0));
+		// Cast safety: it's safe to cast from unsigned char* to char*
+		ghostdata = reinterpret_cast<const char*>(sqlite3_column_text(pStmt, 0));
+	}
+	if (sqlite3_finalize(pStmt) != SQLITE_OK) {
+		ErrorLogAdd("sqlite3_finalize error\n");
+	}
+
+	return ghostdata;
+}
+
 int ReadGhostToScore(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score) {
 	CSTR ghostdata = ReadGhost(sql, songMD5);
 
@@ -772,6 +796,18 @@ int ReadGhostToScore(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score) {
 		return 1;
 	}
 	
+	score->DecodeGhostData(ghostdata);
+	return 1;
+}
+
+int ReadIRGhostToScore(sqlite3* sql, CSTR songMD5, PLAYSCORE* score) {
+	CSTR ghostdata = ReadIRGhost(sql, songMD5);
+
+	if (ghostdata.length() == 0) {
+		score->InitJudgeQueue();
+		return 1;
+	}
+
 	score->DecodeGhostData(ghostdata);
 	return 1;
 }
--- a/LR2/LR2_ghost.cpp
+++ b/LR2/LR2_ghost.cpp
@@ -740,7 +740,7 @@ int WriteGhostInDatabase(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score) {
 	return 0;
 }
 
-CSTR ReadGhost(sqlite3 *sql, CSTR songMD5, bool useImportedGhost = false) {
+CSTR ReadGhost(sqlite3 *sql, CSTR songMD5, bool useImportedGhost) {
 	ErrorLogAdd("データベースからゴーストを読み込みます\n");
 
 	const std::string query = std::format("SELECT ghost FROM {} WHERE hash = ?", useImportedGhost ? "imported_score" : "score");
@@ -766,7 +766,7 @@ CSTR ReadGhost(sqlite3 *sql, CSTR songMD5, bool useImportedGhost = false) {
 	return ghostdata;
 }
 
-int ReadGhostToScore(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score, bool useImportedGhost = false) {
+int ReadGhostToScore(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score, bool useImportedGhost) {
 	CSTR ghostdata = ReadGhost(sql, songMD5, useImportedGhost);
 
 	if (ghostdata.length() == 0) {
--- a/LR2/LR2_ghost.cpp
+++ b/LR2/LR2_ghost.cpp
@@ -740,11 +740,13 @@ int WriteGhostInDatabase(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score) {
 	return 0;
 }
 
-CSTR ReadGhost(sqlite3 *sql, CSTR songMD5) {
+CSTR ReadGhost(sqlite3 *sql, CSTR songMD5, bool useImportedGhost = false) {
 	ErrorLogAdd("データベースからゴーストを読み込みます\n");
 
+	const std::string query = std::format("SELECT ghost FROM {} WHERE hash = ?", useImportedGhost ? "imported_score" : "score");
+	
 	sqlite3_stmt *pStmt;
-	if (sqlite3_prepare_v3(sql, "SELECT ghost FROM score WHERE hash = ?", -1, 0, &pStmt, nullptr) != SQLITE_OK) {
+	if (sqlite3_prepare_v3(sql, query.c_str(), -1, 0, &pStmt, nullptr) != SQLITE_OK) {
 		ErrorLogAdd("sqlite3_prepare_v3 error\n");
 	}
 	if (sqlite3_bind_text(pStmt, 1, songMD5.body, songMD5.length(), nullptr) != SQLITE_OK) {
@@ -764,50 +766,14 @@ CSTR ReadGhost(sqlite3 *sql, CSTR songMD5) {
 	return ghostdata;
 }
 
-CSTR ReadIRGhost(sqlite3* sql, CSTR songMD5) {
-	ErrorLogAdd("データベースからゴーストを読み込みます\n");
-
-	sqlite3_stmt* pStmt;
-	if (sqlite3_prepare_v3(sql, "SELECT ghost FROM imported_score WHERE hash = ?", -1, 0, &pStmt, nullptr) != SQLITE_OK) {
-		ErrorLogAdd("sqlite3_prepare_v3 error\n");
-	}
-	if (sqlite3_bind_text(pStmt, 1, songMD5.body, songMD5.length(), nullptr) != SQLITE_OK) {
-		ErrorLogAdd("sqlite3_bind_text error\n");
-	}
-
-	CSTR ghostdata;
-	if (sqlite3_step(pStmt) == SQLITE_ROW) {
-		ghostdata.resize(sqlite3_column_bytes(pStmt, 0));
-		// Cast safety: it's safe to cast from unsigned char* to char*
-		ghostdata = reinterpret_cast<const char*>(sqlite3_column_text(pStmt, 0));
-	}
-	if (sqlite3_finalize(pStmt) != SQLITE_OK) {
-		ErrorLogAdd("sqlite3_finalize error\n");
-	}
-
-	return ghostdata;
-}
-
-int ReadGhostToScore(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score) {
-	CSTR ghostdata = ReadGhost(sql, songMD5);
+int ReadGhostToScore(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score, bool useImportedGhost = false) {
+	CSTR ghostdata = ReadGhost(sql, songMD5, useImportedGhost);
 
 	if (ghostdata.length() == 0) {
 		score->InitJudgeQueue();
 		return 1;
 	}
 	
-	score->DecodeGhostData(ghostdata);
-	return 1;
-}
-
-int ReadIRGhostToScore(sqlite3* sql, CSTR songMD5, PLAYSCORE* score) {
-	CSTR ghostdata = ReadIRGhost(sql, songMD5);
-
-	if (ghostdata.length() == 0) {
-		score->InitJudgeQueue();
-		return 1;
-	}
-
 	score->DecodeGhostData(ghostdata);
 	return 1;
 }
--- a/LR2/LR2_skinobject.cpp
+++ b/LR2/LR2_skinobject.cpp
@@ -7,7 +7,6 @@
 #include "LR2_statplay.h"
 #include "Scenes.h"
 #include "filesystem.h"
-#include "filesystem.h"
 #include <DxLib/DxLib.h>
 
 #ifndef _WIN32
@@ -20,7 +19,7 @@ bool GetOptionFlag_dst(game *gs, int option) {
 	bool ret = (option >= 0);
 	if (!ret) option = -option;
 	const SONGDATA& songData = gs->sSelect.bmsList[gs->sSelect.cur_song];
-	const std::optional<STATUS>& myIRbest = songData.myIRbest;
+	const STATUS& mybest = songData.myIRbest.has_value() ? *songData.myIRbest : songData.mybest;
 
 	switch (option) {
 		case 0:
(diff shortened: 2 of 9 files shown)


LR2/LR2_songmanage.cpp
```diff
@@ -992,6 +992,7 @@ int GetSongData(CSTR songMD5, SONGDATA *song, sqlite3 *sql, SONGSELECT *ss) {
 	char buf[1024];
 	sqlite3_stmt *stmt;
 	int ret = 0;
+	CSTR IR_DERIVED_RECORD_HASH = "IR_DERIVED_RECORD";
```
chown2: 1. avoid CSTR in new code, it's horrible legacy
  ```suggestion
	constexpr auto&& IR_DERIVED_RECORD_HASH = "IR_DERIVED_RECORD";
  ```
2. move into a common constant at the top of the file

LR2/LR2_songmanage.cpp
```diff
@@ -1075,7 +1078,14 @@ int GetSongData(CSTR songMD5, SONGDATA *song, sqlite3 *sql, SONGSELECT *ss) {
...
 				if (song->mybest.minbp == 0 && song->mybest.clear != 5)
-					song->mybest.minbp = -1; 
+					song->mybest.minbp = -1;
+
+				if (isIRDerivedRecord) {
+					song->mybest.isIRDerivedRecord = 1;
+				}
+				else {
+					song->mybest.isIRDerivedRecord = 0;
+				}
```
```suggestion
				song->mybest.isIRDerivedRecord = isIRDerivedRecord ? 1 : 0;
```

chown2: why is it an int to begin with? should just be a 'bool'

Unengine: I agree.

Unengine:
@chown2 Addressed your suggestions.
By the way, I am currently working on an external player.db builder.
It simply updates player.db with bokutachi IR's personal best scores.
Should I make it a separate repository, or put it inside this repository (with a new pr)?
It is written in C#, and I can publish it after working.
IMO separate repository seems more legit.

chown2:
I think the importer should be in a separate repo because OpenLR2 shouldn't be tied to Tachi specifically

LR2/LR2_songmanage.cpp
```diff
@@ -18,6 +18,7 @@
 #endif // _WIN32
 
 int EnabledInsane;
+constexpr auto&& IR_DERIVED_RECORD_HASH = "IR_DERIVED_RECORD";
```
```suggestion
static constexpr auto&& IR_DERIVED_RECORD_HASH = "IR_DERIVED_RECORD";
```
sorry :D

chown2:
I don't wanna figure out what all of those score loading blocks actually do, but did you make sure that an imported score doesn't get a valid 'scorehash' unless the score is an exscore PB?

luxurya111:
Would it be fine if users cannot check on the UI whether the score was fetched from the IR or not?

Unengine:
> I don't wanna figure out what all of those score loading blocks actually do, but did you make sure that an imported score doesn't get a valid 'scorehash' unless the score is an exscore PB?

I found out every stat in the chart's record is used to determine the scorehash's validity.
It means, not only exscore but also minbp, clear lamp, playcount also matters.
So if there are any updates, scorehash should be forced to be set to `'IR_DERIVED_RECORD'`.
Also, Addressed your review. tysm!

Unengine:
> Would it be fine if users cannot check on the UI whether the score was fetched from the IR or not?

If a select skin can access to the scorehash, I think it can be checked. `scorehash=='IR_DERIVED_RECORD'`
Otherwise for now, if a chart's exscore is updated, the ghost is set to null or fetched from the IR.
For example, bokutachi IR doesn't support ghost.
So fetching from bokutach IR will remove the ghost if exscore is updated.
I could check it directly through the updated player.db or removed ghost or clear lamp.
I think it's fine, because the absence of the ghost can be the conclusive clue.

MatVeiQaaa:
> > Would it be fine if users cannot check on the UI whether the score was fetched from the IR or not?
> 
> If a select skin can access to the scorehash, I think it can be checked. `scorehash=='IR_DERIVED_RECORD'`

LR2 skins can't do complex logic like this. However this is very trivial to implement as a boolean skin accessor.

Unengine:
> > > Would it be fine if users cannot check on the UI whether the score was fetched from the IR or not?
> > 
> > 
> > If a select skin can access to the scorehash, I think it can be checked. `scorehash=='IR_DERIVED_RECORD'`
> 
> LR2 skins can't do complex logic like this. However this is very trivial to implement as a boolean skin accessor.

@MatVeiQaaa 
So I just need to add a case here, right? I don't know how `op` parameter works, so I just added case 300.
https://github.com/GOMazk/OpenLR2/blob/96efc709a4d89a10c4cc361b3ff97963013d31e2/LR2/LR2_skinobject.cpp#L1625

MatVeiQaaa:
> > > > Would it be fine if users cannot check on the UI whether the score was fetched from the IR or not?
> > > 
> > > 
> > > If a select skin can access to the scorehash, I think it can be checked. `scorehash=='IR_DERIVED_RECORD'`
> > 
> > 
> > LR2 skins can't do complex logic like this. However this is very trivial to implement as a boolean skin accessor.
> 
> @MatVeiQaaa So I just need to add a case here, right? I don't know how `op` value works, so I just added case 300.
> 
> https://github.com/GOMazk/OpenLR2/blob/96efc709a4d89a10c4cc361b3ff97963013d31e2/LR2/LR2_skinobject.cpp#L1625

SetObjectValue_Num is used to get a value for #DST_NUMBER.
This isn't a good idea, add it to dst_option instead, then it can be used as condition to draw a texture (exactly what needed).

In GetOptionFlag_dst:
`
case 626:
	if (gs->sSelect.bmsList[gs->sSelect.cur_song].mybest.isIRDerivedRecord) return ret;
	break;
`

id 626 fits, as its in near the numbers belonging to various IR-related information. You can consider other free indexes too, if you think something fits better. You can reference existing indexes list [here](https://lr2skinhelp.vercel.app/dstop.html)

Unengine:
> this isn't a good idea, add it to dst_option instead, then it can be used as condition to draw a texture (exactly what needed).
> 
> In GetOptionFlag_dst: `case 626: if (gs->sSelect.bmsList[gs->sSelect.cur_song].mybest.isIRDerivedRecord) return ret; break;`
> 
> id 626 fits, as its in near the numbers belonging to various IR-related information. You can consider other free indexes too, if you think something fits better. You can reference existing indexes list [here](https://lr2skinhelp.vercel.app/dstop.html)

Ok, I just addressed your feedback and committed the change. Could you check on it? tysm :D

LR2/LR2_skinobject.cpp
```diff
@@ -1417,6 +1417,9 @@ bool GetOptionFlag_dst(game *gs, int option) {
 		case 625:
 			if (gs->sSelect.bmsList[gs->sSelect.cur_song].rivalRecord.stat_exscore > 0) return ret;
 			break;
+		case 626:
+			if (gs->sSelect.bmsList[gs->sSelect.cur_song].mybest.isIRDerivedRecord > 0) return ret;
```
MatVeiQaaa: your variable is a bool already, you don't need to compare against 0.

MatVeiQaaa (dismissed):

LR2/LR2_skinobject.cpp
MatVeiQaaa: you can also leave a newline between case 625 and 626, they are grouped logically.

Unengine: So I just need to return the value itself, right?

LR2/LR2_skinobject.cpp
```diff
@@ -1418,7 +1418,7 @@ bool GetOptionFlag_dst(game *gs, int option) {
 			if (gs->sSelect.bmsList[gs->sSelect.cur_song].rivalRecord.stat_exscore > 0) return ret;
 			break;
 		case 626:
-			if (gs->sSelect.bmsList[gs->sSelect.cur_song].mybest.isIRDerivedRecord > 0) return ret;
+			return gs->sSelect.bmsList[gs->sSelect.cur_song].mybest.isIRDerivedRecord;
```
MatVeiQaaa: no, you shouldn't return the value itself, because then you strip it of a feature of negating the condition. It can be specified as !dst_option or -dst_option in the skin.csv. To make this nuance work, you must return `ret`. So, it must be `if (gs->sSelect.bmsList[gs->sSelect.cur_song].mybest.isIRDerivedRecord) return ret;`

Unengine: tysm for the feedback! I just committed your feedback.

Unengine:
### Before Importing : My satelite sl3 folder contains noplay charts.
LR2 2026-06-20 14-44-48

### Importing new pbs
image

### After Importing : I'm done with those sl3 charts!
LR2 2026-06-20 14-46-44
image

### Also, stats are updated.

Previous LR2 record : BEST EXSCORE 3314, MinBP 100
image

Bokutachi record : BEST EXSCORE 3314, MinBP 86
image

EXSCORE is still 3314, but the miss count is updated!
image

Since previous best exscore is LR2's record, so the ghost is preserved but the scorehash is updated.
image

luxurya111:
There is one more thing I'm concerned about.

Assuming that the data fetched from another IR is:
"clear exhard, exscore 2000, minbp 30, maxcombo 1500/2000, option non-random/exhard"

And the data from the previous LR2IR (reference from lr2ir.com) was played as:
"clear normal, exscore 1800, minbp 35, maxcombo 1200/2000, option random/groove"

If you play a new round and get the following result:
"clear normal, exscore 2050, bp 45, maxcombo 1000/2000, option mirror/groove"

When developing a Custom IR, is it possible to derive the following result during the process of calculating the previous PB?
"clear normal, exscore 2050, bp 35, maxcombo 1200/2000, option mirror/groove"

The intent behind this question is as follows:

1. Records from an IR like Bokutachi are strictly for reference, and there might be cases where one only wants to refer to the records actually played on LR2.
2. As far as I know, the stock LR2 can also transmit records by referring to the existing player.db. (For instance, if an exscore of 2000 was the PB and the user wanted to transmit it, but the transmission failed, it transmits the previous PB upon re-transmission.) Regarding this feature, when transmitting to the IR, are Bokutachi's records isolated separately?

To give a proper example, although it's a different client, I have previously imported LR2's local records into beatoraja. In that case, because they are completely separated, even if you update the exscore/combo/bp, it will absolutely never be updated based on LR2's records.

Unengine:
> The intent behind this question is as follows:
> 
> 1. Records from an IR like Bokutachi are strictly for reference, and there might be cases where one only wants to refer to the records actually played on LR2.
> 2. As far as I know, the stock LR2 can also transmit records by referring to the existing player.db. (For instance, if an exscore of 2000 was the PB and the user wanted to transmit it, but the transmission failed, it transmits the previous PB upon re-transmission.) Regarding this feature, when transmitting to the IR, are Bokutachi's records isolated separately?

Since you can't store two separate records with same md5, so if any records are overwritten, the scorehash is set to 'IR_DERIVED_RECORD' and that record won't be sent to IR.
But if any of the stat is changed, the scorehash is changed and the score would be sent to IR.
Using my importer can cause LR2 to send the record like this:
"clear normal, exscore 2050, bp 30, maxcombo 1500/2000, option mirror/groove"

But I think it depends on the importer's policy.
If a bokutachi importer only imports the clear lamp and judges (pgr, gr, gd, bd, pr), it should be fine.
It will be calculated like this:
"clear normal, exscore 2050, bp 35, maxcombo 1200/2000, option mirror/groove"

So the thing is, the records are not isolated. (because md5 hash is primary key in player.db)
But any of the stat changes by gameplay can make the record legit.
It completely depends on the importer's policy, and as you concerned, matter of trust can be happened.

luxurya111:
> But I think it depends on the importer's policy. If a bokutachi importer only imports the clear lamp and judges (pgr, gr, gd, bd, pr), it should be fine. It will be calculated like this: "clear normal, exscore 2050, bp 35, maxcombo 1200/2000, option mirror/groove"
> 
> So the thing is, the records are not isolated. (because md5 hash is primary key in player.db) But any of the stat changes by gameplay can make the record legit. It completely depends on the importer's policy, and as you concerned, matter of trust can be happened.

Exactly. Ultimately, the issue I want to raise is that the scorehash is updated not just for an exscore update, but also for various other updates such as maxcombo and minbp. The moment the scorehash is overwritten due to these updates, records from other IRs end up being recognized as LR2 records, which could become highly controversial.

In the future, when others implement the IR, it must be ensured that they can retrieve proper LR2 records as well. Alternatively, a separate isolation policy should be established so that anyone planning to implement a Custom IR down the road can choose whether to fetch only LR2's local records or to also allow records from other IRs.

Unengine:
I've submitted a PR regarding the IR score integration.
I’ve introduced a new table, `imported_score`, in `player.db` to handle imported records separately.
The logic now compares these IR scores with the original scores, allowing the skin to display the better stats of the two. Additionally, I’ve updated the folder lamp logic to account for these imported scores.
Please take your time reviewing the changes. Thank you all for the great onboarding!

LR2/LR2_ghost.cpp
```diff
@@ -764,6 +764,30 @@ CSTR ReadGhost(sqlite3 *sql, CSTR songMD5) {
 	return ghostdata;
 }
 
+CSTR ReadIRGhost(sqlite3* sql, CSTR songMD5) {
```
chown2: To avoid spreading the logic, this should better be handled within ReadGhost itself

Unengine: Then, should I change the signature like this?
As you can see, the query to load the raw data of a ghost is separated.
`SELECT ghost FROM score WHERE hash = ?` (ReadGhost)
`SELECT ghost FROM imported_score WHERE hash = ?` (ReadIRGhost)

```suggestion
CSTR ReadIRGhost(sqlite3* sql, CSTR songMD5, bool useIRGhost) {
```

MatVeiQaaa: > Then, should I change the signature like this? As you can see, the query to load the raw data of a ghost is separated. `SELECT ghost FROM score WHERE hash = ?` (ReadGhost) `SELECT ghost FROM imported_score WHERE hash = ?` (ReadIRGhost)

existing `ReadGhost` should be modified to support both use cases. Boolean flag argument would be fine, to then do smth like `std::string querry = std::format("SELECT ghost FROM {} WHERE hash = ?", flag ? "imported_score" : "score");`. Don't name the argument "flag" though, it's just an example where i leave the name to your imagination.

Unengine: Sounds brilliant. I've addressed your feedback, could you take a quick look?
I removed two duplicate functions `ReadIRGhost` and `ReadIRGhostToScore` and merged them into original func.

MatVeiQaaa (approved):

GOMazk (approved):

LR2/Scene04_Play.cpp
```diff
@@ -1905,7 +1905,11 @@ int ProcS_Play(game *g, sqlite3* sql) {
 
 	g->gameplay.highScore.InitJudgeQueue();
 	if (g->gameplay.isGhostDisabled == 0) {
-		ReadGhostToScore(sql,md5,&g->gameplay.highScore);
+		const SONGDATA& songData = g->sSelect.bmsList[g->sSelect.cur_song];
+
+		// Use IR Ghost only if its exscore is higher.
```
chown2: Pointless comment
```suggestion
```

LR2/LR2_songmanage.cpp
```diff
@@ -223,6 +224,98 @@ bool ParseTextCommand(BMSMETA& meta, CSTR& line) {
 	return false;
 }
 
+// IR Records are locally compared and updated to the original record in lr2.
+void LoadIRDerivedRecord(sqlite3* sql, SONGDATA& sd) {
+	const char* sqlQuery = "SELECT * FROM imported_score WHERE hash = ?;";
```
chown2: Use field names explicitly instead of *. That was a really bad practice in LR2.

LR2/LR2_ghost.cpp
```diff
@@ -764,8 +766,8 @@ CSTR ReadGhost(sqlite3 *sql, CSTR songMD5) {
 	return ghostdata;
 }
 
-int ReadGhostToScore(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score) {
-	CSTR ghostdata = ReadGhost(sql, songMD5);
+int ReadGhostToScore(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score, bool useImportedGhost = false) {
```
chown2: This is only needed in headers
```suggestion
int ReadGhostToScore(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score, bool useImportedGhost) {
```

LR2/LR2_ghost.cpp
```diff
@@ -740,11 +740,13 @@ int WriteGhostInDatabase(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score) {
 	return 0;
 }
 
-CSTR ReadGhost(sqlite3 *sql, CSTR songMD5) {
+CSTR ReadGhost(sqlite3 *sql, CSTR songMD5, bool useImportedGhost = false) {
```
chown2:
```suggestion
CSTR ReadGhost(sqlite3 *sql, CSTR songMD5, bool useImportedGhost) {
```

LR2/LR2_ghost.cpp
```diff
@@ -774,4 +776,4 @@ int ReadGhostToScore(sqlite3 *sql, CSTR songMD5, PLAYSCORE *score) {
 	
 	score->DecodeGhostData(ghostdata);
 	return 1;
-}
+}
```
chown2: Our code editors are at war here :D Weirdly enough, only one file's ending changed in your PR

LR2/structure.h
```diff
@@ -462,7 +462,9 @@ struct SONGDATA { /* 712bytes */
 	int courseIR{}; /* IR flag */
 	int grHandle{};
 	struct STATUS mybest;
+	struct STATUS myIRbest;
 	struct STATUS rivalRecord;
+	bool hasIRDerivedRecord{};
```
chown2: This will involve a little refactoring, but will make sure we don't ever use the incorrect one by accident
```suggestion
	// If present, this will contain score for the song imported from the IR with best properties mixed in with 'mybest'.
	std::optional<STATUS> myIRbest;
	STATUS mybest;
	STATUS rivalRecord;
```

Unengine:
@chown2 I’ve addressed your feedback. Additionally, I noticed that IR records weren't being sorted by clear lamps, so I’ve included a quick fix for that as well. Could you please take another look?

LR2/LR2_songmanage.cpp
chown2: Thanks, put please move the list of fields into this string itself, having it separate hurts readability and greppability

LR2/LR2_skinobject.cpp
```diff
@@ -19,13 +19,14 @@ bool GetOptionFlag_dst(game *gs, int option) {
 	int t = 0;
 	bool ret = (option >= 0);
 	if (!ret) option = -option;
-
+	const SONGDATA& songData = gs->sSelect.bmsList[gs->sSelect.cur_song];
+	const std::optional<STATUS>& myIRbest = songData.myIRbest;
```
chown2:
```suggestion
	const STATUS& bestmybest = songData.myIRbest.has_value() ? *songData.myIRbest : songData.mybest;
```
And you can use that instead of branching everywhere

chown2:
Also add `#include <optional>` to structure.h

LR2/LR2_skinobject.cpp
Unengine: That's a brilliant idea! I should have known it earlier XD
Branching in skinobject was a nightmare.

Unengine:
@chown2
I've addressed your feedback:
Moved the list of fields into the query itself instead of formatting it, simplified the branching in skinobject, and included `<optional>` in structure.h.
Are there any further suggestions or improvements you'd recommend?

LR2/Scene02_Songselect.cpp
```diff
@@ -2621,23 +2621,32 @@ int ProcI_Select(game *g, sqlite3 *sql) {
 					}
 				}
 
+				bool hasIRDerivedRecord = g->sSelect.bmsList[bar].myIRbest.has_value();
```
```suggestion
				const STATUS& mybest = g->sSelect.bmsList[bar].myIRbest.has_value() ? *g->sSelect.bmsList[bar].myIRbest : g->sSelect.bmsList[bar].mybest;
```

LR2/structure.h
```diff
@@ -6,6 +6,7 @@
 #include <thread>
 #include <unordered_map>
 #include <vector>
+#include <optional>
```
```suggestion
#include <array>
#include <future>
#include <mutex>
#include <optional>
#include <thread>
#include <unordered_map>
#include <vector>
```

chown2: Sort includes. Did GitHub break the suggestion?

chown2 (approved):
Fix the two little nits and then LGTM. 🚀 🚀 🚀
This should be squash-merged.

LR2/structure.h
Unengine: Seems like it did.

Unengine:
> Fix the two little nits and then LGTM. 🚀 🚀 🚀 This should be squash-merged.

Thanks, I just fixed all of them. If they are looking good to you, I also think it's good to go.

chown2 (approved):
🚀 🚀 🚀 
GOMazk squash merge plss

Unengine:
I personally asked GOMazk if i can I merge it, and he approved. So I squash merge it. tysm for the feedback guys!

PR #143, opened 2026-06-20
feat(customir): add result rank and leaderboards

chown2:
#112 recreated, I force pushed into it incorrectly and it become impossible to reopen it
Sorry


LR2/LR2_configsave.cpp
```diff
@@ -583,6 +583,10 @@ int WriteOpenLr2ConfigXml(game *g, const char *filename){
 	WriteXML_Tab2Str(pFile, "courseresult", (g->config).skin.skinFilePath[15]);
 	fputs("\t</skin>\n", pFile);
 
+	fputs("\t<network>\n", pFile);
+	WriteXML_Tab2Str(pFile, "customir_provider", (g->config).network.customIrProvider);
```
luxurya111: There was a mention of unifying variable names in a previous PR comment, and I personally recommend using "provider_ir" (mProviderIR) or "display_ir" (mDisplayIR) as suggested by @MatVeiQaaa.

LR2/LR2_customir_api.h
```diff
@@ -113,8 +115,76 @@ enum class SendScoreStatus: int {
...
+	int combo{};
+	int pg{};
+	int gr{};
+	int gd{};
+	int bd{};
+	int pr{};
+	int minbp{};
+	int option{};
+	int playcount{};
+	// TODO(stable-api): figure out if this makes sense. It's redundant by IRRankResult::ranking index.
```
luxurya111: Looking at how that variable is handled in RANKING::ParseXML, it might be a specification that is unnecessary for a Custom IR. I think it's fine to remove it. The only reason I used that variable was simply to utilize the original structure as it was.

ExampleIR/ExternalIR.cpp
```diff
@@ -14,25 +15,46 @@ namespace State {
...
+    std::println("Login");
     return true;
 }
 
-static SendScoreStatus SendScore(const IRScoreV1& score) {
-    // Process your score here, and output the result where you want it, perhaps send it to a URL.
-    // This method is ran on its own thread at each score result, both for normal plays and courses.
-    // It will run even for scores that wouldn't be sent to LR2IR or saved to the score.db, it's up to the module to filter them.
+static GetStatus RestoreCachedRank(const char* songHash, IRRankResult& out) {
+    std::println("RestoreCachedRank({})", songHash);
```
luxurya111: It seems that the print code is causing a crash. It occurs when scrolling on the song selection screen.

LR2/LR2_customir.cpp
```diff
@@ -3,12 +3,19 @@
 #include "LR2_customir_api.h"
 #include "LR2_songmanage.h"
 #include "structure.h"
+#include "En_input.h"
 
 #include <algorithm>
+#include <array>
 #include <cmath>
+#include <chrono>
+#include <cmath>
```
luxurya111, line 10:
```suggestion
```
Duplicate include

ExampleIR/ExternalIR.cpp
MatVeiQaaa: > It seems that the print code is causing a crash. It occurs when scrolling on the song selection screen.

do `std::println(std::cout, fmt, ...)` or use `fflush(stdout)` after std::println. Or better yet, use `ErrorLogFmtAdd(fmt, ...)`

luxurya111:
I will also proceed with testing on StellaverseIR. I will edit this message once it is complete.

-- PM3:55 KST
It seems to be working well.

ExampleIR/ExternalIR.cpp
chown2: Classic void stdout, still wondering if it's possible to make println work properly out of the box on Windows

chown2:
I have applied the changes, thanks!
https://github.com/GOMazk/OpenLR2/compare/ed31f4e7c9a0e73dc42995e1cce90a6994bd1717..5e2dcb88345a1c586779714021ea686f642b4c8d
Is the rest to your liking?
I've removed 'ranking' from IRRankPlayer so the old StellaverseIR DLL will definitely not work with new client build. We can add a dummy field like 'reserved' akin to what I did to MethodTable, but I'm not sure if it's needed considering the 'Experimental API' notes.
My plan is to remove the Experimental API note when we get those methods implemented for the Bokutachi IR module so we can have several examples of method usability.

LR2/Scene05_Result.cpp
```diff
@@ -271,10 +279,37 @@ int ProcI_Result(game *g) {
...
+						// instead
+						// NOTE: SetObjectString(20, ...) will be overridden with the value
+						// of g->net.IRresultMessage on result exit somewhere down the line.
+						// But it's visible while result is open, so it can already display an
+						// error, and I think it serves as a decent example.
+						if (std::optional<IRRankResult> result = resultFuture.get(); result)
+						{
+							openlr2::fill_ranking_from_customir(*result, g->net.rankingData);
+							openlr2::fill_status_from_ranking(g->net.rankingData, false, g->sSelect.bmsList[g->sSelect.cur_song].mybest);
+							// Force 'IR READY' so that we can open the leader boards.
```
chown2: Not actually used, RestoreCachedRank is called on song select entracnce

luxurya111:
I have confirmed it.
As for the ranking field, I will take separate measures so that StellaverseIR does not use it either.
On a side note, in the past, I used the length of ranking[] to determine the length of the leaderboard, but thinking back, determining it by the length of ranking[] probably wasn't a good idea. (Why did past-me suggest determining the leaderboard length by its length back then? Was it because of the cache size?) Therefore, I will change the specification in StellaverseIR to allow more than 100 ranking data items, and for ranking[] that exceeds 999, it will retrieve a maximum of up to 999 data entries.

Everything looks good from my end. Thank you for your hard work!

LR2/LR2_customir_api.h
```diff
@@ -113,8 +115,74 @@ enum class SendScoreStatus: int {
...
+	int minbp{};
+	int option{};
+	int playcount{};
+};
+
+// \warning Experimental API, may be changed.
+struct IRRankResult {
+	int myRank{};
+	int totalPlayer{};
+	uint64_t lastupdate{}; // Seconds since Unix Epoch
```
chown2: Should we have a timestamp in IRRankPlayer too? There's nowhere to plug it into right now but it will definitely come in useful in future. Imagine scrolling through leaderborads and seeing '4 weeks ago', '10 seconds ago'. :D @luxurya111

LR2/LR2_customir_api.h
```diff
@@ -113,8 +115,74 @@ enum class SendScoreStatus: int {
 	Fail,
 };
 
+// \warning Experimental API, may be changed.
+struct IRRankPlayer {
+	std::string name;
+	std::string comment;
+	int id{};
+	int sp{};
+	int dp{};
```
chown2: Not actually used anywhere and it's not clear what they mean. LR2IR didn't return them too, they were just placeholders.
https://github.com/GOMazk/OpenLR2/blob/96efc709a4d89a10c4cc361b3ff97963013d31e2/LR2/structure.h#L1489-L1490

LR2/LR2_customir_api.h
```diff
@@ -113,8 +115,74 @@ enum class SendScoreStatus: int {
...
+	int clear{};
+	int notes{};
+	int combo{};
+	int pg{};
+	int gr{};
+	int gd{};
+	int bd{};
+	int pr{};
+	int minbp{};
+	int option{};
```
chown2: Not actually used anywhere and the meaning isn't complete clear as LR2IR didn't return this, so it was another placeholder in LR2.
https://github.com/GOMazk/OpenLR2/blob/96efc709a4d89a10c4cc361b3ff97963013d31e2/LR2/structure.h#L1500
I guess it would make sense to follow the format of `op_best` in the score DB, we don't have anything else to show there at the moment anyway.
https://github.com/GOMazk/OpenLR2/blob/96efc709a4d89a10c4cc361b3ff97963013d31e2/LR2/LR2_statplay.cpp#L1016

LR2/LR2_customir_api.h
luxurya111: Wow, I'm sure it will be helpful.
If you plan to expand the displayed information, I'm certain there will be even better ideas.
I will also ask the testers at StellaverseIR.

LR2/LR2_customir_api.h
luxurya111: As mentioned previously, since it is a direct migration of the structure used in LR2IR, I believe there will be some data whose usage remains unclear.

For now, I will review each and every one of the suggestions you provided.

LR2/LR2_customir_api.h
luxurya111: Mentioned above, how about adding it so that it can be displayed on the host side?
Of course, if this is to be implemented, I can do so in a separate branch. (Skin definitions will also be required.)

LR2/LR2_customir_api.h
luxurya111: I think the intended purpose I had in mind was a ranking option that reflects cases where DP is converted to SP, or vice versa. But yes, since it is not currently in use, I think it's fine to delete it.

LR2/LR2_customir_api.h
```diff
@@ -113,8 +115,74 @@ enum class SendScoreStatus: int {
 	Fail,
 };
 
+// \warning Experimental API, may be changed.
+struct IRRankPlayer {
+	std::string name;
+	std::string comment;
```
luxurya111: IMO comment are also information that only needs to be verified on the IR, and there's probably no strict need to read them within the client.

LR2/LR2_customir_api.h
luxurya111: For now, it seems like the used options (gauge/random options), lastupdate, and rseed are about it.
Do you plan to implement this in the current PR? If it's something to be implemented later, it might be a good idea to gather feedback in the Discussions first before moving forward.

LR2/LR2_customir_api.h
chown2: Removed them

LR2/LR2_customir_api.h
chown2: Added timestamp, for other new ideas make new comments

LR2/LR2_customir_api.h
chown2: Repurposed as described

LR2/LR2_customir_api.h
chown2: In an ideal world we wouldn't ever force the user to look at the IR web-site, everything would be possible to show in-game, if the skin supports it

luxurya111: Understood. Either way is fine with me.

LR2/LR2_customir_api.h
chown2: @luxurya111 sorry GitHub didn't display your comment until page refresh. Yes, we can display the information, but in a follow up PR, for now lets think of the API itself.

chown2:
https://github.com/GOMazk/OpenLR2/compare/ca5f758fc41f73255b8cf5c25df5f43ef05bbd0a..1828942c9c849cc62d773fdd7a741478dcca0335
Added timestamp and inputType
Input type would be useful for BM players who want to hide KB player's scores, we can implement such filter later on using this.
Rseed is missing. It's a bit tricky. The displayed play may not necessarily be from LR2. It could be from raja, which obviously has different RNG implementation and the seeds are different.
Therefore, we should probably add the random layout in addition to rseed.

luxurya111:
> https://github.com/GOMazk/OpenLR2/compare/ca5f758fc41f73255b8cf5c25df5f43ef05bbd0a..1828942c9c849cc62d773fdd7a741478dcca0335 Added timestamp and inputType Input type would be useful for BM players who want to hide KB player's scores, we can implement such filter later on using this. Rseed is missing. It's a bit tricky. The displayed play may not necessarily be from LR2. It could be from raja, which obviously has different RNG implementation and the seeds are different. Therefore, we should probably add the random layout in addition to rseed.

That's right. The rseed is information that can only be obtained by looking up a specific user's ghost on the IR and verifying the details inside that ghost, so it could be difficult. In other words, transmitting the ghost data itself from the IR is possible.
If it's a different IR like Bokutachi, the structure of the rseed itself might be completely different, so it would probably be tough.

Therefore, like the other options, it would be best to look for a solution down the road.

chown2:
https://github.com/GOMazk/OpenLR2/compare/4ac4069a5fda68d5803570c318571a8e09ef2461..bfe5a536efc6e7610f4e9210e85b8eaa8a060f72
Breaking API changes: I reordered the fields and moved new symbols into the 'openlr2' namespace. InputType is now an enum.

I added 'rseed' and 'randomLayout' fields.
'rseed' can be supplied from the IR if it knows the actual LR2 rseed, so that correct #RANDOM and random layout can be applied.
In case the IR doesn't know the rseed, it can supply the random layout instead. We can then calculate a matching rseed ourselves using the unrandomizer feauture, although #RANDOM will not be matching then.
The actual code to apply the seed isn't there, we will do it in a follow-up PR, for now just the API.
This is enough to start G-Battle with matching random but without ghost.
I think we should get ghost alone using a separate new method since ghost is essentially a stripped down replay, and replays are big enough to deserve a separate method. 
Please, tell me if that sounds convincing.

LR2/LR2_customir_api.h
```diff
@@ -113,8 +115,100 @@ enum class SendScoreStatus: int {
...
+	// Second one is useless.
+	std::vector<IRRankPlayer> ranking;
+	std::array<int, 6> clearPlayers{};
+	uint64_t lastupdate{};
+	// XXX: just find the current player in the list? To make the method stateless.
+	int myRank{};
+	int totalPlayer{};
+	int totalPlaycount{};
+	// XXX: just find the rival player in the list? To make the method stateless.
+	int rivalRank{};
```
chown2: TODO. myRank is fine imo, but to get correct rivalRank we will have to pollute the method with rival data, which seems out of place to me

LR2/LR2_customir_api.h
chown2: @MatVeiQaaa what are your thoughts on the field being a number calculated in such way? This could be 4 separate fields instead.

LR2/LR2_customir_api.h
chown2: We could add a new 'Login2' method that would return us the numeric user ID, so that we can then match the user.
And rivalRank feels to me like something on another abstraction level, IR shouldn't even be aware of it, it should just be a song select concept.

luxurya111:
> https://github.com/GOMazk/OpenLR2/compare/4ac4069a5fda68d5803570c318571a8e09ef2461..bfe5a536efc6e7610f4e9210e85b8eaa8a060f72 Breaking API changes: I reordered the fields and moved new symbols into the 'openlr2' namespace. InputType is now an enum.
> 
> I added 'rseed' and 'randomLayout' fields. 'rseed' can be supplied from the IR if it knows the actual LR2 rseed, so that correct #RANDOM and random layout can be applied. In case the IR doesn't know the rseed, it can supply the random layout instead. We can then calculate a matching rseed ourselves using the unrandomizer feauture, although #RANDOM will not be matching then. The actual code to apply the seed isn't there, we will do it in a follow-up PR, for now just the API. This is enough to start G-Battle with matching random but without ghost. I think we should get ghost alone using a separate new method since ghost is essentially a stripped down replay, and replays are big enough to deserve a separate method. Please, tell me if that sounds convincing.

If we can choose between rseed or randomLayout and use either one, I have no objections.

LR2/LR2_customir_api.h
luxurya111: > TODO. myRank is fine imo, but to get correct rivalRank we will have to pollute the method with rival data, which seems out of place to me

The rivalRank field was added without going through the process described later.

As far as I know, there is probably a rivalID that can identify the rival's ID. If we proceed by using this ID to iterate through the ranking[] array and find it, I think it's fine to delete it.

> We could add a new 'Login2' method that would return us the numeric user ID, so that we can then match the user.

If that's possible, we might not even need myRank in the subsequent steps.

chown2: Since 'ranking' is ususally going to be cut off at 999, it makes sense to keep current player's position separately to avoid having to retrieve the whole list.

Same goes for the rival's position, but I'm not sure how to best let know the IR module of the current rival.
It's probably best to force GetResultRank and RestoreCachedRank to have an 'int rival' parameter.

chown2: Proposal: add 'int reserved' param to GetResultRank and RestoreCachedRank, think about this when we actually implement rivals :D

luxurya111: LGTM. It is ideal to have the position looked up from the outside of GetResultRank and RestoreCachedRank.

chown2:
https://github.com/GOMazk/OpenLR2/compare/bfe5a536efc6e7610f4e9210e85b8eaa8a060f72..9d0fbafe78bdbed68da28b4cc71e62969fb09a4e
Breaking API change: added int reserved to RestoreCachedRank and GetResultRank. Removed IRRankResult::rivalRank for now adding a placeholder instead.
Added some 'reserved' fields so we can extend the API a little without breaking ABI.
From my end, I consider this ready :pig2: :rocket:, other than waiting for some feedback from @MatVeiQaaa, if any. :)

LR2/LR2_customir_api.h
MatVeiQaaa: > @MatVeiQaaa what are your thoughts on the field being a number calculated in such way? This could be 4 separate fields instead.

it would be infinitely easier to use on any side if it was separate fields. I don't see a single reason to bitmask it.

MatVeiQaaa:
> https://github.com/GOMazk/OpenLR2/compare/ca5f758fc41f73255b8cf5c25df5f43ef05bbd0a..1828942c9c849cc62d773fdd7a741478dcca0335 Added timestamp and inputType Input type would be useful for BM players who want to hide KB player's scores, we can implement such filter later on using this.

My input is that such filter should never be implemented, as it serves no good purpose and just entices people to divide.

chown2:
Fair enough, but it can still an interesting metric if hooked up into skinning

luxurya111:
As for the filters, I interpreted it like this: when implementing the IR, it would be left up to the IR developers to display the data by option (All, KB, BM, MIDI) based on the logged-in user's preferences. (of course, totalPlayer, ranking, and other fields would also change accordingly.)

In my case, as long as forcing the host to display only KB or BM separately isn't mandatory, I thought it would be fine. Since it doesn't seem like the filters are being implemented right away, it might be better to gather some opinions if things are uncertain.

chown2:
Split the int option bitmask into separate fields and used enums. Should be ready!

chown2:
> As for the filters, I interpreted it like this: when implementing the IR, it would be left up to the IR developers to display the data by option (All, KB, BM, MIDI) based on the logged-in user's preferences. (of course, totalPlayer, ranking, and other fields would also change accordingly.)

Filtering scores out on the IR side sounds nuclear IMO, if there's any filtering, it should be on our side.
Regardless, this is just laying out the groundwork so that such idea, if anyone wants to implement it, isn't impossible to do.
We aren't implementing it right now.

PR #145, opened 2026-06-21
feat: send ghost + play w/ ghost in battle

luxurya111:
This PR includes features to transmit Ghosts and support battles with them.

In this process, I also fixed the ghost data resetting issue that @MatVeiQaaa  pointed out previously. This task includes the refactoring of LR2IR (handling ghostString by receiving it as a parameter) and the process of saving the ghost separately in MakeScoreV1 and initializing the buffer before proceeding with the IR transmission.

The ghostData is currently an LR2-exclusive specification, and we currently support sending and receiving ghostData generated in LR2. (We might need to modify this in the future if other IRs require different Ghost specifications.)

### DIFF
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -87,6 +87,23 @@ static openlr2::GetStatus GetResultRank(const char* songHash, int /*reserved*/, 
     return openlr2::GetStatus::Ok;
 }
 
+static openlr2::GetStatus GetGhost(const IRScoreV1& score, int mode, int /*targetPlayerId*/, IRGhostResult& out) {
+    std::println(std::cout, "GetGhost({{.song.hash={}}}, mode={})", score.song.hash, mode);
+    out = {};
+    if (mode == 8) {
+        out.averageExscore = 800;
+        return openlr2::GetStatus::Ok;
+    }
+    out.displayName = "EXAMPLE";
+    out.ghostData = "#EXAMPLE,1,0,LXZ,"; // this ghost will do almost perfect play.
+    out.optionDigit1 = 1;
+    out.optionDigit2 = 0;
+    out.optionDigit3 = 0;
+    out.optionDigit4 = 0;
+    out.randomSeed = 0;
+    return openlr2::GetStatus::Ok;
+}
+
 extern "C" OLR2_IR_EXPORT void GetMethodTable(MethodTable& table) {
     // Fill out the pointers to methods you want to use. Leave them at nullptr if you don't want to use them.
     // As API gets updated, new methods may appear available at MethodTable, but old ones will never be removed or their
@@ -96,6 +113,7 @@ extern "C" OLR2_IR_EXPORT void GetMethodTable(MethodTable& table) {
     table.SendScoreV1 = &SendScore;
     table.GetResultRank = &GetResultRank;
     table.RestoreCachedRank = &RestoreCachedRank;
+    table.GetGhost = &GetGhost;
 }
 
 #ifdef _WIN32
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -97,7 +97,7 @@ static openlr2::GetStatus GetGhost(const char* songHash, openlr2::GhostMode mode
         return openlr2::GetStatus::Ok;
     }
     out.displayName = "EXAMPLE";
-    out.ghostData = "#EXAMPLE,21,2179,LXZ,"; // this ghost will do almost perfect play.
+    out.ghostData = "LXZ"; // this ghost will do almost perfect play.
     out.gauge = openlr2::Gauge::Survival;
     out.randomOption = { openlr2::Random::Random, openlr2::Random::No };
     out.dpflip = false;
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -98,11 +98,10 @@ static openlr2::GetStatus GetGhost(const char* songHash, openlr2::GhostMode mode
     }
     out.displayName = "EXAMPLE";
     out.ghostData = "#EXAMPLE,1,0,LXZ,"; // this ghost will do almost perfect play.
-    out.gaugeOption = 1;
-    out.p1randomOption = 0;
-    out.p2randomOption = 0;
-    out.dpFlipOption = 0;
-    out.randomSeed = 0;
+    out.gauge = openlr2::Gauge::Survival;
+    out.randomOption = { openlr2::Random::No, openlr2::Random::No };
+    out.dpflip = false;
+    out.rseed = 0;
     return openlr2::GetStatus::Ok;
 }
 
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -101,7 +101,7 @@ static openlr2::GetStatus GetGhost(const char* songHash, openlr2::GhostMode mode
     out.gauge = openlr2::Gauge::Survival;
     out.randomOption = { openlr2::Random::No, openlr2::Random::No };
     out.dpflip = false;
-    out.rseed = 0;
+    out.rseed = 2179; // 1357246
     return openlr2::GetStatus::Ok;
 }
 
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -89,7 +89,7 @@ static openlr2::GetStatus GetResultRank(const char* songHash, int /*reserved*/, 
     return openlr2::GetStatus::Ok;
 }
 
-static openlr2::GetStatus GetGhost(const char* songHash, openlr2::GhostMode mode, int /*targetPlayerId*/, IRGhostResult& out) {
+static openlr2::GetStatus GetGhost(const char* songHash, openlr2::GhostMode mode, int /*targetPlayerId*/, openlr2::IRGhostResult& out) {
     std::println(std::cout, "GetGhost({})", songHash);
     out = {};
     if (mode == openlr2::GhostMode::Average) {
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -26,7 +26,7 @@ static bool Login() {
 }
 
 static SendScoreStatus SendScore(const IRScoreV1& score) {
-    std::println(std::cout, "SendScore({{.song.hash={}}})", score.song.hash);
+    std::println(std::cout, "SendScore({{.song.hash={}}}, ghostData={})", score.song.hash, score.ghostData);
     constexpr const char* lamps[6] = { "NO PLAY", "FAIL", "EASY", "NORMAL", "HARD", "FULL COMBO" };
     if (score.settings.assist[score.state.player]) return SendScoreStatus::Fail;
     std::string filename = std::format("score{}.txt", State::scoresSaved);
@@ -43,7 +43,8 @@ static SendScoreStatus SendScore(const IRScoreV1& score) {
         "fast: {}\n"
         "slow: {}\n"
         "cb: {}\n"
-        "lamp: {}\n",
+        "lamp: {}\n"
+        "ghostData: {}\n",
         score.song.hash, score.state.keymode, score.exscore,
         score.judgements_total.epg + score.judgements_total.lpg,
         score.judgements_total.egr + score.judgements_total.lgr,
@@ -51,7 +52,8 @@ static SendScoreStatus SendScore(const IRScoreV1& score) {
         score.judgements_total.ebd + score.judgements_total.lbd,
         score.judgements_total.epr + score.judgements_total.lpr,
         score.judgements_total.fast, score.judgements_total.slow, score.judgements_total.cb,
-        lamps[score.clearType]
+        lamps[score.clearType],
+        score.ghostData
     );
     std::ofstream dump(State::path / filename);
     dump << processedScore;
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -96,10 +96,10 @@ static openlr2::GetStatus GetGhost(const IRScoreV1& score, int mode, int /*targe
     }
     out.displayName = "EXAMPLE";
     out.ghostData = "#EXAMPLE,1,0,LXZ,"; // this ghost will do almost perfect play.
-    out.optionDigit1 = 1;
-    out.optionDigit2 = 0;
-    out.optionDigit3 = 0;
-    out.optionDigit4 = 0;
+    out.gaugeOption = 1;
+    out.p1randomOption = 0;
+    out.p2randomOption = 0;
+    out.dpFlipOption = 0;
     out.randomSeed = 0;
     return openlr2::GetStatus::Ok;
 }
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -26,7 +26,7 @@ static bool Login() {
 }
 
 static SendScoreStatus SendScore(const IRScoreV1& score) {
-    std::println(std::cout, "SendScore({{.song.hash={}}}, ghostData={})", score.song.hash, score.ghostData);
+    std::println(std::cout, "SendScore({{.song.hash={}}})", score.song.hash);
     constexpr const char* lamps[6] = { "NO PLAY", "FAIL", "EASY", "NORMAL", "HARD", "FULL COMBO" };
     if (score.settings.assist[score.state.player]) return SendScoreStatus::Fail;
     std::string filename = std::format("score{}.txt", State::scoresSaved);
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -97,9 +97,9 @@ static openlr2::GetStatus GetGhost(const char* songHash, openlr2::GhostMode mode
         return openlr2::GetStatus::Ok;
     }
     out.displayName = "EXAMPLE";
-    out.ghostData = "#EXAMPLE,1,0,LXZ,"; // this ghost will do almost perfect play.
+    out.ghostData = "#EXAMPLE,21,2179,LXZ,"; // this ghost will do almost perfect play.
     out.gauge = openlr2::Gauge::Survival;
-    out.randomOption = { openlr2::Random::No, openlr2::Random::No };
+    out.randomOption = { openlr2::Random::Random, openlr2::Random::No };
     out.dpflip = false;
     out.rseed = 2179; // 1357246
     return openlr2::GetStatus::Ok;
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -89,8 +89,8 @@ static openlr2::GetStatus GetResultRank(const char* songHash, int /*reserved*/, 
     return openlr2::GetStatus::Ok;
 }
 
-static openlr2::GetStatus GetGhost(const IRScoreV1& score, int mode, int /*targetPlayerId*/, IRGhostResult& out) {
-    std::println(std::cout, "GetGhost({{.song.hash={}}}, mode={})", score.song.hash, mode);
+static openlr2::GetStatus GetGhost(const char* songHash, int mode, int /*targetPlayerId*/, IRGhostResult& out) {
+    std::println(std::cout, "GetGhost({{}}, mode={})", songHash != nullptr ? songHash : "", mode);
     out = {};
     if (mode == 8) {
         out.averageExscore = 800;
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -89,10 +89,10 @@ static openlr2::GetStatus GetResultRank(const char* songHash, int /*reserved*/, 
     return openlr2::GetStatus::Ok;
 }
 
-static openlr2::GetStatus GetGhost(const char* songHash, int mode, int /*targetPlayerId*/, IRGhostResult& out) {
-    std::println(std::cout, "GetGhost({{}}, mode={})", songHash != nullptr ? songHash : "", mode);
+static openlr2::GetStatus GetGhost(const char* songHash, openlr2::GhostMode mode, int /*targetPlayerId*/, IRGhostResult& out) {
+    std::println(std::cout, "GetGhost({})", songHash);
     out = {};
-    if (mode == 8) {
+    if (mode == openlr2::GhostMode::Average) {
         out.averageExscore = 800;
         return openlr2::GetStatus::Ok;
     }
--- a/LR2/LR2_customir.cpp
+++ b/LR2/LR2_customir.cpp
@@ -269,9 +269,11 @@ bool CUSTOMIR_MANAGER::TryGetTargetInfo(game& g, CSTR songmd5, int mode, CSTR* o
 	case openlr2::GetStatus::Ok:
 		break;
 	case openlr2::GetStatus::Retry:
+		OverlayNotification("'%s' failed to get ghost data - ignoring retry\n", ir->Name().c_str());
 		return false;
 	case openlr2::GetStatus::Fail:
-		return false;
+		OverlayNotification("'%s' failed to get ghost data\n", ir->Name().c_str());
+	return false;
 	}
 
 	if (mode == 8) {
--- a/LR2/LR2_customir.cpp
+++ b/LR2/LR2_customir.cpp
@@ -284,10 +284,10 @@ bool CUSTOMIR_MANAGER::TryGetTargetInfo(game& g, CSTR songmd5, int mode, CSTR* o
 	*oName = result.displayName.c_str();
 	oData->fillzero();
 	if (!result.ghostData.empty()) oData->add(result.ghostData.c_str());
-	*oDigit1 = result.optionDigit1;
-	*oDigit2 = result.optionDigit2;
-	*oDigit3 = result.optionDigit3;
-	*oDigit4 = result.optionDigit4;
+	*oDigit1 = result.gaugeOption;
+	*oDigit2 = result.p1randomOption;
+	*oDigit3 = result.p2randomOption;
+	*oDigit4 = result.dpFlipOption;
 	*oSeed = result.randomSeed;
 	return true;
 }
--- a/LR2/LR2_customir.cpp
+++ b/LR2/LR2_customir.cpp
@@ -5,12 +5,14 @@
 #include "structure.h"
 
 #include <algorithm>
+#include <atomic>
 #include <array>
 #include <chrono>
 #include <cmath>
 #include <cstring>
 #include <filesystem>
 #include <future>
+#include <memory>
 #include <optional>
 #include <ranges>
 #include <string>
@@ -168,25 +170,36 @@ CUSTOMIR_MANAGER::~CUSTOMIR_MANAGER() {
 	}
 }
 
-static void SendScoreWithBlockingRetry(CustomIR& ir, const IRScoreV1& scoreV1) {
+static void SendScoreWithBlockingRetry(CustomIR& ir, const IRScoreV1& scoreV1, game* gamePtr, std::shared_ptr<std::atomic<int>> sendScorePending) {
 	constexpr int tryMax = 6;
 	int tryCount = 1;
-	while (tryCount <= tryMax) {
+	bool finished = false;
+	while (!finished && tryCount <= tryMax) {
 		switch (ir.SendScore(scoreV1)) {
 		case SendScoreStatus::Fail:
 			OverlayNotification("'%s' failed to submit score\n", ir.Name().c_str());
-			return;
+			finished = true;
+			break;
 		case SendScoreStatus::Ok:
-			return;
+			finished = true;
+			break;
 		case SendScoreStatus::Retry:
 			std::this_thread::sleep_for(std::chrono::seconds(static_cast<int>(std::pow(4, tryCount))));
 			tryCount++;
 			break;
 		}
 	}
-	OverlayNotification("'%s' failed to submit score after %d attempts\n", ir.Name().c_str(), tryCount);
+	if (!finished) {
+		OverlayNotification("'%s' failed to submit score after %d attempts\n", ir.Name().c_str(), tryCount);
+	}
+	if (gamePtr == nullptr || sendScorePending == nullptr) {
+		return;
+	}
+	if (sendScorePending->fetch_sub(1) == 1) {
+		gamePtr->gameplay.p1Score.InitJudgeQueue();
+	}
 }
-static void SendScoreMultiplexed(std::vector<std::future<void>>& mSendThreads, const IRScoreV1& scoreV1, std::vector<std::shared_ptr<CustomIR>> irs) {
+static void SendScoreMultiplexed(std::vector<std::future<void>>& mSendThreads, const IRScoreV1& scoreV1, std::vector<std::shared_ptr<CustomIR>> irs, game* gamePtr, std::shared_ptr<std::atomic<int>> sendScorePending) {
 	std::vector<int> finishedThreads;
 	for (const auto& [i, it] : std::views::enumerate(mSendThreads)) {
 		if (it.wait_for(std::chrono::milliseconds(0)) == std::future_status::ready) {
@@ -203,9 +216,13 @@ static void SendScoreMultiplexed(std::vector<std::future<void>>& mSendThreads, c
 	for (const auto& ir : irs) {
 		mSendThreads.push_back(std::async(
 					std::launch::async,
-					[](std::shared_ptr<CustomIR> ir, IRScoreV1 score){ SendScoreWithBlockingRetry(*ir, score); },
+					[](std::shared_ptr<CustomIR> ir, IRScoreV1 score, game* gamePtr, std::shared_ptr<std::atomic<int>> sendScorePending) {
+						SendScoreWithBlockingRetry(*ir, score, gamePtr, sendScorePending);
+					},
 					ir,
-					scoreV1));
+					scoreV1,
+					gamePtr,
+					sendScorePending));
 	}
 }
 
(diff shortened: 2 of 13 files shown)


LR2/LR2_customir.cpp
```diff
@@ -251,6 +257,39 @@ void CUSTOMIR_MANAGER::Login() {
...
+
+	if (mode == 8) {
+		*oName = "AVERAGE";
+		*oExscore = result.averageExscore;
+		return result.averageExscore > 0;
+	}
+	*oName = result.displayName.c_str();
+	oData->fillzero();
+	if (!result.ghostData.empty()) oData->add(result.ghostData.c_str());
+	*oDigit1 = result.optionDigit1;
```
MatVeiQaaa: please give those proper names, its called "digit" because its RE'ed code and gomazk likely wasn't sure what it meant at the time of naming the variable, or because it was a universal function which could parse various .xml. Right now, since you know exactly what data you got there, those should be named properly.

LR2/LR2_customir_api.h
```diff
@@ -115,6 +116,18 @@ enum class SendScoreStatus: int {
 	Fail,
 };
 
+struct IRGhostResult {
+	std::string displayName;
+	std::string ghostData;
+	int optionDigit1{};
```
MatVeiQaaa: same here, name them according to their function.

LR2/LR2_ghost.cpp
```diff
@@ -441,30 +441,6 @@ CSTR PLAYSCORE::EncodeGhostData(void) {
 		o++;
 	}
 
-	this->name.fillzero();
```
MatVeiQaaa, line 444: this must be made into a separate function belonging to `PLAYSCORE` namespace. Something like `ResetGhostData`. Then you call it after performing calls to IR methods. Note that original LR2IR score method must be refactored to receive ghostString as an argument, so it deals with own copy of it.

LR2/LR2_customir_api.h
luxurya111: Fixed it, Thank you.

ExampleIR/ExternalIR.cpp
```diff
@@ -43,15 +43,17 @@ static SendScoreStatus SendScore(const IRScoreV1& score) {
...
         score.song.hash, score.state.keymode, score.exscore,
         score.judgements_total.epg + score.judgements_total.lpg,
         score.judgements_total.egr + score.judgements_total.lgr,
         score.judgements_total.egd + score.judgements_total.lgd,
         score.judgements_total.ebd + score.judgements_total.lbd,
         score.judgements_total.epr + score.judgements_total.lpr,
         score.judgements_total.fast, score.judgements_total.slow, score.judgements_total.cb,
-        lamps[score.clearType]
+        lamps[score.clearType],
+        score.ghostData
```
MatVeiQaaa, line 56: not really necessary, this structure just serves as the basic example with whatever data. Keep it since you have already put it there, but in the future it's unnecessary to bother modifying it.

LR2/LR2_customir.cpp
```diff
@@ -251,6 +257,41 @@ void CUSTOMIR_MANAGER::Login() {
 	}
 }
 
+bool CUSTOMIR_MANAGER::TryGetTargetInfo(game& g, CSTR songmd5, int mode, CSTR* oData, CSTR* oName, int* oDigit1, int* oDigit2, int* oDigit3, int* oDigit4, int* oSeed, int* oExscore) const {
```
MatVeiQaaa: function proto could also use better arg names.

LR2/LR2_ghost.cpp
luxurya111: It appears that the InitJudgeQueue method performs almost the same function, so I have modified it to use this. There is a note in the internal TODO regarding the removal of this method, so if this causes any issues, I will create a separate dedicated method instead. For now, I will proceed with using this method.

As of right now, it only initializes immediately after the CustomIR transmission, and I will move forward with the LR2IR refactoring.

LR2/LR2_customir.cpp
```diff
@@ -168,25 +170,36 @@ CUSTOMIR_MANAGER::~CUSTOMIR_MANAGER() {
...
 	}
-	OverlayNotification("'%s' failed to submit score after %d attempts\n", ir.Name().c_str(), tryCount);
+	if (!finished) {
+		OverlayNotification("'%s' failed to submit score after %d attempts\n", ir.Name().c_str(), tryCount);
+	}
+	if (gamePtr == nullptr || sendScorePending == nullptr) {
+		return;
+	}
+	if (sendScorePending->fetch_sub(1) == 1) {
+		gamePtr->gameplay.p1Score.InitJudgeQueue();
```
MatVeiQaaa: an atomic flag (through a shared_ptr, couldn't it just be a reference?), and a bunch of code, all just to reset the ghost buffer. Which is only used once to get a string, which never changes between threads. Wouldn't it be a whole lot easier to just generate the string on main thread and pass it down to IR threads? And after the string is generated, reset the buffer. If it's uncomfortable because this string is used at different points of execution, it can be saved somewhere to game struct.

LR2/LR2_customir.cpp
luxurya111: Fixed it. Thank you.

LR2/LR2_customir.cpp
MatVeiQaaa: It's only used in score struct, right? All of the data the score struct gets is at gameplay struct, mostly gameplay::player. It sounds like it would make most sense to create a new field at player, like "resultGhost", and fill it at the end of the play before any IR operation. The buffer could be reset as soon as the ghost string is saved. Only concern then is some code to control the state of this new resultGhost field, would probably make sense to put it in InitGameplay, where player struct is reset.

MatVeiQaaa: i'm sorry for this pulling back and forth, but it took a while to realize what the appropriate solution was. I feel that this would be the cleanest way, and the globally available ghost string of last play could appear useful in the future again.

luxurya111: Initially, I thought a lot about how to approach this and think I made it more complicated than it needed to be. It seems there is no need to use it after MakeScoreV1 has been processed. Therefore, I went with a much simpler approach. Thank you for letting me know.

You can check the specific changes below:
https://github.com/GOMazk/OpenLR2/compare/55948659f0d60cdec2ca078e741d597bbfbde3f9..460a1bf9aae83d0e36cc5d2e2f7cf42e3b5316ae

luxurya111: I am sorry, I missed that message and went ahead with the work. If a rollback is still needed for the currently applied commit, I will roll it back to the previous state.

missed messages: 

> It's only used in score struct, right? All of the data the score struct gets is at gameplay struct, mostly gameplay::player. It sounds like it would make most sense to create a new field at player, like "resultGhost", and fill it at the end of the play before any IR operation. The buffer could be reset as soon as the ghost string is saved. Only concern then is some code to control the state of this new resultGhost field, would probably make sense to put it in InitGameplay, where player struct is reset.

> i'm sorry for this pulling back and forth, but it took a while to realize what the appropriate solution was. I feel that this would be the cleanest way, and the globally available ghost string of last play could appear useful in the future again.

MatVeiQaaa: > I am sorry, I missed that message and went ahead with the work. If a rollback is still needed for the currently applied commit, I will roll it back to the previous state.

Well im only sharing an idea, it doesn't have to be implemented this way. If what you did works for your purpose it's fine too, doesn't look bad in current shape.

LR2/LR2_ir.cpp
```diff
@@ -719,7 +719,7 @@ bool NETWORK::GetTargetInfo(game& g, int mode, CSTR songmd5, CSTR *oData, CSTR *
 	return false;
 }
 
-static void ThreadProc_IRsendScore(NETWORK *ir) {
+static void ThreadProc_IRsendScore(NETWORK *ir, CSTR ghostString) {
```
MatVeiQaaa: if you touch any code and have to use strings, don't use CSTR and do std::string instead. Then if you need to pass it to legacy code just do std::string::c_str()

LR2/LR2_ghost.cpp
luxurya111: I also proceeded with the refactoring of LR2IR, modifying it to receive and handle ghostString as a separate argument.

luxurya111:
PR review are ready

LR2/LR2_customir.cpp
```diff
@@ -619,6 +669,9 @@ void CUSTOMIR_MANAGER::BeginResultIr(game& game, sqlite3* sql, int player) {
 	IRScoreInternal internal{ game, sql, player };
 	IRScoreV1 scoreV1;
 	internal.MakeScoreV1(scoreV1);
+	if (player == 0) {
+		game.gameplay.p1Score.InitJudgeQueue();
```
MatVeiQaaa: hmm i still feel weird about this. It feels like the state of any game memory shouldn't be controlled from anywhere within CUSTOMIR_MANAGER, it has no business doing that. Should be done outside by main game code. More of a nitpick, i guess.

LR2/LR2_customir_api.h
```diff
@@ -115,6 +116,18 @@ enum class SendScoreStatus: int {
...
+struct IRGhostResult {
+	std::string displayName;
+	std::string ghostData;
+	int gaugeOption{};
+	int p1randomOption{};
+	int p2randomOption{};
+	int dpFlipOption{};
+	int randomSeed{};
+	int averageExscore{};
+	bool hasPlay{};
```
chown2:
```suggestion
```
Unused

LR2/LR2_customir_api.h
```diff
@@ -115,6 +116,18 @@ enum class SendScoreStatus: int {
...
 };
 
+struct IRGhostResult {
+	std::string displayName;
+	std::string ghostData;
+	int gaugeOption{};
+	int p1randomOption{};
+	int p2randomOption{};
+	int dpFlipOption{};
+	int randomSeed{};
```
chown2:
```suggestion
	std::string ghostData;
```
All of these are already in IRRankingPlay
This is what I mentioned here https://github.com/GOMazk/OpenLR2/pull/143#issuecomment-4758357570

chown2: Oh, it's because ghost can also be called without looking at leaderboards... That's not good....

LR2/LR2_customir.cpp
```diff
@@ -77,9 +77,10 @@ class CustomIR {
 	CustomIR(const std::filesystem::path& directory);
 	bool Initialize();
 	bool Login();
-	SendScoreStatus SendScore(const IRScoreV1& score);
+	SendScoreStatus SendScore(const IRScoreV1& score) const;
 	openlr2::GetStatus GetResultRank(const char* songHash, openlr2::IRRankResult& out);
 	openlr2::GetStatus RestoreCachedRank(const char* songHash, openlr2::IRRankResult& out);
+	openlr2::GetStatus GetGhost(const IRScoreV1& score, int mode, int targetPlayerId, IRGhostResult& out) const;
```
chown2:
```suggestion
	openlr2::GetStatus GetGhost(const char* songHash, GhostMode, int targetPlayerId, IRGhostResult& out) const;
```
According to TryGetTargetInfo below, the IRScoreV1 is just a dummy with only the hash
GhostMode should be new and defined above like 'enum class InputType'.
To convert an LR2 number to GhostMode, use `switch(mode) { case 0: return GhostMode::Whatver0GhostModeIs; // ...`.

LR2/LR2_customir_api.h
```diff
@@ -245,4 +259,6 @@ struct MethodTable {
 	void* reserved4 = nullptr;
 	void* reserved5 = nullptr;
 	void* reserved6 = nullptr;
+	// dream-pro getghost.cgi parity: mode 0=target, 6=top, 7=next, 8=average.
```
chown2:
```suggestion
	// This is called synchronously when play is entered to retrieve the ghost data for the play.
```

LR2/LR2_customir.cpp
chown2: And remove 'const' on the method, I will explain why in the message below

LR2/LR2_customir.cpp
```diff
@@ -77,9 +77,10 @@ class CustomIR {
 	CustomIR(const std::filesystem::path& directory);
 	bool Initialize();
 	bool Login();
-	SendScoreStatus SendScore(const IRScoreV1& score);
+	SendScoreStatus SendScore(const IRScoreV1& score) const;
```
chown2: This shouldn't gain 'const'. Technically, every of these methods can be 'const', since they call functions by pointers in mMethods. However, semantically the CustomIR implementation may choose to modify it's state, making the 'const' contradictory.
If this was needed because something uses CustomIR by const&, make it take a mutable reference instead.

LR2/LR2_customir.h
```diff
@@ -34,6 +35,7 @@ class CUSTOMIR_MANAGER {
 	void BeginResultIr(game& game, sqlite3* sql, int player);
 	void Initialize(const std::filesystem::path& directory, std::string activeProvider);
 	void Login();
+	bool TryGetTargetInfo(game& g, CSTR songmd5, int mode, CSTR* oData, CSTR* oName, int* oGaugeOption, int* oP1randomOption, int* oP2randomOption, int* oDpFlipOption, int* oRandomSeed, int* oExscore) const;
```
chown2: The interface LR2 used for that is hideous. Please just return a `std::optional<IRGhostResult>` instead, and use the fields you want outside.

chown2:
One more usage of `g->net.myRanking.ghost` remained in the CmdSearch function. Please, change it like you did in other places, and then modify structure.h struct MYRANKING:
```
CSTR ghost; /* struct to? */
```
->
```
CSTR _ghost; // TODO: remove when we throw out memset usage
```

LR2/LR2_customir.cpp
chown2: It should definitely not be here. It modifies external state, but this function should be read-only. This should be a part of whaterver cleans up after play.
I think this just eagerly removes the play buffers used for ghost.
In the original code, maybe this was like a safe guard that ghost is only calculated once? But does that matter?
Maybe we can just skip it and it will do the same on the next 'play' enterance? :)

LR2/LR2_customir.cpp
```diff
@@ -251,6 +257,48 @@ void CUSTOMIR_MANAGER::Login() {
...
+	const auto irIt = std::ranges::find(mModules, mDisplayIr, &CustomIR::Name);
+	if (irIt == mModules.end()) { return false; }
+
+	openlr2::GhostMode ghostMode = openlr2::GhostMode::Top; // default is "top"; as dream-pro default
+	switch (mode) {
+	case 0: ghostMode = openlr2::GhostMode::Target; break;
+	case 6: ghostMode = openlr2::GhostMode::Top; break;
+	case 7: ghostMode = openlr2::GhostMode::Next; break;
+	case 8: ghostMode = openlr2::GhostMode::Average; break;
+	}
```
chown2, line 273:
```suggestion
	switch (mode) {
	case 0: ghostMode = openlr2::GhostMode::Target; break;
	case 6: ghostMode = openlr2::GhostMode::Top; break;
	case 7: ghostMode = openlr2::GhostMode::Next; break;
	case 8: ghostMode = openlr2::GhostMode::Average; break;
	default:
			ErrorLogFmtAdd("Invalid ghost mode: %d\n", mode);
			return false;
	}
```

LR2/LR2_customir_api.h
```diff
@@ -134,6 +146,13 @@ enum class InputType : int {
 	Midi,
 };
 
+enum class GhostMode : int {
+	Target = 0,
+	Top = 6,
+	Next = 7,
+	Average = 8,
```
chown2:
```suggestion
	Top,
	Next,
	Average,
```
These values don't have to match LR2's since we convert them.
CustomIR API shouldn't be tied to LR2's internal data structures.

LR2/LR2_customir.cpp
luxurya111: `openlr2::GhostMode ghostMode = openlr2::GhostMode::Top; // default is "top"; as dream-pro default`

Should we avoid using the ghostMode default value? Or is it just for error handling?

luxurya111:
> One more usage of g->net.myRanking.ghost remained in the CmdSearch function. Please, change it like you did in other places, and then modify structure.h struct MYRANKING:

Fixed it. Thank you

LR2/LR2_customir.cpp
MatVeiQaaa: i still think that my proposed solution where we save ghost string once to some global state for any further use is the most elegant solution, which solves multiple problems.

LR2/LR2_customir.cpp
luxurya111: That's fine, I will remove the default value for now and make it return false if the mode is invalid.

LR2/LR2_customir.cpp
luxurya111: I have made the modifications. For now, instead of resetting in Custom IR, I added a process to initialize right before BeginResultIr when accessing Scene05 (Other than that, global variables are initialized at the required timing.), and then generate the current Ghost. I have temporarily added this as a global variable in the gameplay section of structure.h.

ExampleIR/ExternalIR.cpp
```diff
@@ -87,6 +89,23 @@ static openlr2::GetStatus GetResultRank(const char* songHash, int /*reserved*/,
     return openlr2::GetStatus::Ok;
 }
 
+static openlr2::GetStatus GetGhost(const char* songHash, openlr2::GhostMode mode, int /*targetPlayerId*/, IRGhostResult& out) {
+    std::println(std::cout, "GetGhost({})", songHash);
+    out = {};
+    if (mode == openlr2::GhostMode::Average) {
+        out.averageExscore = 800;
+        return openlr2::GetStatus::Ok;
+    }
```
chown2, line 98: No concrete proposal for now, just thinking out loud: since 'average' case has a completely different output, maybe IRGhostResult should be a tagged union of the usual ghost data and average exscore data?
It's a a bit complicated type to put into the API, but it seems fitting to me.

It would be something like this:
```cpp
struct AverageExscore {
	int value;
};
struct GhostData {
	// things that are in the current IRGhostResult
};
using IRGhostResult = std::variant<GhostData, AverageExscore>;
```

std::variant is usually used like this then:

```cpp
// in the beginning of the function
template <class... Ts> struct overloaded : Ts...
{
    using Ts::operator()...;
};

// ...

IRGhostResult result;

std::visit(overloaded{
  [](const AverageExscore& result) {/*handle...*/},
  [](const GhostData& result) {/*handle...*/},
}, result);
```

LR2/LR2_customir_api.h
```diff
@@ -115,6 +116,17 @@ enum class SendScoreStatus: int {
...
 };
 
+struct IRGhostResult {
+	std::string displayName;
+	std::string ghostData;
+	int gaugeOption{};
+	int p1randomOption{};
+	int p2randomOption{};
+	int dpFlipOption{};
+	int randomSeed{};
```
chown2:
```suggestion
	// P1 and P2 random layouts.
	// Should be 0 if the layout is not known, or the corresponding \ref randomOption is not noran, mirror, or random.
	// Examples: 1234567 54321 135792468.
	std::array<int, 2> randomLayout{};
	std::array<Random, 2> randomOption{};
	Gauge gauge{Gauge::Unknown};
	// LR2 rseed. If not known, the layout from \ref randomLayout can be used instead.
	// -1 if unknown, 0-0x7ffe otherwise.
	int rseed{-1};
	bool dpflip{};
	bool reserved1{};
	bool reserved2{};
	bool reserved3{};
```
To match IRRankPlayer.
Also adds randomLayout - not used for now, but so that it's clear from the API that it will work later.
Make dpflip a bool as it can only be yes or no (is this correct?).
NOTE: reserved{1-3} are there to not have any implicit padding in the struct. And we can use them later if we ever need to.

LR2/LR2_customir_api.h
```diff
@@ -115,6 +116,17 @@ enum class SendScoreStatus: int {
 	Fail,
 };
 
+struct IRGhostResult {
```
chown2: please move it under `namespace openlr2 {`

chown2 (commented):
Design-wise, my 'IRGhostResult = std::variant' idea is the only concern I have, so a few little changes and this should be ready!
I didn't verify how the new resultGhostForIr works but I'll look into it a bit later

LR2/LR2_ir.cpp
```diff
@@ -718,11 +718,11 @@ bool NETWORK::GetTargetInfo(game& g, int mode, CSTR songmd5, CSTR *oData, CSTR *
 		*oName = result->displayName.c_str();
 		oData->fillzero();
 		if (!result->ghostData.empty()) oData->add(result->ghostData.c_str());
-		*oDigit1 = result->gaugeOption;
-		*oDigit2 = result->p1randomOption;
-		*oDigit3 = result->p2randomOption;
-		*oDigit4 = result->dpFlipOption;
-		*oSeed = result->randomSeed;
+		*oDigit1 = result->gauge == openlr2::Gauge::Unknown ? 0 : static_cast<int>(result->gauge) - 1;
```
chown2: use a switch instead pls

LR2/LR2_ir.cpp
```diff
@@ -702,7 +707,32 @@ int OpenWebRanking(CSTR songmd5){
...
+			return true;
+		}
+		*oName = result->displayName.c_str();
+		oData->fillzero();
+		if (!result->ghostData.empty()) oData->add(result->ghostData.c_str());
+		*oDigit1 = result->gauge == openlr2::Gauge::Unknown ? 0 : static_cast<int>(result->gauge) - 1;
+		*oDigit2 = static_cast<int>(result->randomOption[0]);
+		*oDigit3 = static_cast<int>(result->randomOption[1]);
+		*oDigit4 = result->dpflip ? 1 : 0;
+		*oSeed = result->rseed >= 0 ? result->rseed : 0;
```
chown2: Leave a note to later fix already existing LR2 bug
```suggestion
		*oSeed = result->rseed >= 0 ? result->rseed : 0; //TOFIX: 0 is a valid seed
```

ExampleIR/ExternalIR.cpp
luxurya111: Understood. The general direction seems to be correct. Since it's a bit late right now, I will review it once more after getting some sleep.

If we don't implement it as something like Ghost2 and instead proceed directly, there might be a way.

chown2: I've given this some though and this would complicate the ABI greatly. Current approach is good enough.

LR2/LR2_customir_api.h
```diff
@@ -107,6 +107,7 @@ struct IRScoreV1 {
 		std::array<int, 1000> exscore{};
 		std::array<int, 1000> rate{};
 	} graphs{};
+	std::string ghostData{};
```
chown2:
```suggestion
	std::string ghostData{}; // In LR2 ghost format. May not be present in some cases, e.g. course scores.
```

LR2/LR2_customir_api.h
```diff
@@ -223,6 +250,7 @@ struct MethodTable {
 	// Ran on its own thread at score result, both for normal plays and courses.
 	// This is called even with scores that wouldn't be sent to LR2IR or saved to the score.db, it's up to the module to
 	// filter them.
+	// score.ghostData is populated when live ghost capture succeeded; upload or ignore per IR policy.
```
chown2:
```suggestion
```

LR2/LR2_customir.cpp
```diff
@@ -251,6 +257,41 @@ void CUSTOMIR_MANAGER::Login() {
...
+		return std::nullopt;
+	case openlr2::GetStatus::Fail:
+		OverlayNotification("'%s' failed to get ghost data\n", (*irIt)->Name().c_str());
+		return std::nullopt;
+	}
+
+	if (ghostMode == openlr2::GhostMode::Average && result.averageExscore <= 0) {
+		return std::nullopt;
+	}
+	return result;
```
chown2: Avoid redundant temporary variable songHash and better handle invalid GetGhost return values.
If GetGhost return value is outside of these three values it's actually undefined behavior, but it's better to be defensive about such things.
```suggestion
	openlr2::IRGhostResult result{};
	switch ((*irIt)->GetGhost(songmd5.body, ghostMode, targetPlayerId, result)) {
	case openlr2::GetStatus::Ok:
		if (ghostMode == openlr2::GhostMode::Average && result.averageExscore <= 0) {
			ErrorLogAdd("CustomIR BUG: invalid averageExscore\n");
			return std::nullopt;
		}
		return result;
	case openlr2::GetStatus::Retry:
		OverlayNotification("'%s' failed to get ghost data - ignoring retry\n", (*irIt)->Name().c_str());
		return std::nullopt;
	case openlr2::GetStatus::Fail:
		OverlayNotification("'%s' failed to get ghost data\n", (*irIt)->Name().c_str());
		return std::nullopt;
	}
	// TODO: remove 'abort()' from other handlers, logging an error and returning nothing instead. 
	ErrorLogAdd("CustomIR BUG: invalid GetGhost return value\n");
	return std::nullopt;
```

ExampleIR/ExternalIR.cpp
```diff
@@ -26,7 +26,7 @@ static bool Login() {
 }
 
 static SendScoreStatus SendScore(const IRScoreV1& score) {
-    std::println(std::cout, "SendScore({{.song.hash={}}})", score.song.hash);
+    std::println(std::cout, "SendScore({{.song.hash={}}}, ghostData={})", score.song.hash, score.ghostData);
```
chown2: I used this for quick glancing at what methods are being called, let's not pollute stdout
```suggestion
    std::println(std::cout, "SendScore({{.song.hash={}}})", score.song.hash);
```
(Don't get me wrong, this only applies to stdout, I'm not against adding this to score{}.txt)

LR2/LR2_customir.h
```diff
@@ -34,6 +35,7 @@ class CUSTOMIR_MANAGER {
 	void BeginResultIr(game& game, sqlite3* sql, int player);
 	void Initialize(const std::filesystem::path& directory, std::string activeProvider);
 	void Login();
+	std::optional<openlr2::IRGhostResult> TryGetTargetInfo(CSTR songmd5, int mode, int targetPlayerId) const;
```
chown2: 1) use const char* instead - all my homies hate CSTR (apply this on the .cpp side too)
2) add some notes to docs
```suggestion
	// \note Delegates to the display IR
	// \retval nullopt - Fail
	std::optional<openlr2::IRGhostResult> TryGetTargetInfo(const char* songmd5, int mode, int targetPlayerId) const;
```

LR2/structure.h
```diff
@@ -1415,16 +1418,18 @@ struct NETWORK {
...
 
 	int HTTPrequest();
+	void WaitForRankingHandle();
 	void WaitAndInitRanking();
 	int GetRanking(CSTR hash, char flagInit);
 
 	int GetRivalInfo(int ID_rival);
 
 	int GetTargetInfo(int mode, CSTR songmd5, CSTR * oStr1, CSTR * oStr2, int * oDigit1, int * oDigit2, int * oDigit3, int * oDigit4, int * oUnk, int * oExscore);
+	bool GetTargetInfo(game& g, int mode, CSTR songmd5, CSTR * oStr1, CSTR * oStr2, int * oDigit1, int * oDigit2, int * oDigit3, int * oDigit4, int * oUnk, int * oExscore);
```
chown2: 1) Rename old method so we don't call it by accident
2) Remove 'game&' from the new method. Only used as `g.net.rankingData.target_ID`? Since it's a member function, you can write `rankingData.target_ID`.
```suggestion
	int LR2IR_GetTargetInfo(int mode, CSTR songmd5, CSTR * oStr1, CSTR * oStr2, int * oDigit1, int * oDigit2, int * oDigit3, int * oDigit4, int * oUnk, int * oExscore);
	bool GetTargetInfo(int mode, CSTR songmd5, CSTR * oStr1, CSTR * oStr2, int * oDigit1, int * oDigit2, int * oDigit3, int * oDigit4, int * oUnk, int * oExscore);
```

ExampleIR/ExternalIR.cpp
```diff
@@ -87,6 +89,22 @@ static openlr2::GetStatus GetResultRank(const char* songHash, int /*reserved*/,
...
+    if (mode == openlr2::GhostMode::Average) {
+        out.averageExscore = 800;
+        return openlr2::GetStatus::Ok;
+    }
+    out.displayName = "EXAMPLE";
+    out.ghostData = "#EXAMPLE,1,0,LXZ,"; // this ghost will do almost perfect play.
+    out.gauge = openlr2::Gauge::Survival;
+    out.randomOption = { openlr2::Random::No, openlr2::Random::No };
+    out.dpflip = false;
+    out.rseed = 0;
```
chown2:
```suggestion
    out.rseed = 2179; // 1357246
```
Seed 0 is buggy in LR2, use something else in the example + an easter egg

Side-note: something seems broken in OpenLR2, even with LR2IR rseed is not applied

chown2: My bad, I forgot to turn on g-battle before testing :D
Both LR2IR and CustomIR work without problems

ExampleIR/ExternalIR.cpp
```diff
@@ -87,6 +89,22 @@ static openlr2::GetStatus GetResultRank(const char* songHash, int /*reserved*/,
...
+    std::println(std::cout, "GetGhost({})", songHash);
+    out = {};
+    if (mode == openlr2::GhostMode::Average) {
+        out.averageExscore = 800;
+        return openlr2::GetStatus::Ok;
+    }
+    out.displayName = "EXAMPLE";
+    out.ghostData = "#EXAMPLE,1,0,LXZ,"; // this ghost will do almost perfect play.
+    out.gauge = openlr2::Gauge::Survival;
+    out.randomOption = { openlr2::Random::No, openlr2::Random::No };
```
chown2:
```suggestion
    out.randomOption = { openlr2::Random::Random, openlr2::Random::No };
```
Make the easter egg actually visible :D

LR2/LR2_customir_api.h
```diff
@@ -245,4 +272,6 @@ struct MethodTable {
 	void* reserved4 = nullptr;
 	void* reserved5 = nullptr;
 	void* reserved6 = nullptr;
+	// This is called synchronously when play is entered to retrieve the ghost data for the play.
+	openlr2::GetStatus(OLR2_IR_API* GetGhost)(const char* songHash, openlr2::GhostMode mode, int targetPlayerId, openlr2::IRGhostResult& out) = nullptr;
```
chown2: Move this before all the 'reserved' fields please

chown2: And replace the comment before reserved1 to this to make their purpose clear (github won't let me leave a suggestion there):
```
	// Forward compatibility.
	// These fields will always be moved to the end of the struct when new fields are added.
	// Thanks to this CustomIR module designed for newer game version won't write out-of-bounds memory when assigning fields of older MethodTable.
```

chown2: Also, feel free to remove `Experimental API, may be changed` notes from other methods and structs. Thinking about it again, we would just be stepping on each other's toes if we actually change any of the already added APIs

LR2/LR2_customir_api.h
```diff
@@ -152,6 +160,25 @@ enum class Random : int {
...
+	std::array<Random, 2> randomOption{};
+	Gauge gauge{Gauge::Unknown};
+	// LR2 rseed. If not known, the layout from \ref randomLayout can be used instead.
+	// -1 if unknown, 0-0x7ffe otherwise.
+	int rseed{-1};
+	bool dpflip{};
+	bool reserved1{};
+	bool reserved2{};
+	bool reserved3{};
+	int averageExscore{};
```
chown2:
```suggestion
	int averageExscore{}; // Only used if \ref GhostMode == \ref GhostMode::Average.
```

ExampleIR/ExternalIR.cpp
MatVeiQaaa: > Make the easter egg actually visible :D

good ole turning on unrandomizer but forgetting to turn on random

luxurya111: Oh my goodness, why didn't I change this..? I have fixed the Ghost along with it. sorry :)

LR2/structure.h
luxurya111: You are right, there was really no reason to take this approach. It has been fixed.

LR2/LR2_customir_api.h
luxurya111: Removed the Experimental API comments

luxurya111:
I have applied the feedback. Thank you.

ExampleIR/ExternalIR.cpp
```diff
@@ -87,6 +89,22 @@ static openlr2::GetStatus GetResultRank(const char* songHash, int /*reserved*/,
...
 
+static openlr2::GetStatus GetGhost(const char* songHash, openlr2::GhostMode mode, int /*targetPlayerId*/, openlr2::IRGhostResult& out) {
+    std::println(std::cout, "GetGhost({})", songHash);
+    out = {};
+    if (mode == openlr2::GhostMode::Average) {
+        out.averageExscore = 800;
+        return openlr2::GetStatus::Ok;
+    }
+    out.displayName = "EXAMPLE";
+    out.ghostData = "#EXAMPLE,21,2179,LXZ,"; // this ghost will do almost perfect play.
```
chown2: The point is exactly going away from the LR2IR's interface.
This seems to work. If you follow the usage of 'ghostData', it directly reaches DecodeGhostData. Please recheck.
```suggestion
    out.ghostData = "LXZ"; // this ghost will do almost perfect play.
```

LR2/LR2_ir.cpp
```diff
@@ -702,7 +707,56 @@ int OpenWebRanking(CSTR songmd5){
...
+		case openlr2::Random::Scatter: *oDigit3 = 4; break;
+		case openlr2::Random::Converge: *oDigit3 = 5; break;
+		default: *oDigit3 = 0; break;
+		}
+		*oDigit4 = result->dpflip ? 1 : 0;
+		*oSeed = result->rseed >= 0 ? result->rseed : 0; //TOFIX: 0 is a valid seed
+		return true;
+	}
+	if (isOnline) {
+		WaitAndInitRanking();
```
chown2: Very minor, but avoids calling WaitForRankingHandle twice.
```suggestion
		this->rankingData.Init();
```

LR2/LR2_customir_api.h
```diff
@@ -152,7 +160,25 @@ enum class Random : int {
 	Converge, // AKA All-Scratch
 };
 
-// \warning Experimental API, may be changed.
+struct IRGhostResult {
+	std::string displayName;
+	std::string ghostData;
```
chown2: Picked a random short example from the LR2IR archive `#~,4,13233,E@3ZZ,`.
To make it clear it's the ghost data itself, not what was getghost.cgi output.
```suggestion
	std::string ghostData; // Example: E@3ZZ
```

chown2 (approved):
Please double check that ghostData works when used like I suggested. Then LGTM!  🐖 🚀 

Please merge with squashing

LR2/LR2_customir_api.h
luxurya111: That's right. When I wrote the example, I was probably using ghostData itself as a template and overlooked that it was being used as EncodeGhostData.

luxurya111:
> Please double check that ghostData works when used like I suggested. Then LGTM! 🐖 🚀

I've verified the ghost random in the example as well. Thank you!

PR #148, opened 2026-06-22
build(msbuild): reorganize folder and project structure

MatVeiQaaa:
- `lib/` -> `dep/`. Now `lib/` stores .lib files. @chown2 pls confirm cmake still works, i edited it accordingly but didn't test.
- each dependency has its own project to avoid polluting the search in visual studio and provide ability to exert more control over their build process.
- removed some (seemingly) unnecessary files (.txt notes), which make sense for beta3 branch, but not here.
- output directory changed to `out/`
- build cache directory changed to `build/`
- small fixes in OpenLR2 proj file, like debug edit and continue for x86, among what i can remember.
- generally standardized and tidied the project.
- remove "_vs22" appendage in main .sln and .vcxproj files.


chown2 (commented):
I've moved DxLib into a subfolder in #132 so that there's a way for DxLib and dxlib-for-linux to be distinct.
Linux build doesn't work with this PR because dep/DxLib/*.h files get preferred now.

chown2 (commented):
Linking with Fmod fails with CMake on Windows.
Please apply this diff (GitHub doesn't let me add a suggestion).
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index e967619..254c851 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,18 +23,37 @@ add_library(
 target_include_directories(TinyXML PUBLIC dep/tinyxml)

 if(WIN32)
-    add_library(FMOD INTERFACE IMPORTED GLOBAL)
+    add_library(FMOD SHARED IMPORTED GLOBAL)
     target_include_directories(FMOD INTERFACE dep)
-    target_link_directories(FMOD INTERFACE dep/FMOD)
-    # TODO(no-vs-solutions): adjust paths with CMAKE_SIZEOF_VOID_P and use instead of target_link_directories
-    # set_target_properties(
-    #     FMOD
-    #     PROPERTIES
-    #         IMPORTED_LOCATION_DEBUG ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin64/fmodL.dll
-    #         IMPORTED_IMPLIB_DEBUG ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin64/fmodL_vc.lib
-    #         IMPORTED_LOCATION ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin64/fmod.dll
-    #         IMPORTED_IMPLIB ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin64/fmod_vc.lib
-    # )
+    if(CMAKE_SIZEOF_VOID_P EQUAL 8)
+        set_target_properties(
+            FMOD
+            PROPERTIES
+                IMPORTED_LOCATION_DEBUG
+                    ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin64/fmodL.dll
+                IMPORTED_IMPLIB_DEBUG
+                    ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin64/fmodL_vc.lib
+                IMPORTED_LOCATION
+                    ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin64/fmod.dll
+                IMPORTED_IMPLIB
+                    ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin64/fmod_vc.lib
+        )
+    elseif(CMAKE_SIZEOF_VOID_P EQUAL 4)
+        set_target_properties(
+            FMOD
+            PROPERTIES
+                IMPORTED_LOCATION_DEBUG
+                    ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin86/fmodL.dll
+                IMPORTED_IMPLIB_DEBUG
+                    ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin86/fmodL_vc.lib
+                IMPORTED_LOCATION
+                    ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin86/fmod.dll
+                IMPORTED_IMPLIB
+                    ${CMAKE_CURRENT_LIST_DIR}/dep/FMOD/bin86/fmod_vc.lib
+        )
+    else()
+        message(FATAL_ERROR "Unsupported CMAKE_SIZEOF_VOID_P")
+    endif()
 elseif(LINUX)
     add_library(FMOD SHARED IMPORTED GLOBAL)
     target_include_directories(FMOD INTERFACE dep)
```

MatVeiQaaa:
> I've moved DxLib into a subfolder in #132 so that there's a way for DxLib and dxlib-for-linux to be distinct. Linux build doesn't work with this PR because dep/DxLib/*.h files get preferred now.

i dislike this solution, it makes it so that dxlib requires special treatment in the project include paths. Trying to think of another way...

MatVeiQaaa:
@chown2 how about having both windows and linux implementations under a common directory DxLib, which holds a pass-through header with macros to fork by system, into appropriate subdirectory?

chown2:
I don't think indirection and lots of boilerplate macros are any better than having files in another folder.
And the whole 'special treatment' is just having a single include path argument having DxLib/DxLib instead of plain DxLib.

MatVeiQaaa:
> I don't think indirection and lots of boilerplate macros are any better than having files in another folder. And the whole 'special treatment' is just having a single include path argument having DxLib/DxLib instead of plain DxLib.

right now the only include path in msbuild is `dep/`. It doesn't make any special distinction for any one dependency it contains. When i looked it seemed to be much the same in cmake too.

MatVeiQaaa:
> I don't think indirection and lots of boilerplate macros are any better than having files in another folder. And the whole 'special treatment' is just having a single include path argument having DxLib/DxLib instead of plain DxLib.

difference is where the solution is delegated to. I propose to make the dependency responsible for system forking, instead of the master project.

chown2:
md5 was used as <md5.h> before the changes in this PR (En_fileutil.cpp‎). Fmod is included as `FMOD/fmod.h` but that is actually incorrect as you are supposed to include just `fmod.h` instead. Same with `<sqlite/sqlite3.h>` that should be `<sqlite3.h>`. In fact, it's `dep/` that should not be in the include paths.

MatVeiQaaa:
> md5 was used as <md5.h> before the changes in this PR (En_fileutil.cpp‎). Fmod is included as `FMOD/fmod.h` but that is actually incorrect as you are supposed to include just `fmod.h` instead. Same with `<sqlite3.h>`. In fact, it's `dep/` that should not be in the include paths.

i really dislike this. If that's the case, i wonder how it managed to build in msbuild, also. Then every single dependency you add requires altering include paths, if we go by this standard.

chown2:
Before your changes

`<AdditionalIncludeDirectories>$(ProjectDir)lib\FMOD;$(ProjectDir)lib\DxLib;$(ProjectDir)lib\md5;$(ProjectDir)lib\;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>`

> i really dislike this

Well, the way headers are included right now goes against how the developers of these libraries intended to.

> Then every single dependency you add requires altering include paths, if we go by this standard.

Yes, and that's exactly how it's usually done

chown2 (approved):

chown2:
DxLib building from source incoming  :kazakhstan:  :rocket::rocket::rocket:

MatVeiQaaa:
@GOMazk if everything is to your liking, this can be merged.

GOMazk (approved):

PR #149, opened 2026-06-22
[Bugfix] Add EXTENDEDPLAYERSTATS.cb by 1 on BAD

Unengine:
LR2 2026-06-23 00-12-46

`SetObjectValue_Num` in skinobject returns combo break stat (`EXTENDEDPLAYERSTATS.cb`) when case number is 216,
but I found `EXTENDEDPLAYERSTATS.cb` is only added by 1 on lpr(Late-poor, missing notes).
So I also made it add `EXTENDEDPLAYERSTATS.cb` by 1 also on BAD.

### DIFF
--- a/LR2/Scene04_Play.cpp
+++ b/LR2/Scene04_Play.cpp
@@ -845,6 +845,7 @@ int ProcSinglenote(game *g, int lane, int keypress, int timing, int player) {
 			auto increment_extended = [](EXTENDEDPLAYERSTATS& stats, bool isFast, int offset) {
 				isFast ? stats.ebd++ : stats.lbd++;
 				isFast ? stats.fast++ : stats.slow++;
+				stats.cb++;
 				stats.lastFastSlow = isFast ? 1 : 2;
 				stats.noteCount++;
 				stats.lastHitOffset = offset;


chown2 (approved):
Didn't test in real life, but makes sense in code
If you look right above it, it calls JudgeToScore with judge=2, which inside itself has logic to reset combo on it
https://github.com/GOMazk/OpenLR2/blob/1cd8cab5b2283475f36ced3ccbc2042c691c24d1/LR2/Scene04_Play.cpp#L80-L82
CC @MatVeiQaaa

MatVeiQaaa:
> Didn't test in real life, but makes sense in code If you look right above it, it calls JudgeToScore with judge=2, which inside itself has logic to reset combo on it
> 
> https://github.com/GOMazk/OpenLR2/blob/1cd8cab5b2283475f36ced3ccbc2042c691c24d1/LR2/Scene04_Play.cpp#L80-L82
> 
> CC @MatVeiQaaa

this is just how LR2 originally did it, unrelated to extended stats. It really is just missing a cb inc for bads.

MatVeiQaaa (approved):

Unengine:
Then, I think it's ok to merge before the next release.
cc @GOMazk

PR #156, opened 2026-06-24
feat: exclusive fullscreen (screenmode=2) with native refresh rate

Skalim112:
## Summary

Adds **exclusive fullscreen** as a third screen mode (`screenmode=2` on config.xml). 

In windowed/borderless, the Windows DWM compositor adds ~1–2 frames of fixed input-to-photon latency regardless of render FPS. Exclusive fullscreen bypasses the DWM via `SetFullScreenResolutionMode(DX_FSRESOLUTIONMODE_DESKTOP)`, noticeably improving responsiveness during gameplay.

The mode is selectable through the existing config.xml (under the option `screenmode`), the in-game **Screen Mode** option in song select, and the **F4** toggle (which now cycles `0 → 1 → 2`: borderless → windowed → exclusive).

Also, it adds compatibility with the screen's refresh rate, so it's now compatible with modern monitors (on the original, the refresh rate of the game was forced to 60hz).

## What changed

- **`Main.cpp`** — New `ApplyScreenMode(int)` helper, maps the screen mode to the correct DxLib calls: `0` borderless, `1` windowed, `2` exclusive (`SetFullScreenResolutionMode(DESKTOP)` + `ChangeWindowMode(0)`). All entry points (startup, F4, in-game option, IR ranking) now go through it instead of bare `ChangeWindowMode`, so exclusive is entered/restored consistently.

- **Refresh rate** — The hardcoded `60` in all `SetGraphMode` calls was replaced with `GetRefreshRate()`, so exclusive fullscreen runs at the monitor's native refresh rate (144/240/etc.) instead of being capped to 60 Hz. No effect in windowed/borderless (the DWM ignores it).

- **Screen Mode option (song select)** — The option now cycles through all three modes (`ButtonByInput` range widened `0, 1` → `0, 2`), shows the correct label (`option_str[12]` count `2 → 3`, with an `"EXCLUSIVE"` fallback injected at startup so the label isn't blank without editing the optionstr CSV).

- **Bug fixes around the exclusive transition:**
  - *Exclusive lost on certain transitions* — entry points re-apply the chosen mode via `ApplyScreenMode` instead of overwriting `screenmode` with `GetWindowModeFlag()`.
  - *IR ranking (F5)* — exclusive now drops to windowed before opening the external browser, instead of opening it while the game owns the display. A defensive measure once the F5 feature gets working in a future

## Why it's safe

- Default `screenmode` is unchanged (borderless), so existing users are unaffected.
- Tried to crash it on numerous situations where the original LR2 usually crashes, no problems on three days of intense testing.

## Testing

- Exclusive fullscreen entered via config, in-game option, and F4 — confirmed lower input latency vs borderless.
- Native refresh rate applied in exclusive (no 60 Hz cap).
- Screen Mode option cycles all three modes with correct labels; no double-toggle; cursor stays responsive after leaving exclusive.
- Alt-tab in/out of 7k gameplay (vsync 0 and 1) and rapid F4 cycling — no crash.
- IR ranking (F5): logic verified; could not be tested live (IR server down).

## Yes, this was deleted, why now?

Exclusive fullscreen brings back something none of the current iterations offer: **truly responsive, low-latency input.**

Rhythm-game simulators inherit input latency almost by default, whatever language or engine they use (Python, C#, Java, etc.). On Windows, any windowed or borderless app goes through the DWM compositor, which adds a fixed 1–2 frames of input-to-photon delay — OS-level, affecting every engine equally — while runtime overhead (GC, interpretation, layered input paths) piles on more. The original LR2 sidestepped the compositor entirely by running in exclusive fullscreen, which is a defining part of why it still feels sharper than its successors.

OpenLR2 dropped exclusive fullscreen for stability reasons (**EDIT**, it was the default behaviour of DxLib 3.24, thanks for the correction), but that tradeoff gave up the very thing that set LR2 apart. Now that the client has stronger crash-handling in place, we can reintroduce it safely — and do it *better* than the original: as an opt-in mode (default stays borderless), running at the monitor's native refresh rate, with the entry/exit edge cases properly handled.

In short: this restores LR2's signature input feel for players who want it, without forcing the risk on anyone who doesn't.
LR2 2026-06-24 01-21-02

## Fast/Slow difference

Borderless Fullscreen
LR2_2026-06-18_23-49-42

Exclusive Fullscreen
LR2_2026-06-18_23-51-34


GOMazk:
I have 3 monitors, and exclusive mode doesn't work properly. looks like it fails to find appropriate monitor.
+ if I turn off one of my monitor (= 2 monitors), It works well but frameLimiter(vSync) works as 60hz.

MatVeiQaaa:
change desktop to 0, borderless to 2, rename last option to "BORDERLESS". This makes a lot more sense.
> OpenLR2 dropped exclusive fullscreen for stability reasons, but that tradeoff gave up the very thing that set LR2 apart.

Also this isn't true. Borderless is just default in DxLib 3.24.

Skalim112:
> change desktop to 0, borderless to 2, rename last option to "BORDERLESS". This makes a lot more sense.
> 
> > OpenLR2 dropped exclusive fullscreen for stability reasons, but that tradeoff gave up the very thing that set LR2 apart.
> 
> Also this isn't true. Borderless is just default in DxLib 3.24.

my bad, i forgot how was made on first instance lmao.

I have two monitor and yea, i tested it with one. I will check this now.

Skalim112:
> I have 3 monitors, and exclusive mode doesn't work properly. looks like it fails to find appropriate monitor.
> 
>     * if I turn off one of my monitor (= 2 monitors), It works well but frameLimiter(vSync) works as 60hz.

Changes applied and tested with two monitors. I tested multiple scenarios where vsync could go crazy and its behaving ok for the moment.

I recommend further testings since i could miss something.

Also applied 0 for exclusive (the new FULL), 1 for WINDOWED and 2 for BORDERLESS

LR2/Main.cpp
```diff
@@ -81,6 +81,78 @@ static bool run_tests() {
...
+	}
+
+	return -1;
+}
+
+static int GetDisplayRefreshRate(int displayIndex) {
+	if (displayIndex < 0) return 0;
+
+	int dx = 0, dy = 0, dw = 0, dh = 0, primary = 0, refreshRate = 0;
+	if (GetDisplayInfo(displayIndex, &dx, &dy, &dw, &dh, &primary, &refreshRate) != 0) return 0;
```
chown2: Does this work?
```suggestion
	if (displayIndex < 0) return 0;
	int refreshRate = 0;
	if (GetDisplayInfo(displayIndex, nullptr, nullptr, nullptr, nullptr, nullptr, &refreshRate) != 0) return 0;
```

LR2/Main.cpp
```diff
@@ -81,6 +81,78 @@ static bool run_tests() {
 	return true;
 }
 
+#ifdef _WIN32
+static int g_exclusiveDisplayIndex = -1;
```
chown2, line 85: Is (g->config).system.maindisplay by any chance the same thing?

Skalim112: No, they represent different things. `maindisplay` is the configured/startup display, but the fix needs the display where the window currently is when switching to exclusive. `g_exclusiveDisplayIndex` caches that runtime-selected display so VSync/frame limiter uses the correct monitor refresh rate and remembers the display where the screen mode is executed.

In short, modifying `maindisplay` implies giving a full repurpose of that solution, which is riskier.

LR2/Main.cpp
Skalim112: Applied and tested, no problems detected

Skalim112:
- Fixed a bug where switching from `Windowed` to `Borderless` could keep using the previous monitor as the display target.
- Improved mouse position handling in `Borderless` mode so cursor detection follows the actual game window instead of the original/default monitor.

EDIT: i am doing some graphical testings right now for check if this does not affect the gameplay in general
EDIT 2: Mouse correction is now out of the scope of gameplay

MatVeiQaaa:
> EDIT: i am doing some graphical testings right now for check if this does not affect the gameplay in general EDIT 2: Mouse correction is now out of the scope of gameplay

if you mean fullscreen mode in general, it doesn't. I and some others have been playing fullscreen for around two months now using [this](https://github.com/MatVeiQaaa/UpscaLR2). It's largely everything the same that OpenLR2 has down the line.

Skalim112:
> > EDIT: i am doing some graphical testings right now for check if this does not affect the gameplay in general EDIT 2: Mouse correction is now out of the scope of gameplay
> 
> if you mean fullscreen mode in general, it doesn't. I and some others have been playing fullscreen for around two months now using [this](https://github.com/MatVeiQaaa/UpscaLR2). It's largely everything the same that OpenLR2 has down the line.

Ooohh i see i see, so thats why i can't make openLR2 crash on ex fullscreen, right?

Also, my testings were more oriented on detecting visual artifacts and micro stuttering in case i messed up something on the process

LR2/Main.cpp
```diff
@@ -81,6 +81,81 @@ static bool run_tests() {
...
+		ApplyCurrentWindowDisplayIndex();
+		SetFullScreenResolutionMode(DX_FSRESOLUTIONMODE_BORDERLESS_WINDOW);
+		ChangeWindowMode(0);
+		break;
+	}
+}
+#endif // _WIN32
+
+static double GetFrameLimiterRefreshRate() {
+#ifdef _WIN32
```
MatVeiQaaa, line 186: why win32? This isnt platform-specific code, as its abstracted behind DxLib.

MatVeiQaaa:
if i looked carefully enough you got absolutely no platform-specific code, all of the ``#if _WIN32`` seem redundant, as there is only code utilizing DxLib, which abstracts away platform code.

LR2/En_input.cpp
```diff
@@ -27,12 +27,45 @@ typedef struct MIDI {
...
+	POINT cursorPoint{};
+	if (GetCursorPos(&cursorPoint) == 0) return;
+	if (ScreenToClient(hWnd, &cursorPoint) == 0) return;
+
+	*mouseX = ScaleClientPointToSkin(cursorPoint.x, clientWidth, skinSizeX);
+	*mouseY = ScaleClientPointToSkin(cursorPoint.y, clientHeight, skinSizeY);
+}
+#endif // _WIN32
+
+void SetClientMousePositionFixFlag(bool enabled) {
```
MatVeiQaaa: why do we need this toggle at all? Whats the point in skipping it for play scene?

LR2/En_input.cpp
```diff
@@ -27,12 +27,45 @@ typedef struct MIDI {
...
 static MIDI midi;
+static bool g_clientMousePositionFix = false;
 
 #ifndef _WIN32
 #define LOWORD(l) ((WORD)(((DWORD_PTR)(l)) & 0xffff))
 #define HIWORD(l) ((WORD)((((DWORD_PTR)(l)) >> 16) & 0xffff))
 #endif
 
+#ifdef _WIN32
+static int ScaleClientPointToSkin(int point, int clientSize, int skinSize) {
```
MatVeiQaaa: make this a lambda in the scope where its used

LR2/En_input.cpp
chown2: I agree
1. If we do want this as a toggle, it should be passed through function arguments instead of a global variable
2. If we don't want to handle mouse input in the play skin, why is only this function being disabled, instead of avoiding all of the mouse input handling?
3. I'm pretty sure we actually *do* want to handle mouse input in play, unless that only worked in Lunatic Vibes. A skin of BKVad1m had ability to toggle sound effects mid-play.

LR2/Main.cpp
chown2: dxlib-for-linux probably doesn't have these methods :(
although ultimately I think that there should be no such ifdefs, instead there should be stubs for linux in `En_dxlibstub.h`, I just never got around to adding all of the unimplemented methods there
In fact, contributors shouldn't even have to spend time thinking about Linux. It's only me using it for development, and I can always fix it up myself.
For now, I think it's fine to proceed with ifdefs, I will later clean all of them up myself in one go

LR2/Main.cpp
```diff
@@ -81,6 +81,81 @@ static bool run_tests() {
...
+	GetWindowPosition(&wx, &wy);
+	GetWindowSize(&ww, &wh);
+
+	const int targetX = wx + ww / 2;
+	const int targetY = wy + wh / 2;
+	const int displayCount = GetDisplayNum();
+	for (int i = 0; i < displayCount; i++) {
+		int dx = 0, dy = 0, dw = 0, dh = 0, primary = 0;
+		if (GetDisplayInfo(i, &dx, &dy, &dw, &dh, &primary) != 0) continue;
+		if (targetX >= dx && targetX < dx + dw && targetY >= dy && targetY < dy + dh) return i;
```
chown2: Isn't this just comparing random variables? How does this help identify a monitor? Please add a comment before the function describing what the function tries to achieve

LR2/Main.cpp
```diff
@@ -81,6 +81,81 @@ static bool run_tests() {
...
+}
+
+static int GetDisplayRefreshRate(int displayIndex) {
+	if (displayIndex < 0) return 0;
+	int refreshRate = 0;
+	if (GetDisplayInfo(displayIndex, nullptr, nullptr, nullptr, nullptr, nullptr, &refreshRate) != 0) return 0;
+	return refreshRate;
+}
+
+static int ApplyCurrentWindowDisplayIndex() {
```
chown2: This function is 1.5 lines of code long and used in only two places, please just inline it

LR2/Main.cpp
```diff
@@ -2260,7 +2342,7 @@ int main(int argc, char** argv) {
 			SetObjectStrings_SongSelect(&gs);
 		}
 #endif // _WIN32
-		SetMouseDispFlag( (gs.KeyInput.mouse_oldX < skinSizeX && gs.KeyInput.mouse_oldY < skinSizeY) ? 0:1  ); //TODO_RESOULUTION
+		SetMouseDispFlag((gs.KeyInput.mouse_oldX >= 0 && gs.KeyInput.mouse_oldX < skinSizeX && gs.KeyInput.mouse_oldY >= 0 && gs.KeyInput.mouse_oldY < skinSizeY) ? 0 : 1); //TODO_RESOULUTION
```
chown2, line 2392: When would cursor position be below 0? And does it even matter here?

LR2/Main.cpp
chown2: If I read your comment correctly, you mean being able to switch between monitors by toggling out of fullscreen - moving the window to the other monitor - toggling back into fullscreen.
If possible, I think it would be better to always display the game in fullscreen on 'maindisplay', unless it's unavailable.
You don't have to act on this if it's gonna complicate the changes, I'm fine with using the current monitor too.

LR2/Main.cpp
chown2: Also, FixMousePointWithClientPosition actually does use plenty of WinAPI. GetClientRect GetCursorPos ScreenToClient.

LR2/Main.cpp
Skalim112: Well, in reality, this problem was present even on the Main branch but it was not visible because the exclusive fullscreen was not a thing. 

But yea, if i have to define this on human: `maindisplay` records the monitor where the game starts just one time, which means if i maximize the game on another monitor, it will return to the recorded one (the main display). `g_exclusiveDisplayIndex` is the monitor index cached at runtime (this index is determined by the number of monitors available and counted by `GetDisplayNum()`.

The logic of maintaining the game on the monitor i want is handled by the entire logic of `ApplyCurrentWindowDisplayIndex()`, and `g_exclusiveDisplayIndex` is exclusively used and recorded for use the hzs of the monitor where the game is right now, and, since it works in runtime, if you move the screen on windowed with VSYNC and on gameplay, you will see how the fps changes at real time.

LR2/En_input.cpp
Skalim112: Passed as an argument. Where can i get that skin btw? it is compatible with LR2?

LR2/En_input.cpp
Skalim112: Converted to a lambda function

LR2/Main.cpp
Skalim112: Commented solution

LR2/Main.cpp
Skalim112: This can happen on multi-monitor setups when the cursor leaves the game window but the game still has focus.

For example, if the game window is on the monitor to the right, moving the cursor to the monitor on the left makes the client-space X coordinate negative. Without the lower-bound check, values like `-10 < skinSizeX` still evaluate as inside the game area.

The opposite layout can produce large positive coordinates instead, depending on where the game window is relative to the virtual desktop.
qwe

GOMazk:
image

this is my setting and still not working. I think it's dxlib bug as so far. should merge this and split it as a new issue?

LR2/En_input.cpp
MatVeiQaaa: > Passed as an argument. Where can i get that skin btw? it is compatible with LR2?

It doesn't really matter, point is, no skins got a cursor in the play scene not because it's not possible, but just no one used it. Even then, you shouldn't need to skip converting the coords anyway, don't micro optimise.

LR2/Main.cpp
```diff
@@ -81,6 +81,86 @@ static bool run_tests() {
...
+
+	return -1;
+}
+
+static int GetDisplayRefreshRate(int displayIndex) {
+	if (displayIndex < 0) return 0;
+	int refreshRate = 0;
+	if (GetDisplayInfo(displayIndex, nullptr, nullptr, nullptr, nullptr, nullptr, &refreshRate) != 0) return 0;
+	return refreshRate;
+}
```
chown2: This helper is also redundant

LR2/Main.cpp
```diff
@@ -81,6 +81,86 @@ static bool run_tests() {
...
+	for (int i = 0; i < displayCount; i++) {
+		// dx/dy are the monitor position. dw/dh are the monitor size.
+		int dx = 0, dy = 0, dw = 0, dh = 0, primary = 0;
+		// Get each monitor rectangle and check if the game window center is inside it.
+		if (GetDisplayInfo(i, &dx, &dy, &dw, &dh, &primary) != 0) continue;
+		if (targetX >= dx && targetX < dx + dw && targetY >= dy && targetY < dy + dh) return i;
+	}
+
+	return -1;
+}
```
chown2: Thanks, it makes much more sense now! To make this more concise, let's rephrase things a bit.
```suggestion
static int FindMonitorThatContainsWindowCenter() {
	// All of these functions return coordinates on the global plane, where all monitors are laid out consecutively.
	int wx = 0, wy = 0;
	GetWindowPosition(&wx, &wy);
	int ww = 0, wh = 0;
	GetWindowSize(&ww, &wh);

	const int centerX = wx + ww / 2;
	const int centerY = wy + wh / 2;
	const int displayCount = GetDisplayNum();
	for (int i = 0; i < displayCount; i++) {
		int dx = 0, dy = 0, dw = 0, dh = 0, primary = 0;
		if (GetDisplayInfo(i, &dx, &dy, &dw, &dh, &primary) != 0) continue;
		if (centerX >= dx && centerX < dx + dw && centerY >= dy && centerY < dy + dh) return i;
	}

	return -1;
}
```

chown2:
NOTE: out of range 'screenmode' in game config crashes the game on start (LR2 bug)

LR2/Main.cpp
Skalim112: Done

LR2/Main.cpp
Skalim112: Done

Skalim112:
> NOTE: out of range 'screenmode' in game config crashes the game on start (LR2 bug)

Fixed.

I was able to reproduce GOMazk's issue. It only happens after changing the main monitor on the control panel and then opening the game immediately.

There seem to be two separate problems here:

1. **FIXED**: On dual horizontal monitors, DxLib can use a different display index order after the main monitor is changed at runtime. This caused exclusive mode to open on the opposite monitor. That part is fixed now by mapping the `GetDisplayInfo()` index to the index expected by `SetUseDisplayIndex()`.

2. **PENDING**: With a vertical monitor, DxLib/Direct3D can still pick the correct monitor but use the wrong fullscreen resolution/orientation after the same hot main-monitor change. That looks like a separate resolution/mode issue and needs further investigation.

For now, this fixes the display-index mismatch seen with horizontal dual-monitor setups. The vertical-monitor case looks like a separate DxLib/Direct3D mode-selection issue after a hot main-monitor change. After several tests, i could not get it resolver on the run, so I think that specific problem should be handled in a separate branch.

LR2/En_input.cpp
Skalim112: I removed that flag and now the conversion runs whenever borderless mode is active.

PR #157, opened 2026-06-24
Implement F/S rate bargraph and fix guage button opcode

MellDa1024:
image

( <- 260623 / This PR branch ->)
This PR implements F/S rate bargraph opcode (from the F/S patched LR2HD).
Opcodes 48 and 49 return Player 1's fast/slow ratio.  
According to the documentation, opcodes 58 and 59 should return Player 2's fast/slow ratio. However, this did not work in the specific LR2HD version I have.

Since the actual implementation details of LR2HD are unknown, there might be some differences. However, I believe the intended return values should remain the same.

This PR also fixes button opcode 41, which was returning Player 1's gauge instead of Player 2's.

I'm not sure if the F/S opcodes need to be added to OpcodeScript, and I'm also not exactly sure how to modify it.


GOMazk:
OpcodeScript is almost for SkinEditor, don't worry about that

PR #161, opened 2026-06-25
feat(customir): drive startup login and isOnline from CustomIR modules

luxurya111:
Replace dream-pro NETWORK::Login in Main with CUSTOMIR_MANAGER::Login; rename the legacy handler to LR2IR_Login. Record per-module login text in login_result and set isOnline when the configured display IR logs in.

By configuring this feature, all commands that previously depended on the isOnline variable (for example, the ranking not being displayed on the results screen in W-MIX) can now function correctly.

### DIFF
--- a/LR2/LR2_skinobject.cpp
+++ b/LR2/LR2_skinobject.cpp
@@ -176,7 +176,7 @@ bool GetOptionFlag_dst(game *gs, int option) {
 			break;
 
 		case 50:
-			if (!gs->net.customIR.IsDisplayIrOnline()) return ret;
+			if (gs->net.isOnline != 1 && !gs->net.customIR.IsDisplayIrOnline()) return ret;
 			break;
 
 		case 51:
--- a/LR2/LR2_skinobject.cpp
+++ b/LR2/LR2_skinobject.cpp
@@ -180,7 +180,7 @@ bool GetOptionFlag_dst(game *gs, int option) {
 			break;
 
 		case 51:
-			if (gs->net.customIR.IsDisplayIrOnline()) return ret;
+			if (gs->net.isOnline || gs->net.customIR.IsDisplayIrOnline()) return ret;
 			break;
 
 		case 52:
--- a/LR2/LR2_skinobject.cpp
+++ b/LR2/LR2_skinobject.cpp
@@ -176,11 +176,11 @@ bool GetOptionFlag_dst(game *gs, int option) {
 			break;
 
 		case 50:
-			if (gs->net.isOnline != 1) return ret;
+			if (!gs->net.customIR.IsDisplayIrOnline()) return ret;
 			break;
 
 		case 51:
-			if (gs->net.isOnline) return ret;
+			if (gs->net.customIR.IsDisplayIrOnline()) return ret;
 			break;
 
 		case 52:
--- a/LR2/Main.cpp
+++ b/LR2/Main.cpp
@@ -384,14 +384,6 @@ int main(int argc, char** argv) {
 		gs.net.IR_passMD5 = MD5str(gs.config.player.pass);
 		gs.net.getrival = gs.config.network.getrival;
 		gs.net.IR_ID = gs.gameplay.playerstat.irid;
-		if (gs.net.Login(gs.cmd_directplay) == 1) {
-			SaveIRID(gs.net.rankingData.myID, gs.config.player.id);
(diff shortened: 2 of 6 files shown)


chown2 (changes requested):
Let's do this differently
1. instead of having 'login_result' as a member variable, make CUSTOMIR_MANAGER::Login() return 'login_result' as a string.
  Look above your `ErrorLogAdd(gs.net.customIR.login_result.c_str());` change in main().
  Collect the result into a variable there.
  Later print from that collected variable
1. keep the list of IRs that were able to successfully log in by adding a member `std::vector<std::string> mLoggedInIrs;`
1. instead of assigning gs.net.isOnline, add a new method to CUSTOMIR_MANAGER - let's say `[[nodiscard]] bool IsDisplayIrOnline() const;`, and then call this method in skin accessors mentioned in the tracking issue https://github.com/GOMazk/OpenLR2/blob/0ea4a7f5120eb0fa0537ee466d9ca511ef88b4b9/LR2/LR2_skinobject.cpp#L178-L184
   The method would be implemented similar to this `return std::ranges::contains(mLoggedInIrs, mDisplayIr);`
   Such implementation will help us in future to switch between several display IRs.
   We don't want to touch gs.net.isOnline because there is existing code that will misbehave after that.
1. don't remove the part about logging into LR2IR in main(). I still use that. :D 
   Having LR2IR still work shouldn't break anything, I've kept that in mind while working on CustomIR APIs and have been checking that there are no regressions and both can work.
  Currently, they can even work in parallel, but we will want to be going away from that as it complicates code flow.

luxurya111:
Thank you for the review. I have applied the changes, so please check it out.

LR2/LR2_skinobject.cpp
```diff
@@ -176,11 +176,11 @@ bool GetOptionFlag_dst(game *gs, int option) {
 			break;
 
 		case 50:
-			if (gs->net.isOnline != 1) return ret;
+			if (!gs->net.customIR.IsDisplayIrOnline()) return ret;
```
chown2:
```suggestion
			if (gs->net.isOnline != 1 && !gs->net.customIR.IsDisplayIrOnline()) return ret;
```
It's imperfect, but let's keep LR2IR support for now. As I've said above, I still use LR2IR.

LR2/LR2_skinobject.cpp
```diff
@@ -176,11 +176,11 @@ bool GetOptionFlag_dst(game *gs, int option) {
...
 			break;
 
 		case 51:
-			if (gs->net.isOnline) return ret;
+			if (gs->net.customIR.IsDisplayIrOnline()) return ret;
```
chown2:
```suggestion
			if (gs->net.isOnline || gs->net.customIR.IsDisplayIrOnline()) return ret;
```
Same as above

LR2/Main.cpp
```diff
@@ -425,6 +426,8 @@ int main(int argc, char** argv) {
 			gs.net.ApplyInsaneList();
 		}
 		if (gs.is_starter == false) {
+			printfDx(loginResult.c_str());
+			printfDx("\n");
```
chown2, line 430: I'm fine with this as is, but doesn't loginResult already end with a newline?
```suggestion
```

chown2 (approved):
LGTM after you accept my changes. Thanks! Please squash when merging

LR2/Main.cpp
chown2: LR2IR integration also does that, although inconsistently.

chown2:
Hm, with my proposed changes both 50 and 51 DSTs can be active at once... Let me think about it

chown2:
False alarm. Please accept my suggestions and then LGTM! :rocket: 
```cpp
#include <initializer_list>
#include <print>
int main(int argc, char *argv[]) {
  for (bool is_online : {false, true}) {
    for (bool is_display_ir_online : {false, true}) {
      std::println("lr2ir={} customir={} 50={} 51={}", is_online, is_display_ir_online, !is_online && !is_display_ir_online, is_online || is_display_ir_online);
    }
  }
  return 0;
}
```
```
lr2ir=false customir=false 50=true 51=false
lr2ir=false customir=true 50=false 51=true
lr2ir=true customir=false 50=false 51=true
lr2ir=true customir=true 50=false 51=true
```

PR #167, opened 2026-06-27
feat(customir): add songPlayLevel and reserved tail fields to IRScoreV1

### LINKED ISSUE #163: [CustomIR] struct IRScoreV1 missing level field
This is a memo for future reference; since IRScoreV1 does not contain level information, the level data is missing from the IR.

Implementing IRScoreV2 might be necessary.

luxurya111:
Fixes https://github.com/GOMazk/OpenLR2/issues/163

The `level` field was not included in the `IRScoreV1` structure, preventing `CustomIR` from reading the level. This patch addresses that issue.
As advised in the issue, to avoid breaking ABI compatibility, `level` has been added to the very end of the structure, and reserved slots have also been included following the previous format.

### DIFF
--- a/LR2/LR2_customir_api.h
+++ b/LR2/LR2_customir_api.h
@@ -111,13 +111,10 @@ struct IRScoreV1 {
 	std::string ghostData;
 	// Chart level at play time (SONGDATA::level). 0 if unknown or not applicable.
 	int songPlayLevel{};
-	bool reserved1{};
-	bool reserved2{};
-	bool reserved3{};
+	uint64_t reserved1{};
+	uint64_t reserved2{};
+	uint64_t reserved3{};
 	uint64_t reserved4{};
-	uint64_t reserved5{};
-	uint64_t reserved6{};
-	uint64_t reserved7{};
 };
 
 enum class SendScoreStatus: int {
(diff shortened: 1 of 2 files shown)


LR2/LR2_customir_api.h
```diff
@@ -109,6 +109,15 @@ struct IRScoreV1 {
...
 	std::string ghostData;
+	// Chart level at play time (SONGDATA::level). 0 if unknown or not applicable.
+	int songPlayLevel{};
+	bool reserved1{};
+	bool reserved2{};
+	bool reserved3{};
+	uint64_t reserved4{};
+	uint64_t reserved5{};
+	uint64_t reserved6{};
+	uint64_t reserved7{};
```
chown2: The point of bools in IRGhostResult was to use up the padding bytes.
In C and C++, structs have alignment. It's the minimal amount of memory a field will take up, even if the field is smaller.
Booleans have a size of one byte, but due to alignment they can actually take up 4 or even 8 bytes. That's why IRGhostResult has had some of those 'bool reserved' fields, to occupy the bytes that would otherwise be used for padding used for alignment.
```suggestion
	int songPlayLevel{};
	uint64_t reserved1{};
	uint64_t reserved2{};
	uint64_t reserved3{};
	uint64_t reserved4{};
```

chown2:
Note: if you add 'Fixes #issuenumber' into the PR description, the issue will be automatically closed when the PR is merged

luxurya111:
> Note: if you add 'Fixes #issuenumber' into the PR description, the issue will be automatically closed when the PR is merged

Thank you for your advice!

chown2 (approved):
Tested. LGTM!

PR #168, opened 2026-06-27
fix(ghostbattle): fix ghost battle does not save score

luxurya111:
https://x.com/ntsjdjd/status/2070725151201841590 (Japanese)

Resolved the issue where the results of Ghost Battles were not being saved to the local DB.

### DIFF
--- a/LR2/LR2_statplay.cpp
+++ b/LR2/LR2_statplay.cpp
@@ -820,7 +820,7 @@ int SaveResult(game *g, sqlite3* sql) {
 			}
 		}
 
-		if (g->config.play.battle != 1 || g->is_starter) {
+		if (g->config.play.battle != 1 || g->is_starter || g->gameplay.ghostBattle) {
 
 			if ((g->gameplay.freqSpeedMultiplier < 1.0 || g->config.play.m_lunaris == 1) && g->is_starter == 0) {
 				if (g->gameplay.replay.status == 2) return -1;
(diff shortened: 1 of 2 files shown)


chown2:
Not RE issue. I wonder why this became necessary.

luxurya111:
As I recall, even in the original LR2, when a record was updated during a Ghost Battle, it was still saved. I have made the modification accordingly.

Is there any reason why this might not be considered a RE Issue?
If so, what actions should I take?

chown2:
> Is there any reason why this might not be considered a RE Issue?

When I say "RE issue", I mean direct differences to LR2body.exe as decompiled by Ghidra.
In this case, there was no 'ghostBattle' check in the SaveResult function of LR2body.exe.

> If so, what actions should I take?

For now, I don't know. Judging by this code, we might be incorrectly setting 'g->config.play.battle' somewhere. I will try taking a look around.

@GOMazk are there any plans to make the Ghidra project public?

GOMazk:
> @GOMazk are there any plans to make the Ghidra project public?

no problem. I'll upload it

luxurya111:
I thought it would be a simple condition modification, but it might turn out to be a trickier problem than expected.
I will check if it can be fixed from a different direction.

LR2/LR2_statplay.cpp
```diff
@@ -820,7 +820,7 @@ int SaveResult(game *g, sqlite3* sql) {
 			}
 		}
 
-		if (g->config.play.battle != 1 || g->is_starter) {
+		if (g->config.play.battle != 1 || g->is_starter || g->gameplay.ghostBattle) {
```
luxurya111: Upon checking, I found that G-BATTLE is assigned to number 4 in config.play.battle. Why is this conditional statement not passing?

chown2 (approved):

Issue #127, opened 2026-06-13
[Feature Request] Consider implementing EX-Hard gauge


boonsboos:
Since EX-Hard gauge was added to 2DX a couple years ago, it wouldn't be crazy to consider adding this new gauge type to LR2. Other BMS players also implement this.

Sonic-TV:
ASSISTED EASY as well, then.
Hopefully we can see 1%, 0.1%, and 0.01% groove gauge percentages in OpenLR2 (float digits and odd digits for the gauge).

BKVad1m:
> ASSISTED EASY as well, then.
> Hopefully we can see 1%, 0.1%, and 0.01% groove gauge percentages in OpenLR2 (float digits and odd digits for the gauge).

that's all about to migrate from Lunatic Vibes

PR #176, opened 2026-06-28
refactor: miscellaneousness

### LINKED ISSUE #39: plan refactor numbers into enums
> ideally dirty work must be done and all of those magic numbers around the code should be made into enums. 

 _Originally posted by @MatVeiQaaa in [#11](https://github.com/GOMazk/OpenLR2/issues/11#issuecomment-3568942320)_

this is currently just todo list

ez2patternBMS:
Remove some hardcoding or large bundles of code
Fixes #39 (chown2: mostly, I don't see the need to keep the issue around)


LR2/LR2_bmsload.cpp
```diff
@@ -748,16 +641,13 @@ int InitGameplay_retry(gameplay *gp, AUDIO *snd, game *g) {
...
-	snd->param.stageKeyVolume[1] = 1.0;
-	snd->param.stageKeyVolume[2] = 1.0;
-	snd->param.stageKeyVolume[3] = 1.0;
-	snd->param.stageKeyVolume[4] = 1.0;
-	snd->param.stageKeyVolume[0] = 1.0;
+	for (int i = 1; i < 10; i++){
+		snd->param.stageBgmVolume[i] = 0.0;
+	}
+	for (int i = 0; i < 10; i++){
+		snd->param.stageKeyVolume[i] = 0.0;
```
GOMazk: 1.0, not 0.0

MatVeiQaaa:
oh wow, thanks a lot, something i believe we all knew had to be done but were lazy to do.

Unengine:
If it is not ready for review for now, I recommend you to make it draft.

ez2patternBMS:
Okay, I was going to PR only 3 commit, but new commits are continuously being added.

BTW, is it okay to change `char` to `bool`?

For example, `char m_lunaris{};`

LR2/LR2_statplay.cpp
```diff
@@ -64,74 +64,12 @@ uint ConvertOptionHistory(game *g){
...
-				case 4:
-					return ret | 0x100000;
-				case 5:
-					return ret | 0x200000;
-				case 6:
-					return ret | 0x400000;
-				case 7:
-					return ret | 0x800000;
-			}
+			if (gauge < 8)
```
chown2: I'd revert this hunk.
IMO, while it's true that you can use bit operations here, if we replace magic numbers with enum values, it will be more explicit.
And if we don't add 'default' case here, the compiler will warn us on new values being added, which will assist us implement #127, if we decide it's something we want.

LR2/LR2_bmsload.cpp
```diff
@@ -16,6 +16,461 @@
 
 #include <DxLib/DxLib.h>
 
+unsigned char channelConvertIIDX[] = {  -1,                     //00
```
chown2: Maybe channelConvertBME would be a better name?
https://hitkey.nekokan.dyndns.info/cmds.htm#KEYMAP-TABLE

LR2/LR2_statplay.cpp
ez2patternBMS: Yeah, but this function is not considering whether it's actual value.
As you can see, it handles gauge 8 option, although it actually have only 6 gauge option, same as random.

LR2/LR2_bmsload.cpp
chown2: Please add 'static constexpr' before these arrays

chown2:
> BTW, is it okay to change char to bool?

If it's only used as 0 and 1 - definitely

chown2:
I trust you constants got replaced correctly, because it's impossible to eyeball review such refactorings. Thanks a lot, this is a very good refactoring!

LR2/LR2_statplay.cpp
ez2patternBMS: If you want to put more than 8 things in one option related, you'll have to rip and fix all the routines related to this anyway.
Just keep this commit and let's add compiler warning when when we add 9th thing.

chown2: When we change the types of these values to 'enum class ThisThing', the code will change into it's previous form again

> it handles gauge 8 option

You can remove those unnecessary cases
m_HIDSUD1 appears to also be wrong, it should be in range 0-3 https://github.com/GOMazk/OpenLR2/blob/15b2fc066a8bdbfcc28d3d2a88e6467048f7bd7f/LR2/Scene02_Songselect.cpp#L110

ez2patternBMS:
> I trust you constants got replaced correctly

I'd little hasted so maybe there's come typo, let me check again.

LR2/LR2_statplay.cpp
chown2: Well, I'm fine with these changes too.
But please add `else ErrorLogFmtAdd("BUG: invalid gauge: %d", gauge);` (and similarly to the other 2 values) and fix up the ranges, `gauge <= 5`, `random <= 5`, `m_HIDSUD1 <= 3`.

LR2/LR2_bmsload.cpp
ez2patternBMS: bms/bml/bme classification is meaningless on LR2 era client, so how about `channelConvertBMS` and `channelConvertPMS`

chown2: Sure :+1:

LR2/LR2_statplay.cpp
ez2patternBMS: Still this code can not handle 9 things, I just wrote something like this.
```c++
			if (gauge <= OPTION_GAUGE_END && gauge < 8)
				ret = ret | 0x1 << gauge ;
			else
				ErrorLogFmtAdd("BUG: invalid gague: %d", gauge);
```

It's right to use compile error, but I'm lazy.

LR2/LR2_statplay.cpp
```diff
@@ -64,74 +64,18 @@ uint ConvertOptionHistory(game *g){
...
+			else
+				ErrorLogFmtAdd("BUG: invalid gague: %d", gauge);
+			if (g->config.play.random[0] <= OPTION_RANDOM_END && g->config.play.random[0] < 8)
+				ret = ret | 0x100 << g->config.play.random[0];
+			else
+				ErrorLogFmtAdd("BUG: invalid random: %d", g->config.play.random);
+			if (g->config.play.m_HIDSUD1 <= 3 && g->config.play.m_HIDSUD1 < 8)
+				ret = ret | 0x10000 << g->config.play.m_HIDSUD1;
+			else
+				ErrorLogFmtAdd("BUG: invalid m_HIDSUD1: %d", g->config.play.m_HIDSUD1);
```
chown2:
```suggestion
			if (gauge <= OPTION_GAUGE_END)
				ret |= 0x1 << gauge;
			else
				ErrorLogFmtAdd("BUG: invalid gauge: %d", gauge);
			if (g->config.play.random[0] <= OPTION_RANDOM_END)
				ret |= 0x100 << g->config.play.random[0];
			else
				ErrorLogFmtAdd("BUG: invalid random: %d", g->config.play.random);
			if (g->config.play.m_HIDSUD1 <= 3)
				ret |= 0x10000 << g->config.play.m_HIDSUD1;
			else
				ErrorLogFmtAdd("BUG: invalid m_HIDSUD1: %d", g->config.play.m_HIDSUD1);
```

ez2patternBMS: The reason why I keep adding `< 8` is
For example, assuming a 9th gauge is added, `1 << 8` is `0x100`, doesn't it overlap with bits of `config.play.random[0]` field?

ez2patternBMS:
I found the typo as much as I can. I want to squash, but I can't because my git is spagettied.

I'm not familiar with git.

LR2/LR2_statplay.cpp
chown2: `static_assert(OPTION_GAUGE_END < 8);`

LR2/structure.h
```diff
@@ -90,23 +90,196 @@ typedef enum SKINTYPE {
...
+	OPTION_GAUGE_EASY,
+	OPTION_GAUGE_PATTCK,
+	OPTION_GAUGE_GATTACK,
+
+	OPTION_RANDOM_OFF = 0,
+	OPTION_RANDOM_MIRROR,
+	OPTION_RANDOM_RANDOM,
+	OPTION_RANDOM_SRANDOM,
+	OPTION_RANDOM_SCATTER,
+	OPTION_RANDOM_CONVERGE,
```
chown2, line 278: Does this work instead of #define? #define is evil
```suggestion
	OPTION_RANDOM_END = OPTION_RANDOM_CONVERGE,
```

LR2/LR2_songmanage.cpp
```diff
@@ -3073,12 +3012,13 @@ int ParseBMSMETA(BMSMETA *meta, CSTR filepath, char flag) {
 			}
 			else if (lineLen > 5 && IsDigitAscii(line[1]) && IsDigitAscii(line[2]) && IsDigitAscii(line[3]) && IsDigitAscii(line[4]) && IsDigitAscii(line[5])) {
 				//TOFIX : There is no need to decrease variable data, still it does. it makes karinote include invisible and longnote(0.5 to 1)
+				//TOFIX : This part can't use CHANNEL directly because they are not converted to LR2's inner channel
```
chown2: TOFIX has been used for LR2 bugs in the code base so far, so it's better to use a TODO here

LR2/LR2_songmanage.cpp
```diff
@@ -3089,19 +3029,19 @@ int ParseBMSMETA(BMSMETA *meta, CSTR filepath, char flag) {
 					else if (data == 28 || data == 29 || data == 48 || data == 49 || data == 68 || data == 69) {
 						meta->keymode = 14;
 					}
-					else if ((data > 20 && data < 30) || (data > 40 && data < 50) || (61 <= data && data < 70)) {
+					else if ((data > CHANNEL_2P_NOTE_SC && data < CHANNEL_2P_NOTE_END + 1) || (data > CHANNEL_2P_HIDDEN_SC && data < CHANNEL_2P_HIDDEN_END + 1) || (CHANNEL_2P_LN_SC + 1 <= data && data < CHANNEL_2P_LN_END + 1)) {//TOFIX: inconsitency
```
chown2: Same as above, TOFIX -> TODO

LR2/structure.h
ez2patternBMS: I haven't used enum like that, I'll fix it.

LR2/LR2_statplay.cpp
```diff
@@ -63,17 +63,20 @@ uint ConvertOptionHistory(game *g){
 			return ret;
 		}
 		if (clear > 2) {
+			static_assert(OPTION_GAUGE_END < 8, "OPTION_GAUGE_END must be less than 8");
```
chown2: This error message is redundant, modern compilers will even tell the value of OPTION_GAUGE_END on error :)

LR2/LR2_bmsload.cpp
```diff
@@ -16,6 +16,461 @@
 
 #include <DxLib/DxLib.h>
 
+static constexpr unsigned char channelConvertBMS[] = {  -1, //00
+									CHANNEL_BGM,            //01
```
chown2: Please move the first element onto the next line and reduce the indentation
```suggestion
static constexpr unsigned char channelConvertBMS[] = { 
	-1, //00
	CHANNEL_BGM,            //01
```

LR2/LR2_statplay.cpp
ez2patternBMS: done

LR2/LR2_statplay.cpp
```diff
@@ -63,75 +63,22 @@ uint ConvertOptionHistory(game *g){
...
+			else
+				ErrorLogFmtAdd("BUG: invalid gague: %d", gauge);
+			if (g->config.play.random[0] <= OPTION_RANDOM_END && g->config.play.random[0] < 8)
+				ret = ret | 0x100 << g->config.play.random[0];
+			else
+				ErrorLogFmtAdd("BUG: invalid random: %d", g->config.play.random);
+			if (g->config.play.m_HIDSUD1 <= 3 && g->config.play.m_HIDSUD1 < 8)
+				ret = ret | 0x10000 << g->config.play.m_HIDSUD1;
+			else
+				ErrorLogFmtAdd("BUG: invalid m_HIDSUD1: %d", g->config.play.m_HIDSUD1);
```
chown2: With the static_asserts above, these are now redundant
```suggestion
			if (gauge <= OPTION_GAUGE_END)
				ret = ret | 0x1 << gauge ;
			else
				ErrorLogFmtAdd("BUG: invalid gague: %d", gauge);
			if (g->config.play.random[0] <= OPTION_RANDOM_END)
				ret = ret | 0x100 << g->config.play.random[0];
			else
				ErrorLogFmtAdd("BUG: invalid random: %d", g->config.play.random);
			if (g->config.play.m_HIDSUD1 <= 3)
				ret = ret | 0x10000 << g->config.play.m_HIDSUD1;
			else
				ErrorLogFmtAdd("BUG: invalid m_HIDSUD1: %d", g->config.play.m_HIDSUD1);
```

chown2:
3c8fbfbe61795c614b57c98cacbd8164e79770f0 "fix: fix wrong FP/DSC channel mapping and add BME-type support"
Let's do refactorings only for now. There is no point adding support for something not used anywhere. Maybe we can even remove "FP/DSC" support completely later.

LR2/LR2_statplay.cpp
ez2patternBMS: Missed by accident. Fixed

chown2 (commented):
Looks very good at a glance, but I'd like to look at it again later with a fresh mind

ez2patternBMS:
> [3c8fbfb](https://github.com/GOMazk/OpenLR2/commit/3c8fbfbe61795c614b57c98cacbd8164e79770f0) "fix: fix wrong FP/DSC channel mapping and add BME-type support" Let's do refactorings only for now. There is no point adding support for something not used anywhere. Maybe we can even remove "FP/DSC" support completely later.

There're skin and chart FP/DSC. they just couldn't use it because I didn't know "#FP/DSC" existed, so it's easy to make some kind of somthing related FP/DSC
However, it's out of context of refactoriing as you said, so let's take it out.

I cannot rebase because this commit is before [2d18e23](https://github.com/GOMazk/OpenLR2/pull/176/changes/2d18e2333356d2c18dadfbe52b919b6b96c4ac83)merge, so I reverted

MatVeiQaaa:
> Okay, I was going to PR only 3 commit, but new commits are continuously being added.
> 
> BTW, is it okay to change `char` to `bool`?
> 
> For example, `char m_lunaris{};`

please do, it would be amazing.

LR2/structure.h
```diff
@@ -274,7 +274,7 @@ typedef enum OPTION {
 	OPTION_RANDOM_CONVERGE,
 	OPTION_RANDOM_END = OPTION_RANDOM_CONVERGE,
 
-	//Maybe just use bool?
+	//Maybe better use bool?
```
MatVeiQaaa: i personally would rather have it as bool.

ez2patternBMS: Maybe next PR, I'm a little exhusted.
And I used them because somthing like `LoopInRange(false, true, 1,~~)`  looks cursed

chown2: 1) next PR, definitely
2) :D `value = !value`

ez2patternBMS: This commit is reseted. Let's handle more deeply at next PR

ez2patternBMS:
Discussion for refactoring;

There're two naming Convention used.

```c++
    struct gameplay {
        //snake_case
	int bpmt_count;
	int bpmt_buffersize;
	int bpmt_start;

        //camelCase
	double freqSpeedMultiplier;
	char fxChangeInRecording;
	int procGameCallCount;
	bool isSpeedChanged; 
	char trialClear; 
}
```
And name of variable is different from xml file's one.
```c++
        //xml is hstype, but code is hsfix
	ReadXml_Int("config", "play", "hstype", OPTION_HS_FIX_OFF, &g->config.play.hsfix, hXml);
```
And boolean's name also differ from each other
```c++
	char flag_gameinput;
	char isPreviewLoad;
	int hasTxt;
```
so I suggest for 3 things.

* use snake_case for variable (or camelCase, just use one)
* set order of priority to xml variable name
* don't use `flag`, use `is_`/`has_`/`can_` ...etc

LR2/LR2_bmsload.cpp
```diff
@@ -16,6 +16,461 @@
 
 #include <DxLib/DxLib.h>
 
+static constexpr unsigned char channelConvertBMS[] = {-1, //00
```
chown2: > error: narrowing conversion of ‘-1’ from ‘int’ to ‘unsigned char’ [-Wnarrowing]

chown2:
> Original switch-case was hard too read, cause of that, there was also bug.

What bug? [fix: fix wrong FP/DSC channel mapping and add BME-type support](https://github.com/GOMazk/OpenLR2/pull/176/commits/de9b353158f1fb72b415700c8e4d0206d14f5a49)?

LR2/LR2_bmsload.cpp
chown2: I will convert this into a function with a switch case myselfl There is a bit of a mess with the commit order.

chown2:
I've cleaned this up myself, but in future PRs, please, try to avoid commits depending on each other.
Both logically and especially with fixup commits.

Logically my example is `enum OPTION`, each commit should have added a new enum instead of adding discriminants to a single new one.

    Revert "fix: fix wrong FP/DSC channel mapping and add BME-type support"
    refactor: remove #define and use enum
    refactor: change channelConvert name and type
    refactor: use channel enum for some missing part
    fix: Fix channel constant typo

And these are essentially fixup commits, but the changes between commits all depend on each other, so they were not directly mergable in the initial commits.

chown2:
TODO: add editorconfig file to trim whitespaces. This PR shows why.
https://editorconfig.org/#supported-properties

LR2/LR2_bmsload.h
```diff
@@ -42,3 +42,148 @@ int DPsplitLaneScratch(LaneStruct *lane, int start, CHARTCONVERTER *cc);
 int SPtoDP(LaneStruct * lane, int baseNoteID, CHARTCONVERTER * cc);
 
 int ParseBmsFile(gameplay * gp, CSTR filename, AUDIO * aud, ConfigStruct * cfg, BMSMETA * meta, int bgaFlag, int scratchSide);
+
```
chown2: (for myself) do this in this commit instead of spreading over two

LR2/structure.h
```diff
@@ -258,6 +258,13 @@ enum OPTION {
 	OPTION_BATTLE_SP2DP, //not only sp-to-dp, dp-to-sp and pms-to-sp
 	OPTION_BATTLE_GBATTLE,
 	OPTION_BATTLE_END = OPTION_BATTLE_GBATTLE,
+
+	OPTION_GAUGE_GROOVE = 0,
+	OPTION_GAUGE_HARD,
+	OPTION_GAUGE_DEATH,
+	OPTION_GAUGE_EASY,
+	OPTION_GAUGE_PATTCK,
+	OPTION_GAUGE_GATTACK,
```
chown2, line 270: (for myself) lost OPTION_GAUGE_END

LR2/LR2_bmsload.cpp
```diff
@@ -745,16 +1096,13 @@ int InitGameplay_retry(gameplay *gp, AUDIO *snd, game *g) {
...
-	snd->param.stageBgmVolume[2] = 0.0;
-	snd->param.stageBgmVolume[3] = 0.0;
-	snd->param.stageBgmVolume[4] = 0.0;
 	snd->param.stageBgmVolume[0] = 1.0;
-	snd->param.stageKeyVolume[1] = 1.0;
-	snd->param.stageKeyVolume[2] = 1.0;
-	snd->param.stageKeyVolume[3] = 1.0;
-	snd->param.stageKeyVolume[4] = 1.0;
-	snd->param.stageKeyVolume[0] = 1.0;
+	for (int i = 1; i < 10; i++){
```
chown2: (I will fix this) Avoid raw array indexing, all of these should be range loops
`warning: iteration 4 invokes undefined behavior [-Waggressive-loop-optimizations]`

LR2/LR2_bmsload.cpp
```diff
@@ -3379,58 +3490,33 @@ int ParseBmsFile(gameplay *gp, CSTR filename, AUDIO *aud, ConfigStruct* cfg, BMS
...
-				noteRandomTable[0][1] = 9;
-				noteRandomTable[0][2] = 8;
-				noteRandomTable[0][3] = 7;
-				noteRandomTable[0][4] = 6;
-				noteRandomTable[0][5] = 5;
-				noteRandomTable[0][6] = 4;
-				noteRandomTable[0][7] = 3;
-				noteRandomTable[0][8] = 2;
-				noteRandomTable[0][9] = 1;
+				for (int i = 0; i < 9; i++)
```
chown2: This differs from old code but the original code looks weird

chown2: It was the same in original LR2. Do notes start from index 1 in PMS?

LR2/LR2_bmsload.cpp
```diff
@@ -1321,17 +1321,17 @@ void MakeExtraChart(gameplay *gp, CHARTCONVERTER *cc) {  //test completed
 	qsort(gp->bmsobj.notes, gp->bmsobj.count, sizeof(NoteStruct), CMP_NotesByRealTimingOp);
 
 	for (int i = 0; i < gp->bmsobj.count; i++) {
-		if (10 <= gp->bmsobj.notes[i].op && gp->bmsobj.notes[i].op <= 30) {
+		if (CHANNEL_1P_NOTE_SC <= gp->bmsobj.notes[i].op && gp->bmsobj.notes[i].op <= CHANNEL_1P_HIDDEN_SC) {
```
chown2, line 1141: Might be an LR2 bug and 29 (CHANNEL_2P_NOTE_END) was meant instead of 30 (CHANNEL_1P_HIDDEN_SC)

LR2/LR2_bmsload.cpp
```diff
@@ -1357,7 +1357,7 @@ void MakeExtraChart(gameplay *gp, CHARTCONVERTER *cc) {  //test completed
 		laneOfSound[i] = -1;
 
 		for (int j = 0; j < gp->bmsobj.count; j++) {
-			if (10 <= gp->bmsobj.notes[j].op && gp->bmsobj.notes[j].op <= 30) {
+			if (CHANNEL_1P_NOTE_SC <= gp->bmsobj.notes[j].op && gp->bmsobj.notes[j].op <= CHANNEL_2P_NOTE_END + 1) {
```
chown2: Same

LR2/LR2_bmsload.cpp
chown2: They do

LR2/LR2_bmsload.cpp
```diff
@@ -1811,12 +1811,12 @@ void PMStoSP(gameplay *gp) { //test&fix completed
...
 					ErrorLogFmtAdd("つめるレーンは%dと%d\n", emptyLaneL, emptyLaneR);
 					for (int j = prev; j < i; j++) {
-						if (emptyLaneL + 10 < gp->bmsobj.notes[j].op && gp->bmsobj.notes[j].op <= 19) {
+						if (emptyLaneL + 10 < gp->bmsobj.notes[j].op && gp->bmsobj.notes[j].op <= CHANNEL_PMS_NOTE_9) {
 							gp->bmsobj.notes[j].op--;
 						}
 					}
 					for (int j = prev; j < i; j++) {
-						if (emptyLaneR + 9 < gp->bmsobj.notes[j].op && gp->bmsobj.notes[j].op <= 19) {
+						if (emptyLaneR + 9 < gp->bmsobj.notes[j].op && gp->bmsobj.notes[j].op <= CHANNEL_PMS_NOTE_9) {
```
chown2: + 9 instead of + 10? Might be an LR2 bug

chown2:
I've went through everything other than `refactor: add dedicated converter functions for BMS channels` with my eyes. Gotta review that commit + test this.

ez2patternBMS:
> > Original switch-case was hard too read, cause of that, there was also bug.
> 
> What bug? [fix: fix wrong FP/DSC channel mapping and add BME-type support](https://github.com/GOMazk/OpenLR2/pull/176/commits/de9b353158f1fb72b415700c8e4d0206d14f5a49)?

Yes, it just fixed wrong channel mapping.

> I've cleaned this up myself, but in future PRs, please, try to avoid commits depending on each other. Both logically and especially with fixup commits.
> 
> Logically my example is `enum OPTION`, each commit should have added a new enum instead of adding discriminants to a single new one.
> 
> ```
> Revert "fix: fix wrong FP/DSC channel mapping and add BME-type support"
> refactor: remove #define and use enum
> refactor: change channelConvert name and type
> refactor: use channel enum for some missing part
> fix: Fix channel constant typo
> ```
> 
> And these are essentially fixup commits, but the changes between commits all depend on each other, so they were not directly mergable in the initial commits.

I also wanted to clean up, but I'm still not used to git. I need to study how to use commands like rebase properly. Anyway, I'll be careful.

ez2patternBMS:
> Logically my example is `enum OPTION`, each commit should have added a new enum instead of adding discriminants to a single new one.

Letme double check, should I write code like A instead of B if they are not single commit?
```c
//A
typedef enum OPTION_HS_FIX{
       OPTION_HS_FIX_OFF = 0,
       ...
}
typedef enum OPTION_RANDOM{
       OPTION_RANDOM_OFF = 0,
       ...
}

//b
typedef enum OPTION{
       OPTION_HS_FIX_OFF = 0,
       ...
       OPTION_RANDOM_OFF = 0,
       ...
}
```
I saw some project did like this, so I code it same way.
It should not be, I'll try to avoid them.

chown2:
Option A.

Also, in your example (and in the code in this PR before my changes), the typedefs are wrong as they are never given the type name.
`typedef enum OPTION{};` -> `typedef enum OPTION{} OPTION;`.
But such construct only makes sense in C.
In C++ you don't need typedefs for enums, `enum OPTION{};` does the same thing.

chown2:
@ez2patternBMS how did come up with this mapping? 8b1190b99852b5ea292c61eb715cf529569c4da4
I'm trying to replicate these changes to understand them

chown2:
I broke it while converting to switch case, I see

chown2:
For reference, here is how I replicated the conversion
By the way, you didn't handle the case where both isPMS and isDSC are true

```cpp
static int convert(int channel) {
  switch(channel) {
    case 0x0: return -1;
    case 0xa: return -1;
    case 0x10: return 0xa;
    case 0x11: return -1;
    case 0x12: return 0x10;
    case 0x13: return 0x11;
    case 0x14: return -1;
    case 0x1a: return 0x14;
    case 0x1b: return -1;
    case 0x1c: return 0x1a;
    case 0x1d: return 0x1b;
    case 0x1e: return -1;
    case 0x24: return 0x1e;
    case 0x25: return -1;
    case 0x26: return 0x24;
    case 0x27: return 0x25;
    case 0x28: return -1;
    case 0x2e: return 0x28;
    case 0x2f: return -1;
    case 0x30: return 0x2e;
    case 0x31: return 0x2f;
    case 0x32: return -1;
    case 0x38: return 0x32;
    case 0x39: return -1;
    case 0x3a: return 0x38;
    case 0x3b: return 0x39;
    case 0x3c: return -1;
    case 0x42: return 0x3c;
    case 0x43: return -1;
    case 0x44: return 0x42;
    case 0x45: return 0x43;
    case 0x82: return -1;
    case 0x88: return 0x82;
    case 0x89: return -1;
    case 0x8a: return 0x88;
    case 0x8b: return 0x89;
    case 0x8c: return -1;
    case 0x92: return 0x8c;
    case 0x93: return -1;
    case 0x94: return 0x92;
    case 0x95: return 0x93;
    default: return -2;
  }
}
static int convertDSC(int channel) {
  switch(channel) {
    case 0x0: return -1;
    case 0xa: return -1;
    case 0xb: return 0xc;
    case 0xc: return 0xd;
    case 0xd: return 0xe;
    case 0xe: return 0xf;
    case 0xf: return 0x10;
    case 0x10: return 0xb;
    case 0x11: return -1;
    case 0x12: return 0x11;
    case 0x13: return 0x12;
    case 0x14: return -1;
    case 0x1a: return 0x13;
    case 0x1b: return -1;
    case 0x1c: return 0x1a;
    case 0x1d: return 0x1b;
    case 0x1e: return -1;
    case 0x1f: return 0x20;
    case 0x20: return 0x21;
    case 0x21: return 0x22;
    case 0x22: return 0x23;
    case 0x23: return 0x24;
    case 0x24: return 0x1f;
    case 0x25: return -1;
    case 0x26: return 0x25;
    case 0x27: return 0x26;
    case 0x28: return -1;
    case 0x2e: return 0x27;
    case 0x2f: return -1;
    case 0x30: return 0x2e;
    case 0x31: return 0x2f;
    case 0x32: return -1;
    case 0x33: return 0x34;
    case 0x34: return 0x35;
    case 0x35: return 0x36;
    case 0x36: return 0x37;
    case 0x37: return 0x38;
    case 0x38: return 0x33;
    case 0x39: return -1;
    case 0x3a: return 0x39;
    case 0x3b: return 0x3a;
    case 0x3c: return -1;
    case 0x42: return 0x3b;
    case 0x43: return -1;
    case 0x44: return 0x42;
    case 0x45: return 0x43;
    case 0x82: return -1;
    case 0x88: return 0x83;
    case 0x89: return -1;
    case 0x8a: return 0x89;
    case 0x8b: return 0x8a;
    case 0x8c: return -1;
    case 0x92: return 0x8b;
    case 0x93: return -1;
    case 0x94: return 0x92;
    case 0x95: return 0x93;
    default: return -2;
  }
}
static int convertPMS(int channel) {
  switch(channel) {
    case 0x0: return -1;
    case 0xa: return -1;
    case 0x10: return 0x12;
    case 0x11: return 0x13;
    case 0x12: return 0x10;
    case 0x13: return 0x11;
    case 0x14: return -1;
    case 0x16: return 0x10;
    case 0x17: return 0x11;
    case 0x18: return 0x12;
    case 0x19: return 0x13;
    case 0x1a: return 0x14;
    case 0x1b: return -1;
    case 0x1c: return 0x1a;
    case 0x1d: return 0x1b;
    case 0x1e: return -1;
    case 0x24: return 0x26;
    case 0x25: return 0x27;
    case 0x26: return 0x24;
    case 0x27: return 0x25;
    case 0x28: return -1;
    case 0x2a: return 0x24;
    case 0x2b: return 0x25;
    case 0x2c: return 0x26;
    case 0x2d: return 0x27;
    case 0x2e: return 0x28;
    case 0x2f: return -1;
    case 0x30: return 0x2e;
    case 0x31: return 0x2f;
    case 0x32: return -1;
    case 0x38: return 0x3a;
    case 0x39: return 0x3b;
    case 0x3a: return 0x38;
    case 0x3b: return 0x39;
    case 0x3c: return -1;
    case 0x3e: return 0x38;
    case 0x3f: return 0x39;
    case 0x40: return 0x3a;
    case 0x41: return 0x3b;
    case 0x42: return 0x3c;
    case 0x43: return -1;
    case 0x44: return 0x42;
    case 0x45: return 0x43;
    case 0x82: return -1;
    case 0x88: return 0x8a;
    case 0x89: return 0x8b;
    case 0x8a: return 0x88;
    case 0x8b: return 0x89;
    case 0x8c: return -1;
    case 0x8e: return 0x88;
    case 0x8f: return 0x89;
    case 0x90: return 0x8a;
    case 0x91: return 0x8b;
    case 0x92: return 0x8c;
    case 0x93: return -1;
    case 0x94: return 0x92;
    case 0x95: return 0x93;
    default: return -2;
  }
}
static int convertPMSDSC(int channel) {
  switch(channel) {
    case 0x0: return -1;
    case 0xa: return -1;
    case 0xb: return 0xc;
    case 0xc: return 0xd;
    case 0xd: return 0xe;
    case 0xe: return 0xf;
    case 0xf: return 0x10;
    case 0x10: return 0xb;
    case 0x11: return 0x13;
    case 0x12: return 0x11;
    case 0x13: return 0x12;
    case 0x14: return -1;
    case 0x16: return 0x10;
    case 0x17: return 0x11;
    case 0x18: return 0x12;
    case 0x19: return 0x13;
    case 0x1a: return 0x13;
    case 0x1b: return -1;
    case 0x1c: return 0x1a;
    case 0x1d: return 0x1b;
    case 0x1e: return -1;
    case 0x1f: return 0x20;
    case 0x20: return 0x21;
    case 0x21: return 0x22;
    case 0x22: return 0x23;
    case 0x23: return 0x24;
    case 0x24: return 0x1f;
    case 0x25: return 0x27;
    case 0x26: return 0x25;
    case 0x27: return 0x26;
    case 0x28: return -1;
    case 0x2a: return 0x24;
    case 0x2b: return 0x25;
    case 0x2c: return 0x26;
    case 0x2d: return 0x27;
    case 0x2e: return 0x27;
    case 0x2f: return -1;
    case 0x30: return 0x2e;
    case 0x31: return 0x2f;
    case 0x32: return -1;
    case 0x33: return 0x34;
    case 0x34: return 0x35;
    case 0x35: return 0x36;
    case 0x36: return 0x37;
    case 0x37: return 0x38;
    case 0x38: return 0x33;
    case 0x39: return 0x3b;
    case 0x3a: return 0x39;
    case 0x3b: return 0x3a;
    case 0x3c: return -1;
    case 0x3e: return 0x38;
    case 0x3f: return 0x39;
    case 0x40: return 0x3a;
    case 0x41: return 0x3b;
    case 0x42: return 0x3b;
    case 0x43: return -1;
    case 0x44: return 0x42;
    case 0x45: return 0x43;
    case 0x82: return -1;
    case 0x88: return 0x83;
    case 0x89: return 0x8b;
    case 0x8a: return 0x89;
    case 0x8b: return 0x8a;
    case 0x8c: return -1;
    case 0x8e: return 0x88;
    case 0x8f: return 0x89;
    case 0x90: return 0x8a;
    case 0x91: return 0x8b;
    case 0x92: return 0x8b;
    case 0x93: return -1;
    case 0x94: return 0x92;
    case 0x95: return 0x93;
    default: return -2;
  }
}
```
Here is how I made it
```cpp
#include <cstdlib>
#include <print>

static unsigned char channelConvert[] = {
    0x00, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x00, 0x01,
    0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x00, 0x3c, 0x0a, 0x0b,
    0x0c, 0x0d, 0x0e, 0x00, 0x0f, 0x10, 0x00, 0x11, 0x12, 0x13, 0x14, 0x15,
    0x16, 0x17, 0x18, 0x19, 0x00, 0x3c, 0x1a, 0x1b, 0x1c, 0x1d, 0x1e, 0x00,
    0x1f, 0x20, 0x00, 0x21, 0x22, 0x23, 0x24, 0x25, 0x26, 0x27, 0x28, 0x29,
    0x00, 0x3c, 0x2a, 0x2b, 0x2c, 0x2d, 0x2e, 0x00, 0x2f, 0x30, 0x3c, 0x3c,
    0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c,
    0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c,
    0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c,
    0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c,
    0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x3c, 0x00, 0x3c,
    0x3c, 0x3c, 0x3c, 0x3c, 0x31, 0x32, 0x33, 0x34, 0x00, 0x3c, 0x35, 0x36,
    0x37, 0x38, 0x39, 0x00, 0x3a, 0x3b};

static int doConvert(int channel, bool isPMS, bool isDSC) {
  switch (channelConvert[channel]) {
  case 0:
    channel = -1;
    break;

  case 1:
    if (isDSC)
      channel = 12;
    break;
  case 2:
    if (isDSC)
      channel = 13;
    break;
  case 3:
    if (isDSC)
      channel = 14;
    break;
  case 4:
    if (isDSC)
      channel = 15;
    break;
  case 5:
    if (isDSC)
      channel = 16;
    break;
  case 6:
    if (isDSC)
      channel = 11;
    else if (isPMS)
      channel = 18;
    else
      channel = 10;
    break;
  case 7:
    if (isPMS)
      channel = 19;
    else
      channel = -1;
    break;
  case 8:
    if (isDSC)
      channel = 17;
    else
      channel = 16;
    break;
  case 9:
    if (isDSC)
      channel = 18;
    else
      channel = 17;
    break;
  case 10:
    if (isPMS)
      channel = 16;
    break;
  case 11:
    if (isPMS)
      channel = 17;
    break;
  case 12:
    if (isPMS)
      channel = 18;
    break;
  case 13:
    if (isPMS)
      channel = 19;
    break;
  case 14:
    if (isDSC)
      channel = 19;
    else
      channel = 20;
    break;
  case 15:
    channel = 26;
    break;
  case 16:
    channel = 27;
    break;
  case 17:
    if (isDSC)
      channel = 32;
    break;
  case 18:
    if (isDSC)
      channel = 33;
    break;
  case 19:
    if (isDSC)
      channel = 34;
    break;
  case 20:
    if (isDSC)
      channel = 35;
    break;
  case 21:
    if (isDSC)
      channel = 36;
    break;
  case 22:
    if (isDSC)
      channel = 31;
    else if (isPMS)
      channel = 38;
    else
      channel = 30;
    break;
  case 23:
    if (isPMS)
      channel = 39;
    else
      channel = -1;
    break;
  case 24:
    if (isDSC)
      channel = 37;
    else
      channel = 36;
    break;
  case 25:
    if (isDSC)
      channel = 38;
    else
      channel = 37;
    break;
  case 26:
    if (isPMS)
      channel = 36;
    break;
  case 27:
    if (isPMS)
      channel = 37;
    break;
  case 28:
    if (isPMS)
      channel = 38;
    break;
  case 29:
    if (isPMS)
      channel = 39;
    break;
  case 30:
    if (isDSC)
      channel = 39;
    else
      channel = 40;
    break;
  case 31:
    channel = 46;
    break;
  case 32:
    channel = 47;
    break;
  case 33:
    if (isDSC)
      channel = 52;
    break;
  case 34:
    if (isDSC)
      channel = 53;
    break;
  case 35:
    if (isDSC)
      channel = 54;
    break;
  case 36:
    if (isDSC)
      channel = 55;
    break;
  case 37:
    if (isDSC)
      channel = 56;
    break;
  case 38:
    if (isDSC)
      channel = 51;
    else if (isPMS)
      channel = 58;
    else
      channel = 50;
    break;
  case 39:
    if (isPMS)
      channel = 59;
    else
      channel = -1;
    break;
  case 40:
    if (isDSC)
      channel = 57;
    else
      channel = 56;
    break;
  case 41:
    if (isDSC)
      channel = 58;
    else
      channel = 57;
    break;
  case 42:
    if (isPMS)
      channel = 56;
    break;
  case 43:
    if (isPMS)
      channel = 57;
    break;
  case 44:
    if (isPMS)
      channel = 58;
    break;
  case 45:
    if (isPMS)
      channel = 59;
    break;
  case 46:
    if (isDSC)
      channel = 59;
    else
      channel = 60;
    break;
  case 47:
    channel = 66;
    break;
  case 48:
    channel = 67;
    break;
  case 49:
    if (isDSC)
      channel = 131;
    else if (isPMS)
      channel = 138;
    else
      channel = 130;
    break;
  case 50:
    if (isPMS)
      channel = 139;
    else
      channel = -1;
    break;
  case 51:
    if (isDSC)
      channel = 137;
    else
      channel = 136;
    break;
  case 52:
    if (isDSC)
      channel = 138;
    else
      channel = 137;
    break;
  case 53:
    if (isPMS)
      channel = 136;
    break;
  case 54:
    if (isPMS)
      channel = 137;
    break;
  case 55:
    if (isPMS)
      channel = 138;
    break;
  case 56:
    if (isPMS)
      channel = 139;
    break;
  case 57:
    if (isDSC)
      channel = 139;
    else
      channel = 140;
    break;
  case 58:
    channel = 146;
    break;
  case 59:
    channel = 147;
    break;
  case 60:
    break;
  default:
    abort();
  }
  return channel;
}

static void dump(bool isPMS, bool isDSC) {
  std::println("static int convert{}{}(int channel) {{", isPMS ? "PMS" : "",
               isDSC ? "DSC" : "");
  std::println("  switch(channel) {{");
  for (int channel = 0; channel < 150; ++channel) {
    auto converted = doConvert(channel, isPMS, isDSC);
    if (converted != channel) {
      if (converted == -1)
        std::println("    case {:#x}: return -1;", channel);
      else
        std::println("    case {:#x}: return {:#x};", channel, converted);
    }
  }
  std::println("    default: return -2;");
  std::println("  }}");
  std::println("}}");
}

int main(int argc, char *argv[]) {
  dump(false, false);
  dump(false, true);
  dump(true, false);
  dump(true, true);
  return 0;
}
```

ez2patternBMS:
> @ez2patternBMS how did come up with this mapping? [8b1190b](https://github.com/GOMazk/OpenLR2/commit/8b1190b99852b5ea292c61eb715cf529569c4da4) I'm trying to replicate these changes to understand them

~~At FP/DSC, a part or mine channel are shift to left, so PMS_6(dsc key 5) never handle mines, and PMS_1(dsc key l-sc) handle mines from two channel.~~

Sorry, misread

ez2patternBMS:
> For reference, here is how I replicated the conversion By the way, you didn't handle the case where both isPMS and isDSC are true
> ```
Is it possible to enable both flag at the same time? let me check

ez2patternBMS:
Yes, they can both flaged in same time, I've never though that becase it's seems never indended.

ez2patternBMS:
> @ez2patternBMS how did come up with this mapping? [8b1190b](https://github.com/GOMazk/OpenLR2/commit/8b1190b99852b5ea292c61eb715cf529569c4da4)
> I'm trying to replicate these changes to understand them

I just wrote them hand by hand.

ez2patternBMS:
> ```c++
> if (converted != channel) {
>      ...
>     }
> ```

This can optimize switch-case, but makes code non-intuitive.

chown2:
It optimizes the readibility for the weird, more important case, so it's better IMO. When there is a mix of numbers on the left and named constants on the right, you don't know which of those is important and which isn't, they all look the same.

ez2patternBMS:
> It optimizes the readibility for the weird, more important case, so it's better IMO. When there is a mix of numbers on the left and named constants on the right, you don't know which of those is important and which isn't, they all look the same.

I think still readibility is low...  I'm planning removing non-note cases[00-09,70-C9] at next fix and refactor, so why don't we optimize this then?

https://github.com/GOMazk/OpenLR2/blob/f459e9f28285512cd08f009fbc289e162295f785/LR2/LR2_bmsload.cpp#L2163

~~BTW, just don't use hexdemical, they are intended to be demical~~
I'm now using andoird phone for review, so it's can be slow

chown2:
For now with no named constants, I just moved it into a proper switch 6ea9e28b5d292a21d3439a0f2569aa47321b395c

> I'm now using andoird phone for review, so it's can be slow

I'm still working on this so don't review it too deply yet :D

chown2 (commented):
Good to review now. Very briefly tested. Will approve after more testing tomorrow.

chown2:
> I think still readibility is low... I'm planning removing non-note cases[00-09,70-C9] at next fix and refactor, so why don't we optimize this then?

If you mean changing the way 'channel' vaue is converted, I'm against it, as we would have to touch every place using it and we may miss some

ez2patternBMS:
> > I think still readibility is low... I'm planning removing non-note cases[00-09,70-C9] at next fix and refactor, so why don't we optimize this then?
> 
> If you mean changing the way 'channel' vaue is converted, I'm against it, as we would have to touch every place using it and we may miss some

I mean, just pass though.
As you can see on my original array code,  channel of range [01-09,70-C9] are have no change of channel value, so just pass through to optimize readibility, and make array(or case-switch) shorter 

But, code is now changed to no named constant, so now meaningless. If there;s any plan to rewrite in kind of my style in future, that would be useful.

> Good to review now. Very briefly tested. Will approve after more testing tomorrow.

Let me add some comment, and change `HS_FIX` to `HSFIX`. I noticed it today, and I have no idea why I named like that.
Except this,  nothing will be commited.

chown2:
> Is there any plan to rewrite in kind of my style in future?

Yes! But let's not get ahead of ourselves and do one thing at the time. The transformation of jumptable + switch case to a plain switch case is already big enough of a change.

ez2patternBMS:
https://github.com/GOMazk/OpenLR2/pull/176/commits/c6b8bf9f2ad425f98379c1e9c2798c9580a40cdf: Add TOFIX comment
https://github.com/GOMazk/OpenLR2/pull/176/commits/dde4317298530d85557b20b0658dbb3dd84d5fc3: Change `OPTION_HS_FIX_xxx` to `OPTION_HSFIX_xxx`

ready for review

GOMazk (approved):

chown2 (approved):
Didn't have a chance to deeply test this, but I'm not goign to block this any further. Code-wise LGTM!  💨 🚀 💨 🚀

PR #186, opened 2026-06-29
[Bugfix] Fix FMOD initialization order

Unengine:
# Background
Audio drivers have been **replaced** after initializing fmod system.
For example, at first the driver is set to default one, and after initializing, it is replaced into the intended one.
In my case, it was first set to ASIO4ALL (probably default ASIO driver), and then replaced into the intended driver.
It should be initialized after setting parameters to prevent overhead and potential deadlock possibilities.
In short, I fixed the audio initialization sequence.

# Changes
Fixed fmod init sequence like this :
1. FMOD_System_Create
2. FMOD_System_SetDSPBufferSize
3.  FMOD_System_SetSoftwareChannels
4.  FMOD_System_SetOutput 
5. FMOD_System_SetDriver 
6. FMOD_System_Init


Unengine:
I made a mistake, I should have pushed the branch into my cloned repo. Closing it.

PR #187, opened 2026-06-29
[Bugfix] Fix FMOD initialization order

Unengine:
# Background
Audio drivers have been **replaced** after initializing fmod system.
For example, at first the driver is set to default one, and after initializing, it is replaced into the intended one.
In my case, it was first set to ASIO4ALL (probably default ASIO driver), and then replaced into the intended driver.
It should be initialized after setting parameters to prevent overhead and potential deadlock possibilities.
In short, I fixed the audio initialization sequence.

# Changes
Fixed fmod init sequence like this :
1. FMOD_System_Create
2. FMOD_System_SetDSPBufferSize
3.  FMOD_System_SetSoftwareChannels
4.  FMOD_System_SetOutput 
5. FMOD_System_SetDriver 
6. FMOD_System_Init

### DIFF
--- a/LR2/En_audio.cpp
+++ b/LR2/En_audio.cpp
@@ -1089,11 +1089,6 @@ int InitSound(AUDIO *aud, uint bufferLength, int numBuffer, char fDisable, int o
 		FMOD_System_SetDSPBufferSize(aud->fmodSys, bufferLength, numBuffer);
 		ErrorLogAdd("\n");
 		FMOD_System_SetSoftwareChannels(aud->fmodSys, 0x100);
-		if (FMOD_System_Init(aud->fmodSys, 0x100, FMOD_INIT_NORMAL, NULL) != FMOD_OK) {
-			ErrorLogAdd("FMOD_System_Init failed!\n");
-			EndSound(aud);
-			return 1;
-		}
 
 #ifdef _WIN32
 		if ([&] {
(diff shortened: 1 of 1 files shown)


MatVeiQaaa:
what problem does this solve? As in, what effect the previous code had?

chown2:
NOTE: d6b22d1c850eabc0fea5fbce410570af81ff899b is when FMOD_System_Init got moved to be before FMOD_System_SetDriver

Unengine:
> what problem does this solve? As in, what effect the previous code had?

Replacing audio driver after a FMOD_SYSTEM is initialized causes overhead.
Setting it before initialization is more safer.
And some audio drivers like ASIO4ALL notifies that the driver is loaded, and it clearly hurts the user experience.
At first, I thought the client ignored my driver settings because of the ASIO4ALL notification, but in result it was actually set to default driver(ASIO4ALL) at first, and set again to the intended driver.

chown2 (approved):
FMOD documentation reference: https://www.fmod.com/docs/2.03/api/core-api-system.html#system_setdriver

> If this function is called after [System::init](https://www.fmod.com/docs/2.03/api/core-api-system.html#system_init), the current driver will be shutdown and the newly selected driver will be initialized / started.

MatVeiQaaa (approved):

GOMazk (approved):

PR #193, opened 2026-07-01
Bug hunt

### LINKED ISSUE #146: GAS is not disabled for ghost autoplay
CC @MatVeiQaaa 
NOTE: Ghost autoplay is basically IR replay. Open the leaderboards with button 4 and press 'autostart', it plays out the ghost.

### LINKED ISSUE #182: Courses disappeared
ST, SL and GENOSIDE 2018 courses disappeared, but PMS courses remained. They don't show up as undefined, they just don't exist. This happened after F8 on song folder and trying to create a new course in the NONSTOP tab, although this may not be true since I didn't check my courses before these steps. Also tried to install new GENOSIDE 2018 courses through Jukebox1, nothing was added, everything remained the same. And yea, I also tried to change the keymode/

### LINKED ISSUE #192: UTF-8 support is completely broken on WIN7 OS
All tests were base on the latest version (v260627)

UTF-8 support was perfect on WIN10/11, but completely broken in WIN7. For example, the first run caution window:

Image

<br>
In WIN10/11 (Simplified Chinese Locale), run the x64/x86 execution directly, the window text was correct. And everything in game was perfect.
<br>
<br>
Image
<br>
But in WIN7 (Simplified Chinese Locale), run the x64/x86 execution directly, the window text was in messy codes. 
<br>
<br>
Image
<br>
Even run the x86 version and change to Japanese Locale with Locale Emulator, still in (another kind of) messy codes. 

And in-game there're many issues like the original LR2 with Non-Japanese Locale, like No sound, Missing charts, ETC. Also, I'm afraid the issue was in any WIN7 locale, including the Japanese Locale.

I know WIN7 is an ancient OS, but OpenLR2 runs perfectly except the UTF-8 support issue, really hoping there is a way to solve the problem, thanks very much!!

chown2:
Fixes https://github.com/GOMazk/OpenLR2/issues/182 "Courses disappeared". [(RE issue)](https://github.com/GOMazk/OpenLR2/commit/13baf525490e62b260ed3b80369c6850fdf23472). In Ghidra the code looks the way we have it in 'main', but it clearly doesn't match reality. Maybe the code actually had 'keys' as an unsigned int and the compiler abused that?
Fixes https://github.com/GOMazk/OpenLR2/issues/146 "GAS is not disabled for ghost autoplay".
Fixes https://github.com/GOMazk/OpenLR2/issues/192 "UTF-8 support is completely broken on WIN7 OS".

### DIFF
--- a/LR2/LR2_statplay.cpp
+++ b/LR2/LR2_statplay.cpp
@@ -251,6 +251,7 @@ int FlipScore(game *g){
 // Also see PerformGAS
 static int GetBestClearedGauge(const gameplay& gameplay, int playerIdx, const CONFIG_PLAY& cfg, bool limitToCourse) {
 	const PLAYERSTATUS& player = gameplay.player[playerIdx];
+	if (gameplay.isAutoplay) return player.gaugeType;
 	if (playerIdx == 1 && gameplay.ghostBattle) return player.gaugeType;
 	if (cfg.gaugeOption[playerIdx] == 5) return 5;
 	constexpr std::array<int, 5> gaugeArr({ 4, 2, 1, 0, 3 });
--- a/LR2/LR2_statplay.cpp
+++ b/LR2/LR2_statplay.cpp
@@ -248,8 +248,9 @@ int FlipScore(game *g){
 	return 1;
 }
 
-static int GetBestClearedGauge(gameplay& gameplay, int playerIdx, CONFIG_PLAY& cfg, bool limitToCourse) {
-	PLAYERSTATUS& player = gameplay.player[playerIdx];
+// Also see PerformGAS
+static int GetBestClearedGauge(const gameplay& gameplay, int playerIdx, const CONFIG_PLAY& cfg, bool limitToCourse) {
+	const PLAYERSTATUS& player = gameplay.player[playerIdx];
 	if (playerIdx == 1 && gameplay.ghostBattle) return player.gaugeType;
 	if (cfg.gaugeOption[playerIdx] == 5) return 5;
 	constexpr std::array<int, 5> gaugeArr({ 4, 2, 1, 0, 3 });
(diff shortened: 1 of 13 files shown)


LR2/LR2_statplay.cpp
```diff
@@ -251,6 +251,7 @@ int FlipScore(game *g){
 // Also see PerformGAS
 static int GetBestClearedGauge(const gameplay& gameplay, int playerIdx, const CONFIG_PLAY& cfg, bool limitToCourse) {
 	const PLAYERSTATUS& player = gameplay.player[playerIdx];
+	if (gameplay.isAutoplay) return player.gaugeType;
```
MatVeiQaaa, line 254: this is redundant, isAutoplay is already filtered out in SaveResult where GetBestClearedGauge is called.

chown2: Considering we eventually want to merge PerformGAS and GetBestClearedGauge together, it makes sense to keep their implementations in line with each other

MatVeiQaaa (approved):

LR2/Main.cpp
```diff
@@ -210,6 +226,11 @@ int main(int argc, char** argv) {
 #ifndef NDEBUG
 	while (!IsDebuggerPresent()) std::this_thread::sleep_for(std::chrono::milliseconds(200));
 #endif // NDEBUG
+
+	if (!IsWindowsVersionAbove1903()) {
+		MessageBoxA(nullptr, "Windows version is too old. Expected at least Windows 10 1903.", "エラー", 0);
+		return -1;
```
GOMazk: https://github.com/GOMazk/OpenLR2/issues/192#issuecomment-4861797696
what about omit "reutrn -1;"?

chown2:
I've removed the 'return -1' so the game can be started even if broken now.
Also snuck in a little refactor regarding ghosts in CustomIR. There is an issue of them being lost sometimes but I haven't dug any deeper yet.

MatVeiQaaa (approved):

PR #200, opened 2026-07-03
feat(customir): implement F5 with a dedicated method and make song result non-blocking

chown2:
See commit message for details.

### DIFF
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -5,6 +5,7 @@
 #include <fstream>
 #include <iostream>
 #include <print>
+#include <string>
 
 #ifdef _WIN32
 #include <windows.h>
@@ -13,7 +14,6 @@
 namespace State {
     static std::filesystem::path path;
     static int scoresSaved = 0;
-    static std::string webRankingTemplate = "https://example.com/{hash}";
 }
 
 static const char* OLR2_IR_API GetName() {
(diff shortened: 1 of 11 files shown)


chown2:
@luxurya111 the commit message explains why I had to break the API. On Stellaverse IR module side this can be implemented as easily as it is done in ExampleIR. It can co-exist with webRankingUrlTemplate too, although webRankingUrlTemplate is not used after this PR and you will get a compiler warning for it.

chown2:
Pushed more work into this PR to avoid merge conflicts

luxurya111:
I have checked the updated changes. I agree that this approach looks much cleaner and better.
I will test the non-blocking score transmission in that specific branch.
IMO it would be a good idea to reflect the changes to the transmission method in the PR title as well. (So that it remains visible when squashed.)

chown2:
Please don't squash my PRs, I always make my commits granular :open_mouth: :smiley:

luxurya111:
Ah, I'm not the one doing the squash. (I don't have merge permissions anyway.)
I apologize for causing any misunderstanding.

At any rate, I tested the changes on StellaverseIR as well, and everything goes through properly.

ExampleIR/ExternalIR.cpp
```diff
@@ -106,6 +109,10 @@ static openlr2::GetStatus OLR2_IR_API GetGhost(const char* songHash, openlr2::Gh
     return openlr2::GetStatus::Ok;
 }
 
+static std::string GetWebRankingUrl(const char* songHash) {
```
MatVeiQaaa: lacks OLR2_IR_API function type

MatVeiQaaa (approved):

Issue #158, opened 2026-06-24
RGB values of #DST doesn't work in certain situations


MellDa1024:
<img width="2564" height="754" alt="Image" src="https://github.com/user-attachments/assets/cf6d4db0-9b10-4fc1-9f60-f95c8278b4eb" />
Tested with LR2HD(F/S Patched) and [W-MIX Skin(OpenLR2 Patch)](https://wispworks-web.vercel.app/)

\* Both Client used OpenLR2 Patched W-MIX Skin.
I spent some time investigating this issue but could not find a clear cause.
Even with the same `DST_NUMBER`, the Level Number displayed in the Play Scene has the RGB values applied correctly (and a similar behavior is observed with `DST_BARGRAPH`).
Interestingly, changing the `blend` value to `0` or `2` sometimes makes the RGB values apply.
Changing other parts of the skin sometimes makes the RGB values apply too. (e.g., changing the F/S number's blend mode to 2 will make the F/S bargraph work properly.)

```csv
/B-AN
//SRC定義,(null),gr,x,y,w,h,div_x,div_y,cycle,timer,num,align,keta,,,,,,,
#SRC_NUMBER,0,0,957,57,180,13,10,1,0,0,49,2,2,,,,,,,
//DST定義,(NULL),time,x,y,w,h,acc,a,r,g,b,blend,filter,angle,center,loop,timer,op1,op2,op3
#DST_NUMBER,0,200,733,654,18,13,0,30,233,69,154,1,0,0,0,500,0,155,0,0
#DST_NUMBER,0,500,733,654,18,13,0,255,233,69,154,1,0,0,0,0,0,0,0,0
#DST_NUMBER,0,2000,733,654,18,13,0,135,233,69,154,1,0,0,0,0,0,0,0,0
#DST_NUMBER,0,3500,733,654,18,13,0,255,233,69,154,1,0,0,0,0,0,0,0,0
```
\* Level Number
```csv
//fast
//SRC定義,(null),gr,x,y,w,h,div_x,div_y,cycle,timer,num,align,keta,,,,,,,
#SRC_NUMBER,0,0,970,78,132,11,11,1,0,0,212,0,4,,,,,,,
//DST定義,(NULL),time,x,y,w,h,acc,a,r,g,b,blend,filter,angle,center,loop,timer,op1,op2,op3
#DST_NUMBER,0,700,58,565,12,11,0,0,85,141,224,0,0,0,0,900,0,0,0,0
[...]

Issue #59, opened 2026-01-28
NEWSONG highlight not shown or shown on wrong entries


chown2:

chown2:
The titleflash branch is hit correctly, so this is apparently something with time animation
https://github.com/GOMazk/OpenLR2/blob/f3a7f3509d82238fa39f14182689d66b470edc41/LR2/Scene02_Songselect.cpp#L2573-L2578

chown2:
DxLib update related, same thing was observed in an old version of UpscaLR2. By the way, the highlight may even be applied to only a part of a single text line, not necessarily binary on and off.
@MatVeiQaaa was this ever solved in UpscaLR2?

MatVeiQaaa:
> DxLib update related, same thing was observed in an old version of UpscaLR2. By the way, the highlight may even be applied to only a part of a single text line, not necessarily binary on and off. [@MatVeiQaaa](https://github.com/MatVeiQaaa) was this ever solved in UpscaLR2?

I never noticed it myself, so i didn't look into this. If it was solved, it was incidental.

PR #202, opened 2026-07-04
Shader compilation from source

MatVeiQaaa:
Compiles both DX9 and DX11 shaders from hardcoded plaintext source, instead of using DXA, as we got an outdated DXA version for last ver shader packages.

This has potential to fix some problems with DX9 render.

DX11 can now be used and won't default to DX9EX.

DX11 had two bugs which this PR addresses: 
 - Render had transparent background, which could be seen on screenshots and skin select preview.
 - In fullscreen mode, first and last columns of pixels were swapped. I am not entirely sure if this work around is adequate, we might be losing two pixels of information this way, but we might also not be. The bug certainly originates elsewhere.


chown2:
NOTE: now that shaders get enabled successfully, the picture is borked in DX9 mode when combined with DXVK 3.0 (but not 2.7). Works with vanilla Wine and on Windows however. It's unclear where they issue exactly is, and code wise it looks good, so it's fine by me.
Work-around which breaks NEWSONG highlight (https://github.com/GOMazk/OpenLR2/issues/59) and some minor color blending (https://github.com/GOMazk/OpenLR2/issues/158):
Put this into dxvk.conf:
```conf
; avoid broken picture
; breaks NEWSONG highlight https://github.com/GOMazk/OpenLR2/issues/59
; or = 1
d3d9.shaderModel = 0
```

LR2/Main.cpp
```diff
@@ -1996,6 +1990,15 @@ int main(int argc, char** argv) {
 				//ProcGame(&gs); //why this is here
 			}
 		}
+		{
+			int oldMode, oldParam;
+			GetDrawBlendMode(&oldMode, &oldParam);
+			SetDrawBlendMode(DX_BLENDMODE_ADD, 255);
+			SetDrawZ(1.);
+			DrawBox(0, 0, 640, 480, GetColor(0, 0, 0), 1);
```
chown2: With what I learned in #201, this should probably use the canvas size
```suggestion
			DrawBox(0, 0, resX, resY, GetColor(0, 0, 0), 1);
```

LR2/Scene07_Skinselect.cpp
```diff
@@ -275,6 +275,15 @@ int ProcI_SkinSelect(game *g) {
 	for (int i = 0; i < g->skstruct2.drBuf.count; i++) {
 		LRDraw(&g->skstruct2.drBuf, &g->txtStruct, &g->sSelect, &g->skstruct2, i, 0, 0);
 	}
+	{
+		int oldMode, oldParam;
+		GetDrawBlendMode(&oldMode, &oldParam);
+		SetDrawBlendMode(DX_BLENDMODE_ADD, 255);
+		SetDrawZ(1.);
+		DrawBox(0, 0, 640, 480, GetColor(0, 0, 0), 1);
```
chown2: Same as above
```suggestion
		DrawBox(0, 0, tx, ty, GetColor(0, 0, 0), 1);
```

dep/DxLib/Windows/DxGraphicsD3D11.cpp
```diff
@@ -1787,7 +1787,7 @@ extern int Graphics_D3D11_ShaderCode_Base_Initialize( void )
 		SCBASE->ShaderTxt["DxShader_DataType_D3D11.h"] = { ShaderTxt_DataTypeSubInclude_D3D11, ShaderTxt_DataTypeSubInclude_D3D11_len };
 		SCBASE->ShaderTxt["DxShader_PS_D3D11.h"] = { ShaderTxt_PSInclude_D3D11, ShaderTxt_PSInclude_D3D11_len };
 		SCBASE->ShaderTxt["DxShader_VS_D3D11.h"] = { ShaderTxt_VSInclude_D3D11, ShaderTxt_VSInclude_D3D11_len };
-		}
+	}
```
chown2, line 1790: This change was supposed to be in the [fix(dxlib): shader compilation, dx11](https://github.com/GOMazk/OpenLR2/pull/202/commits/2439b7d4ca2d53ee120f52f59cd3f2aa2a34df14) commit but ended up in [fix(dxlib): fullscreen edge pixels swapped](https://github.com/GOMazk/OpenLR2/pull/202/commits/4b84ca617300ecffbb1090ff2bf5e0e9d5cc279e)

chown2:
NOTE: SHIKI - Air doesn't work in DX11 mode. Randomly either freezes the game or crashes it within dx11.dll.

chown2 (approved):

MatVeiQaaa:
> NOTE: SHIKI - Air doesn't work in DX11 mode. Randomly either freezes the game or crashes it within dx11.dll.

The crash is caused by LoadBmsResource doing LoadGraph for bga async, while main thread is performing rendering. DX device state can't be altered asynchronously, this is UB, and causes our crash in DX11. LR2 bug.

PR #203, opened 2026-07-05
Refactor of this weekend;mostly boolean refactoring.

ez2patternBMS:
Most of them are boolean refactoring.
And include find missing gauge hardcoded value, and change some 1p/2p duplicated variable to use array.

NOTE: I know boolean to int can use Implicit casting, but I purposely put cast in to clarify that this is boolean.


chown2 (dismissed):
Please undo all changes to data types in LR2/LR2_customir_api.h to keep compatibility with existing CustomIR modules

chown2:
Avoiding raw loops `for (int i_and_p = PLAYER_1; i < PLAYER_MAX; i++) {` -> `for (int p : { PLAYER_1, PLAYER_ 2}) {`

ez2patternBMS:
> Please undo all changes to data types in LR2/LR2_customir_api.h to keep compatibility with existing CustomIR modules

https://github.com/ez2patternBMS/OpenLR2/compare/45967bd19519615b25a8e9b37dd484e84a3bc327...d228e39adc17ef7ece8fe2a4ccfa32a603ef2ff3
is it okay doing like this? (Don't mind commit message, they will be squashed)

> Avoiding raw loops `for (int i_and_p = PLAYER_1; i < PLAYER_MAX; i++) {` -> `for (int p : { PLAYER_1, PLAYER_ 2}) {`

Unlike other options, you think there won't be anything added on this side?
I know it sound strange to considering add PLAYER_3, but just checking.

chown2:
> is it okay doing like this? 

Yes, as long as LR2_customir_api.h‎ itself keeps the same data types, it's okay. Everything else can be changed without problem.

> Unlike other options, you think there won't be anything added on this side?
> I know it sound strange to considering add PLAYER_3, but just checking.

No such plans yet, but even if we ever want it, `PLAYER_1, PLAYER_2` is easily greppable.

chown2:
I noticed the last commit also makes LR2_customir_api.h‎ use PLAYER_MAX. Please, don't do that either. LR2_customir_api.h is meant to be independent of OpenLR2 code base as it's directly copied into IR modules.
Example: https://github.com/MatVeiQaaa/BokutachiIR_OpenLR2/blob/master/include/LR2_customir_api.h

chown2:
Now I noticed some fields got renamed in LR2_customir_api.h - please undo that for the same reason :)
We should do that eventually, but for now let's keep them the old way, just so we have some easier time in CustomIR modules.

ez2patternBMS:
> Now I noticed some fields got renamed in LR2_customir_api.h - please undo that for the same reason :) We should do that eventually, but for now let's keep them the old way, just so we have some easier time in CustomIR modules.

Fixed

LR2/LR2_skinload.cpp
```diff
@@ -1305,56 +1305,56 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 							tSkin_num += 2;
 						}
 					}
-					else if (fBuf.left(16).isSame("#SRC_NOWJUDGE_1P")) {
```
chown2: whoops

ez2patternBMS: Fixed

LR2/LR2_configsave.cpp
```diff
@@ -484,7 +484,7 @@ int WriteConfigXml(game *g, const char *filename){
 	WriteXML_Tab2Str(pFile, "play_5_b", (g->config).skin.skinFilePath[13]);
 	WriteXML_Tab2Str(pFile, "play_9_b", (g->config).skin.skinFilePath[14]);
 	WriteXML_Tab2Str(pFile, "fontname", (g->config).skin.fontname);
-	sprintf(buf, "\t\t<%s>%d</%s>\n", "disableimagefont", (g->config).skin.disableimagefont, "disableimagefont");
+	sprintf(buf, "\t\t<%s>%d</%s>\n", "disableImageFont", (int)(g->config).skin.disableImageFont, "disableImageFont");
```
MatVeiQaaa: don't rename the config entry, especially in original LR2 config file. (I imagine just find-replace mistake)

LR2/LR2_configsave.cpp
```diff
@@ -1125,7 +1125,7 @@ int ReadConfig(game* g, const char* filepath) {
 	ReadXml_Str("config", "skin", "play_5_b", "", &g->config.skin.skinFilePath[13], hXml);
 	ReadXml_Str("config", "skin", "play_9_b", "", &g->config.skin.skinFilePath[14], hXml);
 	ReadXml_Str("config", "skin", "fontname", "Ariel", &g->config.skin.fontname, hXml);
-	ReadXml_Int("config", "skin", "disableimagefont", 0, &g->config.skin.disableimagefont, hXml);
+	ReadXml_PositiveIntAsBool("config", "skin", "disableImageFont", false, &g->config.skin.disableImageFont, hXml);
```
MatVeiQaaa: same here

LR2/LR2_skinload.cpp
```diff
@@ -1305,56 +1305,56 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
...
+					else if (fBuf.left(16).isSame("#src_NOWJUDGE_1P")) {
```
MatVeiQaaa: still "whoops". Should be all capital letters. I think isSame() is case-sensitive?

LR2/LR2_skinload.cpp
```diff
@@ -1369,28 +1369,28 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 							ReadDST(&sk->dst_GROOVEGAUGE[csv.val[1]], &csv, tSkin_num);
 						}
 					}
-					else if (fBuf.left(18).isSame("#SRC_GAUGECHART_1P")) {
+					else if (fBuf.left(18).isSame("#src_GAUGECHART[PLAYER_1]")) {
```
MatVeiQaaa: original "whoops" + capitalize

LR2/LR2_skinload.cpp
chown2: isSame() is case-sensitive

LR2/LR2_skinobject.cpp
```diff
@@ -4344,7 +4352,7 @@ int ReadOptionstrFile(OptionString *arrOpStr, CSTR filepath) {
 			else if (fBuf.left(8).isSame("#BGADRAW")) {
 				ReadOptionstr(&arrOpStr[9], csv);
 			}
-			else if (fBuf.left(11).isSame("#SCOREGRAPH")) {
+			else if (fBuf.left(11).isSame("#scoreGraph")) {
```
MatVeiQaaa: capitalize

LR2/LR2_statplay.cpp
```diff
@@ -199,8 +199,8 @@ int FlipScore(game *g){
 static int GetBestClearedGauge(const gameplay& gameplay, int playerIdx, const CONFIG_PLAY& cfg, bool limitToCourse) {
 	const PLAYERSTATUS& player = gameplay.player[playerIdx];
 	if (gameplay.isAutoplay) return player.gaugeType;
-	if (playerIdx == 1 && gameplay.ghostBattle) return player.gaugeType;
-	if (cfg.gaugeOption[playerIdx] == 5) return 5;
+	if (playerIdx == PLAYER_1 && gameplay.ghostBattle) return player.gaugeType;
+	if (cfg.gaugeType[playerIdx] == 5) return 5;
```
MatVeiQaaa: GAS methods should have gauge names too. 5 -> OPTION_GAUGE_GATTACK, etc

LR2/LR2_statplay.cpp
```diff
@@ -734,17 +737,17 @@ int SaveResult(game *g, sqlite3* sql) {
...
-				g->gameplay.player[1].clearGaugeTypeCourse = g->gameplay.player[1].gaugeType;
+			if (g->gameplay.courseStageNow == 0 || is_gauge_better(g->gameplay.player[PLAYER_2].clearGaugeTypeCourse, g->gameplay.player[PLAYER_2].gaugeType)) {
+				g->gameplay.player[PLAYER_2].clearGaugeTypeCourse = g->gameplay.player[PLAYER_2].gaugeType;
 			}
-			CheckClear(&g->gameplay.player[1], g->gameplay.player[1].gaugeType, g->gameplay.isCourse);
+			CheckClear(&g->gameplay.player[PLAYER_2], g->gameplay.player[PLAYER_2].gaugeType, g->gameplay.isCourse);
 		}
 		else {
-			g->gameplay.player[1].clearType = g->gameplay.player[0].clearType;
+			g->gameplay.player[PLAYER_2].clearType = g->gameplay.player[PLAYER_1].clearType;
```
MatVeiQaaa, line 760: clearType could also use named values. Note that they are different from gauge types.

LR2/Main.cpp
```diff
@@ -350,23 +350,23 @@ int main(int argc, char** argv) {
...
 			gs.is_recordmode = 1;
-			gs.config.select.preview = 0;
+			gs.config.select.showPreview = false;
 		}
 		else if (tStr2.isSame("-bga2avi")) {
 			gs.rec.recMode = 3;
 			gs.skstruct.drBuf.isDisabled = 1;
 			gs.is_recordmode = 1;
-			gs.config.select.preview = 0;
+			gs.config.select.showPreview = false;
```
MatVeiQaaa: technically not "show" but "play". Tbh just "preview" was fine. By the convention you set before it would be isPreview, i guess.

LR2/Scene02_Songselect.cpp
```diff
@@ -401,22 +401,23 @@ int Print_ManiacOptions(game *g) {
 				break;
 			case 20:
 				printfDx("LUNARIS             ");
-				if (g->config.play.m_lunaris == 0) printfDx("OFF");
-				if (g->config.play.m_lunaris == 1) printfDx("ON");
+				if (g->config.play.m_isLunaris) 	printfDx("ON");
```
MatVeiQaaa: g->config.play.m_isLunaris ? printfDx("ON") : printfDx("OFF");
Maybe?

LR2/Scene04_Play.cpp
```diff
@@ -1812,18 +1813,20 @@ int ProcS_Play(game *g, sqlite3* sql) {
...
 		}
 		else {
-			int origGauge = g->config.play.gaugeOption[0];
+			int origGauge = g->config.play.gaugeType[PLAYER_1];
 			//TOFIX : seed is not putted into replaydata, when use ghostbattle. (retry puts seed) (see also ParseBmsFile())
-			if(g->sSelect.bmsList[g->sSelect.cur_song].keymode > 9)
-				g->net.GetTargetInfo(0, md5, &gData, &gName, &g->config.play.gaugeOption[0], &g->config.play.random[0], &g->config.play.random[1], &g->config.play.dpflip, &g->gameplay.randomseed, &iTemp);
+			if(g->sSelect.bmsList[g->sSelect.cur_song].keymode > 9) {
+				g->net.GetTargetInfo(0, md5, &gData, &gName, &g->config.play.gaugeType[PLAYER_1], &g->config.play.random[PLAYER_1], &g->config.play.random[PLAYER_2], &iDpFlipTemp, &g->gameplay.randomseed, &iTemp);
+				g->config.play.dpFlip = (bool)iDpFlipTemp;//bool hack
```
MatVeiQaaa: can you explain it? We are setting a bool by casting an uninit int to a bool? This will evaluate "true" in debug build, and "false" in release (given we are in msvc), if i understand it correctly.

LR2/Scene02_Songselect.cpp
chown2: Better do `printfDx(g->config.play.m_isLunaris ? "ON" : "OFF");`, but imo it's best to keep this PR minimal

ez2patternBMS: That's better

LR2/LR2_skinload.cpp
ez2patternBMS: yes all capitalize is oops, I'll fix all of that

LR2/LR2_statplay.cpp
ez2patternBMS: (Note for me) why playerIdx == 1 is PLAYER_1 instead PLAYER 2

LR2/Scene04_Play.cpp
chown2: It's passed to GetTargetInfo which is then supposed to fill it.
But
1) it might not, so the original variable must be initialized
2) the check should be `> 0` instead of casting to bool. Also note, you used a C-style cast here. Never use them in C++, because they silently fall back to reinterpret_cast, which is almost always erroneous.

chown2: Also, move the variable declaration closer to the scope it's used in. The rest of the function shouldn't know about it.
And add `// TODO: make GetTargetInfo take a bool for dp flip`.

LR2/LR2_statplay.cpp
ez2patternBMS: I know that the clearType is different from the gaugeType. Is this a notice about it?
I don't understand, so I'm asking again.

LR2/Main.cpp
ez2patternBMS: Note for me: fix to isPreview

LR2/Scene02_Songselect.cpp
ez2patternBMS: Two option are both more optimized, but I think current one is still readable and basic.

chown2: I'd keep explicit if and else. But it's up to you to decide.

LR2/Scene04_Play.cpp
ez2patternBMS: > can you explain it? We are setting a bool by casting an uninit int to a bool? This will evaluate "true" in debug build, and "false" in release (given we are in msvc), if i understand it correctly.

I'm using VSCode and VSCode shows me red line. (maybe redline is wrong)

```C++
typedef unsigned int    ImageBaseOffset32;
//typedef unsigned char    bool;
typedef unsigned char    byte;
```

is it really bool is uint casted to bool? Or not in structure.h?

ez2patternBMS: > It's passed to GetTargetInfo which is then supposed to fill it. But
> 
> 1. it might not, so the original variable must be initialized
> 2. the check should be `> 0` instead of casting to bool. Also note, you used a C-style cast here. Never use them in C++, because they silently fall back to reinterpret_cast, which is almost always erroneous.

Then is it okay to replace all `(bool)` to `static_cast<bool>(i)` or use Ternary Operator?

LR2/LR2_statplay.cpp
MatVeiQaaa: > (Note for me) why playerIdx == 1 is PLAYER_1 instead PLAYER 2

ah good eye! I tried paying attention to those but its very difficult over such quantity.

LR2/LR2_statplay.cpp
MatVeiQaaa: > I know that the clearType is different from the gaugeType. Is this a notice about it? I don't understand, so I'm asking again.

I'm just saying that it should be named too. Note about difference with gauge type is just in case, because i think it can be not obvious.

LR2/Scene04_Play.cpp
chown2:
```
			if(g->sSelect.bmsList[g->sSelect.cur_song].keymode > 9) {
				g->net.GetTargetInfo(0, md5, &gData, &gName, &g->config.play.gaugeType[PLAYER_1], &g->config.play.random[PLAYER_1], &g->config.play.random[PLAYER_2], &iDpFlipTemp, &g->gameplay.randomseed, &iTemp);
				g->config.play.dpFlip = (bool)iDpFlipTemp;//bool hack
```
->
```
			if(g->sSelect.bmsList[g->sSelect.cur_song].keymode > 9) {
				int iDpFlipTemp{};
				// TODO: make GetTargetInfo take a bool for dp flip
				g->net.GetTargetInfo(0, md5, &gData, &gName, &g->config.play.gaugeType[PLAYER_1], &g->config.play.random[PLAYER_1], &g->config.play.random[PLAYER_2], &iDpFlipTemp, &g->gameplay.randomseed, &iTemp);
				g->config.play.dpFlip = iDpFlipTemp > 0;
```

LR2/LR2_configsave.cpp
ez2patternBMS: fixed

LR2/LR2_statplay.cpp
ez2patternBMS: fixed both

LR2/Scene04_Play.cpp
MatVeiQaaa: > ```c++
> typedef unsigned int    ImageBaseOffset32;
> //typedef unsigned char    bool;
> typedef unsigned char    byte;
> ```
> 
> is it really bool is uint casted to bool? Or not in structure.h?

I dont exactly understand the question, but in case you are asking what those are: its a remnant of this project being reverse engineered compiled code. The size of "bool" is not necessarily one byte - its compiler dependent -, and from compiled code its impossible to tell if its really a bool. So saying its "char" is guaranteed to not be wrong when reverse engineering. We don't need this anymore.

ez2patternBMS: > I dont exactly understand the question, but in case you are asking what those are: its a remnant of this project being reverse engineered compiled code. The size of "bool" is not necessarily one byte - its compiler dependent -, and from compiled code its impossible to tell if its really a bool. So saying its "char" is guaranteed to not be wrong when reverse engineering. We don't need this anymore.

I understand that you are talking about just defining temporary variable. My misread, and will be fixed to `iDpFlipTemp > 0`

LR2/LR2_statplay.cpp
ez2patternBMS: Yes, I'm also going to name it later,
But, I'm going to focus on boolean mostly.

LR2/Scene04_Play.cpp
ez2patternBMS: Done for all `(bool)`

LR2/Scene02_Songselect.cpp
ez2patternBMS: just use if-case, don't need more optimizing for this

note: tab fix added

LR2/structure.h
```diff
@@ -350,43 +356,39 @@ struct CONFIG_JUKEBOX {
...
-	int p2_assist{};
-	int dpflip{};
+	int hiSpeed[PLAYER_MAX]{};
+	int gaugeType[PLAYER_MAX]{}; /* 0:groove 1:survival 2:death 3:easy 4:pattack 5:gattack */
+	int random[PLAYER_MAX]{}; // 1:mirror 2:random 3:s-random 4:scatter 5:converge
+	int m_HIDSUD[PLAYER_MAX]{};
+	int lanecoverv[PLAYER_MAX]{};
+	int lanecover[PLAYER_MAX]{};
+	bool autokey{};
+	bool assist[PLAYER_MAX]{};
```
chown2: Please, make 'assist' back into an int. There is no 'assist' toggle, there's only auto-scratch. There is no need to pessimize extendability of 'assist' as we might expand it in the future.

LR2/LR2_statplay.cpp
```diff
@@ -200,26 +200,26 @@ static int GetBestClearedGauge(const gameplay& gameplay, int playerIdx, const CO
 	const PLAYERSTATUS& player = gameplay.player[playerIdx];
 	if (gameplay.isAutoplay) return player.gaugeType;
 	if (playerIdx == PLAYER_2 && gameplay.ghostBattle) return player.gaugeType;
-	if (cfg.gaugeType[playerIdx] == 5) return 5;
-	constexpr std::array<int, 5> gaugeArr({ 4, 2, 1, 0, 3 });
+	if (cfg.gaugeType[playerIdx] == OPTION_GAUGE_GATTACK) return OPTION_GAUGE_GATTACK;
+	constexpr std::array<int, OPTION_GAUGE_END> gaugeArr({ OPTION_GAUGE_PATTACK, OPTION_GAUGE_DEATH, OPTION_GAUGE_HARD, OPTION_GAUGE_GROOVE, OPTION_GAUGE_EASY });
```
chown2: OPTION_GAUGE_END makes no sense here as it has no relation to array sizes. In fact, this array doesn't contain GATTACK.
If you really want it, this can be `static constexpr auto gaugeArr = std::to_array({OPTION_GAUGE_PATTACK, OPTION_GAUGE_DEATH, OPTION_GAUGE_HARD, OPTION_GAUGE_GROOVE, OPTION_GAUGE_EASY})`.

LR2/LR2_customir.cpp
```diff
@@ -454,12 +458,12 @@ void IRScoreInternal::MakeScoreV1(IRScoreV1& scoreOut) const {
 	scoreOut.settings.m_wave = settings.m_wave;
 	scoreOut.settings.m_spiral = settings.m_spiral;
 	scoreOut.settings.m_sidejump = settings.m_sidejump;
-	scoreOut.settings.is_extra = settings.is_extra;
+	scoreOut.settings.is_extra = (int)settings.m_isExtra;
```
chown2, line 456: No need to change anything here, but in future use 'static_cast<int>(...)` instead (explain in a comment above)

LR2/structure.h
ez2patternBMS: oh, iidx has 'legacy note' option as assite, okay i turn it back to int

```
int CheckMission(game *g){
	int level;
	int gauge;

	if (g->config.play.battle) 
		return 0;

	if (g->config.play.assist[PLAYER_1]) 
		return 0;
	/*
	//TOFIX: assits is bool variable, why 2?
	if (g->config.play.assist[PLAYER_1] == 2)
		return 0;
	*/

```

Talking about that, should I uncomment this?
assist == 2 is only exist at here.

LR2/LR2_customir.cpp
ez2patternBMS: > NOTE: I know boolean to int can use Implicit casting, but I purposely put cast in to clarify that this is boolean.

LR2/LR2_customir_api.h
```diff
@@ -13,6 +13,7 @@
 #define OLR2_IR_EXPORT
 #endif // _WIN32
 
+//NOTE: Do not change this structrue, should be done at once on V2
```
chown2: Not exactly true. We should add a note on top of this file saying something like "when making a CustomIR module, you should copy this file into your project. On any updates, we will preserve ABI compatilibty". Let me think how to phrase it better...

LR2/LR2_customir.cpp
ez2patternBMS: And isn't it safe to use C style cast for bool to int? caz it actually do nothing.

LR2/structure.h
chown2:
```
	if (g->config.play.assist[PLAYER_1]) 
		return 0;
	/*
	//TOFIX: assits is bool variable, why 2?
	if (g->config.play.assist[PLAYER_1] == 2)
		return 0;
	*/
```
->
```
	if (g->config.play.assist[PLAYER_1] != 0) 
		return 0;
```

LR2/LR2_statplay.cpp
ez2patternBMS: will be fixed

LR2/LR2_customir_api.h
ez2patternBMS: Then, I'm going to drop this commit

LR2/structure.h
ez2patternBMS: I mean that was original code, btw then i will remove that.

LR2/LR2_customir.cpp
chown2: It's never safe to use C style casts because they can implicitly fall back to reinterpret_cast if you change the type of settings.m_isExtra. https://www.youtube.com/watch?v=DAvZ3OG9cNo

LR2/LR2_customir_api.h
ez2patternBMS: droped

LR2/LR2_customir.cpp
ez2patternBMS: I understand. I'll fix that later. (it's better to use only one style)

LR2/structure.h
```diff
@@ -968,32 +968,26 @@ struct skstruct {
...
+	struct SRCstruct src_GAUGECHART[PLAYER_MAX][2]{};
+	struct DSTstruct dst_GAUGECHART[PLAYER_MAX][2]{};
 	struct SRCstruct src_SCORECHART[3]{};
 	struct DSTstruct dst_SCORECHART[3]{};
 	struct SRCstruct src_THUMBNAIL {};
 	struct DSTstruct dst_THUMBNAIL {};
-	struct SRCstruct src_README[2]{};
-	struct DSTstruct dst_README[2]{};
+	struct SRCstruct src_README[PLAYER_MAX]{};//TOFIX?: why 2P readme?
+	struct DSTstruct dst_README[PLAYER_MAX]{};
```
chown2: Doesn't seem related to P1/P2. I'd revert this for now.
https://github.com/GOMazk/OpenLR2/blob/b8538fad697a1d300ff51e6932ada1be5b0583e6/LR2/LR2_skinload.cpp#L1488-L1499

LR2/LR2_customir.cpp
chown2: Don't waste any time on this, just keep it as it is. clang-tidy can do this automatically later.

LR2/structure.h
```diff
@@ -968,31 +968,25 @@ struct skstruct {
...
-	struct SRCstruct src_GAUGECHART_2P[2]{};
-	struct DSTstruct dst_GAUGECHART_2P[2]{};
+	struct SRCstruct src_GAUGECHART[PLAYER_MAX][2]{};
+	struct DSTstruct dst_GAUGECHART[PLAYER_MAX][2]{};
 	struct SRCstruct src_SCORECHART[3]{};
 	struct DSTstruct dst_SCORECHART[3]{};
 	struct SRCstruct src_THUMBNAIL {};
 	struct DSTstruct dst_THUMBNAIL {};
-	struct SRCstruct src_README[2]{};
+	struct SRCstruct src_README[2]{};//TOFIX?: why 2P readme?
```
```suggestion
	struct SRCstruct src_README[2]{};
```
Comment still there

LR2/structure.h
```diff
@@ -350,43 +356,39 @@ struct CONFIG_JUKEBOX {
 };
 
 struct CONFIG_NETWORK {
-	int lr1ir{};
-	CSTR lr1id;
-	CSTR lr1pass;
+	int lr1ir{};//TOFIX: unfinished implement?
```
chown2: No need for any of these 'TOFIX', and for 'randomclose' below. These are probably legacy options carried from the original Lunatic Rave. They have to be kept so that they are not removed from config files.

ez2patternBMS: actually, lr1 api is different.
for example, lr1 used `http://www.dream-pro.info/~lavalse/cgi-bin/login.cgi` instead `http://www.dream-pro.info/~lavalse/LR2IR/2/login.cgi` so there's no need to lr2ir api accept lr1id.

So it's not related to auth, I think this is dummy variable from LR1 source code or, just maybe some lr1 and lr2 linking system planed?

ez2patternBMS: Oh I misread again, I don't know why I misread these days a lot

anyway i'm going to remove all tofix and just comment as unimplemented(or unused)as you said

chown2: Maybe. But nothing you have said is related to my comment. These TOFIX will never be acted upon.

ez2patternBMS: By the way, I'm thinking that remove them from structure and only leave xml writing part (just generate line as default, and never read)

Is it bad idea?

> Maybe. But nothing you have said is related to my comment. These TOFIX will never be acted upon.

sorry

chown2: > By the way, I'm thinking that remove them from structure and only leave xml writing part (just generate line as default, and never read)
> Is it bad idea?

Don't do that for now, maybe we can come up with a better way to do it later, like storing excessive fields in a single std::map

> sorry

No problem :) Looks like we sent messages at the same time, so it could've been confusing

LR2/Scene07_Skinselect.cpp
```diff
@@ -165,21 +165,21 @@ int ProcI_SkinSelect(game *g) {
 	}
 
 	if (g->skinData.select <= 4 || g->skinData.select == 12 || g->skinData.select == 13) {
-		for (int i = 0; i < 2; i++) {
-			if (g->gameplay.isCourse == 0 && (g->config.play.gaugeOption[i] == OPTION_GAUGE_GROOVE || g->config.play.gaugeOption[i] == OPTION_GAUGE_EASY)) 
-				AddDrawingBuffer_Gauge(&g->skstruct2.drBuf, &g->skstruct2.src_GROOVEGAUGE[i], &g->skstruct2.dst_GROOVEGAUGE[i], &g->timer2, (int)g->gameplay.player[i].HP_print / 2, 0);
+		for (int p = PLAYER_1; p < PLAYER_MAX; p++) {
```
chown2: Accidental change https://github.com/GOMazk/OpenLR2/pull/203/changes/0b0d1a662294f4d732f45d9b72952a07133c468b#diff-decf19fb112fb42a6676469be325fa13261f30ee8b2a2d0e767b098b8320cb59R168

LR2/LR2_replay.cpp
```diff
@@ -186,28 +186,28 @@ int AddReplayData(REPLAY *rp, int timing, uchar op, short value){
...
+	AddReplayData(rp, 0, 0x65, static_cast<short>(cfg->gaugeType[PLAYER_1]));
+	AddReplayData(rp, 0, 0x97, static_cast<short>(cfg->gaugeType[PLAYER_2]));
+	AddReplayData(rp, 0, 0x66, static_cast<short>(cfg->lanecoverv[PLAYER_1]));
+	AddReplayData(rp, 0, 0x98, static_cast<short>(cfg->lanecoverv[PLAYER_2]));
+	AddReplayData(rp, 0, 0x6b, static_cast<short>(cfg->assist[PLAYER_1]));
+	AddReplayData(rp, 0, 0x9d, static_cast<short>(cfg->assist[PLAYER_2]));
+	AddReplayData(rp, 0, 0x6a, static_cast<short>(cfg->randSC[PLAYER_1]));
+	AddReplayData(rp, 0, 0x9c, static_cast<short>(cfg->randSC[PLAYER_2]));
+	AddReplayData(rp, 0, 0x69, static_cast<short>(cfg->randSC[PLAYER_1]));
+	AddReplayData(rp, 0, 0x9b, static_cast<short>(cfg->randSC[PLAYER_2]));
```
chown2: This got changed from randFix
```suggestion
	AddReplayData(rp, 0, 0x69, static_cast<short>(cfg->randFix[PLAYER_1]));
	AddReplayData(rp, 0, 0x9b, static_cast<short>(cfg->randFix[PLAYER_2]));
```

chown2: There is a bunch of these, replaced with a wrong thing accidentaly

chown2 (commented):
Eyeballed all commits, looks good outside of my comments

LR2/Scene07_Skinselect.cpp
ez2patternBMS: That's strange, I checked that it compiled
I'll fix it.

ez2patternBMS:
commit added ;document link on readme.md

chown2:
I will test this later today/tomorrow

chown2 (changes requested):
After this PR song select shows 'SCORE: NO SAVE' regardless of assist options

ez2patternBMS:
Typo in https://github.com/GOMazk/OpenLR2/commit/dddd3cfb7e471838599c15f88372e76f5c6a8e6a#diff-f334f6519c86abdf1745675e5fd0b6e689212643a91f3cf81d147dc34a892a53L8
Fixed and add a commit that unoptimize if function.

chown2:
What was '[chore: Add comment in structure.h](https://github.com/GOMazk/OpenLR2/pull/203/commits/0ff9de48d6b62823212e86b7349ae10f3e030222)' for? I don't see the need for it (if you are planning to change someting, you can keep this commit for reference in a separate branch, there is enough TODO spam in the code base as it is), and it doesn't compile. I dropped it.

chown2:
I rebased this, if you nede to make any changes, please, pull first

chown2 (approved):
Seems to work fine. This is impossible to review. YOLO! 🚀

PR #209, opened 2026-07-06
Converted else if tree in ReadSkin() to switch statement

tenaibms:
The else if tree in ReadSkin becomes too large for the MSVC compiler if only a few more commands get added, so this is future proofing for future changes.

### DIFF
--- a/LR2/LR2_skinload.cpp
+++ b/LR2/LR2_skinload.cpp
@@ -368,7 +368,7 @@ int InitSkin(skstruct *sk, int /*unused*/, char font) {
 		sk->customfile[i].fillzero();
 	}
 	sk->customfile_count = 0;
-	
+
 	DeleteGraph(sk->GrHandle[GRHTYPE_PREVIEW]);
 	sk->GrHandle[GRHTYPE_PREVIEW] = MakeGraph(skinSizeX, skinSizeY); //TODO_RESOULUTION
 	DeleteGraph(sk->GrHandle[104]);
@@ -427,7 +427,7 @@ int InitImageFont(ImageFont *imgfont) {
 
 int ReadImageFont(CSTR filename, ImageFont *imgfont) {
 	CSTR str1;
-	
+
 	str1 = filename.getDirectory();
 
 	if (strcmp(str1, imgfont->filepath)) {
@@ -449,11 +449,11 @@ int ReadImageFont(CSTR filename, ImageFont *imgfont) {
 			ErrorLogFmtAdd("画像フォントファイル%sの読み込みに失敗しました\n",filename.body);
 			return -1;
 		}
-		
+
 		CSTR str2(256);
 		CSVbuf csvBuf;
 		strcpy(imgfont->filepath, str1);
-		
+
 		while (FileRead_gets(str2, 250, f) != -1) {
 			if (*str2.atPos(0) == '#') {
 				str2.trimWhiteSpace();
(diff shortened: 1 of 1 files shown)


LR2/LR2_skinload.cpp
```diff
@@ -927,7 +927,22 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
...
+					constexpr auto fnv1a = [](std::string_view s) {
+						std::size_t hash = 14695981039346656037ull;
+						for (unsigned char c : s)
+							hash = (hash ^ c) * 1099511628211ull;
+						return hash;
+					};
+
+					auto hashedCommand = fnv1a(command);
+					
+					switch (hashedCommand) {
```
MatVeiQaaa: just run fnv1a() through the switch, no need for that variable as far as i can see.

chown2:
```C++
[[nodiscard]] consteval int operator""_hash(const char* s, size_t len)
{
    return hash(std::string_view{s, len});
}
// ...
// use like "#BPM"_hash
```

LR2/LR2_skinload.cpp
```diff
@@ -927,7 +927,22 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 
 
 				if (!fBuf.left(1).isDiff("#")) {
-					if (fBuf.left(6).isSame("#IMAGE")) {
+					/* extract skin command to std::string_view */
+					std::string_view fBufStringView(fBuf.c_str());
+					auto command = fBufStringView.substr(0, fBufStringView.find(','));
+
+					/* lambda for the Fowler–Noll–Vo hash function */
+					constexpr auto fnv1a = [](std::string_view s) {
```
MatVeiQaaa: maybe we could use a less cryptic name for this method, something really simple like "hash" lol?

LR2/LR2_skinload.cpp
chown2: same with fBufStringView and command

LR2/LR2_skinload.cpp
MatVeiQaaa: not gonna matter with a string literal operator i guess.

MatVeiQaaa:
i rebased your branch. Seems to work, do you want to do anything else?

tenaibms:
No, unless someone feels strongly about the hash operator.

LR2/LR2_skinload.cpp
```diff
@@ -961,22 +974,26 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
...
-							ErrorLogFmtAdd("スキン読み込みエラー %d行目\n%s\nこれ以上#FONTを登録できません。\n", line, fBuf.body);
-						}
-						else {
-							SplitCSV(fBuf, &csv, ",");
-							//sk->fontHandle[sk->num_of_struct] = CreateFontToHandle(sk->fontname, csv.val[1], csv.val[2], csv.val[3], 0, -1, 0, -1, -1); //DxLib3.02
-							sk->fontHandle[sk->num_of_struct] = CreateFontToHandle(sk->fontname.body, csv.val[1], csv.val[2], csv.val[3], DX_CHARSET_UTF8, -1, 0, -1); //DxLib3.24f
-							if (sk->fontHandle[sk->num_of_struct] == -1) {
-								sk->fontHandle[sk->num_of_struct] = 0;
+					case hash("#FONT"): {
+						if (!flag_skipFont) {
```
chown2:
```suggestion
						if (flag_skipFont) break;
```
you should inverse these to improve readibility (avoid indentation drift) and it will also decrease the diff size

MatVeiQaaa (approved):

LR2/LR2_skinload.cpp
tenaibms: I'll leave fBufStringView because otherwise we need to duplicate `std::string_view(fBuf.c_str())` for finding the comma delimiter.

tenaibms:
I have no further changes now, so you can merge.

chown2:
Reformatted and changed style of operator""_hash
Side-note:
why do these have `tSkin_num++;`? gotta check if that is an RE issue
DST_BARGRAPH
DST_BGA

chown2 (approved):

PR #225, opened 2026-07-10
Stretch to fit, Fullscreen scaling filter selection

### LINKED ISSUE #178: Stretch to fit the screen in SD mode fullscreen
Some users use the SD fullscreen mode stretched to fit the screen, but the current fullscreen mode displays according to the aspect ratio.
We need a feature to stretch it to fit the screen by adding another mode.

MatVeiQaaa:
Adds SRC_BUTTON 400 and SRC_TEXT 300 to control the filter that fullscreen scaling uses. Current selection is BILINEAR and NEAREST. It used to be BILINEAR by default. Additionally accepts #FULLSCREEN_FILTER in LR2files/Config/optionstr.csv.

Adds SRC_BUTTON 401 and SRC_TEXT 301 to control if fullscreen scaling will stretch to fit the screen. OFF or ON value. Additionally accepts #FULLSCREEN_FITSTRETCH in LR2files/Config/optionstr.csv.

Extends objectStr array which holds the strings available to the skins to 400 elements from previous 300.
Refactors how options strings are read a little. Should have been just pointless code removal.

New features can be used by editing the config file manually, instead of using GUI.

Closes https://github.com/GOMazk/OpenLR2/issues/178

### DIFF
--- a/LR2/En_xml.cpp
+++ b/LR2/En_xml.cpp
@@ -139,6 +139,10 @@ void WriteXML_Tab2Int(FILE *hFile, const char *tag, int value){
 	fputs(buf, hFile);
 }
 
+void WriteXML_Tab2BoolAsInt(FILE* hFile, const char* tag, bool value) {
+	WriteXML_Tab2Int(hFile, tag, value ? 1 : 0);
+}
+
 void WriteXML_Tab2Str(FILE *hFile, const char *tag, CSTR str){
 	char buf[256];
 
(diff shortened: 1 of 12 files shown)


MatVeiQaaa:
This is how it looks like when used. "CRT" filter mode is something i will add in a separate PR later.

https://github.com/user-attachments/assets/d284ccf3-04f6-4b0c-9c16-1d580d9359cc

PR #228, opened 2026-07-11
Unload stale keysounds of previous song

MatVeiQaaa:
doesn't necessarily fix any bugs, but keeps the state clean, and prevents keysounds of particularly big songs staying in memory forever.

### DIFF
--- a/LR2/En_audio.cpp
+++ b/LR2/En_audio.cpp
@@ -257,10 +257,11 @@ int IsAltSoundExist(CSTR *filepath){
 int ReleaseSound(AUDIO *aud, SOUNDDATA *sound){
 	CSTR &filepath = sound->filename;
 
+	sound->streaming = false;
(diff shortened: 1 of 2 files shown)


LR2/En_audio.cpp
```diff
@@ -257,10 +257,11 @@ int IsAltSoundExist(CSTR *filepath){
 int ReleaseSound(AUDIO *aud, SOUNDDATA *sound){
 	CSTR &filepath = sound->filename;
 
+	sound->streaming = false;
 	if (!sound->load) {
 		return -1;
 	}
-	if (filepath.length() <= 1) {
+	if (filepath.length() == 0) {
```
chown2, line 264: Even the old condition is probably unreachable, IMO the whole block isn't needed

chown2 (approved):

PR #229, opened 2026-07-11
Use correct buffer size when converting encodings, refactor some UTF-8 code

chown2:
See commit messages for details. I have no idea how this worked without problems, but we better not rely on some cryptic magic.

### DIFF
--- a/LR2/En_fileutil.cpp
+++ b/LR2/En_fileutil.cpp
@@ -15,18 +15,6 @@
 
 #ifdef _WIN32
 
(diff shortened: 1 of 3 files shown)


MatVeiQaaa:
why W file methods to A?

chown2:
@MatVeiQaaa it's not necessary to use W methods as we are under the UTF-8 locale. Less code = less bugs.

MatVeiQaaa (approved):

Issue #230, opened 2026-07-11
Opens in wrong resolution despite resolution correctly being set in config


aak64:
<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/d80a674b-a46c-482a-a2a3-3895f02cd761" />

<img width="806" height="445" alt="Image" src="https://github.com/user-attachments/assets/e8a23196-1d49-4b8f-9d0e-8926241d9fae" />

Resolution set to 1, which is supposed to be 720p HD, and should scale 2x to fill a 1440p monitor.

This happens almost every time I update OpenLR2 and last time I fixed it somehow but I forgot how.

Pressing F4 does nothing, except when it goes back to windowed mode, where it displays correctly.

Setting the other options below resolution to 1 also does nothing.

MatVeiQaaa:
SetDrawArea again?

MatVeiQaaa:
@aak64 can you create a shortcut to OpenLR2.exe, add `-dx11` at the end of `target` field in shortcut properties, and see if the problem persists if you launch the game with this shortcut?

aak64:
> [@aak64](https://github.com/aak64) can you create a shortcut to OpenLR2.exe, add `-dx11` at the end of `target` field in shortcut properties, and see if the problem persists if you launch the game with this shortcut?

This works, is there any way to put this in the config or should I just keep using the shortcut?

MatVeiQaaa:
@aak64 -dx11 makes the game render with DirectX 11. It is somewhat WIP, eventually the game will just default to this. I don't think a proper setting for it is necessary. Right now i needed you to do this to confirm a theory about this bug. It will (hopefully) be fixed next release.

MatVeiQaaa:
@aak64 can you confirm this issue is resolved in last release?

aak64:
> [@aak64](https://github.com/aak64) can you confirm this issue is resolved in last release?

Sadly, it didn't work...

<img width="2560" height="1440" alt="Image" src="https://github.com/user-attachments/assets/77b41e14-d809-4c78-bb49-5e5d89f728b0" />

PR #233, opened 2026-07-12
Fix cropped image bug in fullscreen

MatVeiQaaa:
Addresses https://github.com/GOMazk/OpenLR2/issues/230


chown2 (approved):
I can't replicate the original issue, but this doesn't seem to cause any regressions

MatVeiQaaa:
omfg it wasnt merged yet

Issue #215, opened 2026-07-08
Launches on the wrong monitor during borderless fullscreen in multi-monitor setup


PARK-SU:
**Describe the bug**
In a multi-monitor setup, when toggling the screen mode using F4 or launching the game in borderless fullscreen, it launches on an incorrect monitor.

**Expected behavior**
The normal behavior is that LR2 should run on the monitor where the mouse cursor is currently located.

**Actual behavior**
The game is executed on the wrong monitor(Monitor is Randomly Selected) instead of the active one.

+ This seems to be related to #174 .

MatVeiQaaa:
i dont think launching where the cursor is would be good design. Instead, it could launch on the last screen where it was, or on main screen if there is no saved screen or that screen is not present. Maybe, launch at cursor by default instead of main. Also if it would launch where the cursor is, it would drive me insane when debugging lol.

PR #234, opened 2026-07-12
fix(exclusive_fullscreen): correct mouse coordinates when multiscreen environment is present

Skalim112:
Version 260707 introduced the monitor-switching issue tracked in #215, which can be avoided by running the game with the -dx11 argument.

However, with -dx11 there is still a mouse-coordinate issue in multi-monitor setups. This is the same kind of issue that was already fixed for borderless mode: after moving the game between monitors, DxLib can report mouse coordinates relative to the wrong monitor. I mentioned this case in the second point of my comment here:
https://github.com/GOMazk/OpenLR2/pull/156#issuecomment-4794432595

This change reuses the existing WinAPI-based coordinate correction for desktop fullscreen/exclusive mode as well, so both borderless and exclusive fullscreen use the cursor position relative to the actual game window.

### DIFF
--- a/LR2/En_input.cpp
+++ b/LR2/En_input.cpp
@@ -35,7 +35,7 @@ static MIDI midi;
 
 #ifdef _WIN32
 static void FixMousePointWithClientPosition(int* mouseX, int* mouseY) {
-	if (GetUseFullScreenResolutionMode() != DX_FSRESOLUTIONMODE_BORDERLESS_WINDOW) return;
+	if (GetWindowModeFlag() == TRUE && GetUseFullScreenResolutionMode() != DX_FSRESOLUTIONMODE_BORDERLESS_WINDOW) return;
 
 	const HWND hWnd = GetMainWindowHandle();
 	if (hWnd == nullptr) return;


MatVeiQaaa:
At this point you should just have "!= windowed" condition. But honestly this shouldn't even be in LR2 code as thats pretty much DxLib bug.

Skalim112:
> At this point you should just have "!= windowed" condition. But honestly this shouldn't even be in LR2 code as thats pretty much DxLib bug.

Reduced the condition to only skip the correction in windowed mode.

Keeping this as a local workaround as a patch for now while the DxLib-side fix is done.

MatVeiQaaa:
> Keeping this as a local workaround as a patch for now while the DxLib-side fix is done.

Does this imply you will be working on DxLib-side fix? Or do you mean you will leave it to someone else?

Skalim112:
> > Keeping this as a local workaround as a patch for now while the DxLib-side fix is done.
> 
> Does this imply you will be working on DxLib-side fix? Or do you mean you will leave it to someone else?

I don´t know if i am capable of managing something that complex... i meant leave the patch while someone works on the dxlib side

chown2 (approved):
Can't test multiscreen, but no regressions were detected on a single screen setup

PR #237, opened 2026-07-14
Fix crash on scene transition after a video

### LINKED ISSUE #236: FHD Skin crashes several times
https://www.youtube.com/watch?v=GjDiS14LlQ8

(This demonstration video reproduces an issue where a game crash occurs, using a skin shared by another user.)

Frequent game crashes are continuously occurring when using the FHD skin. It happens most often when transitioning to the Result screen. 

For the play skin, #RESOLUTION,1920,1080 is set, and for the result skin, #RESOLUTION,1280,720 is set.

Could this be a simple skin error? For now, I am sharing the video and attaching the log.

https://drive.google.com/file/d/1N4KvLSk2KlJdfh8K-nWka8mYhmr1y9CD/view?usp=sharing

Additionally, it does not occur in Windowed or Borderless mode, and seems to only happen in Exclusive Fullscreen mode.

CC. @Skalim112

---

I have discovered something additional. Game crashes do not occur on songs with BGAs consisting only of BMPs. A game crash always occurs right after a song where the BGA is a video.

MatVeiQaaa:
Closes https://github.com/GOMazk/OpenLR2/issues/236


chown2 (approved):
DX9 DX9EX DX11 - all good
Note: needs a rebase after #233

PR #238, opened 2026-07-14
Add hiterror functionality

tenaibms:
# Summary
Adds the following commands
```csv
#SRC_HITERROR,player,(NULL),(NULL),(NULL),(NULL),(NULL),barcount,(NULL),(NULL),(NULL),(NULL),(NULL),(NULL)
#DST_HITERROR,player,time,x,y,w,h,acc,a,r,g,b,blend,filter,angle,center,loop,timer,op1,op2,op3
#SRC_HITERROR_CENTER,0,0,0,0,0,0,0,0,0,0,0,0,0
#DST_HITERROR_CENTER,(NULL),time,x,y,w,h,acc,a,r,g,b,blend,filter,angle,center,loop,timer,op1,op2,op3

#SRC_HITERROR_POOR,0,0,0,0,0,0,0,0,0,0,0,0,0
#DST_HITERROR_POOR,(NULL),time,x,y,w,h,acc,a,r,g,b,blend,filter,angle,center,loop,timer,op1,op2,op3
#SRC_HITERROR_BAD,0,0,0,0,0,0,0,0,0,0,0,0,0
#DST_HITERROR_BAD,(NULL)

,time,x,y,w,h,acc,a,r,g,b,blend,filter,angle,center,loop,timer,op1,op2,op3
#SRC_HITERROR_GOOD,0,0,0,0,0,0,0,0,0,0,0,0,0
#DST_HITERROR_GOOD,(NULL),time,x,y,w,h,acc,a,r,g,b,blend,filter,angle,center,loop,timer,op1,op2,op3
#SRC_HITERROR_GREAT,0,0,0,0,0,0,0,0,0,0,0,0,0
#DST_HITERROR_GREAT,(NULL),time,x,y,w,h,acc,a,r,g,b,blend,filter,angle,center,loop,timer,op1,op2,op3
#SRC_HITERROR_PGREAT,0,0,0,0,0,0,0,0,0,0,0,0,0
#DST_HITERROR_PGREAT,(NULL),time,x,y,w,h,acc,a,r,g,b,blend,filter,angle,center,loop,timer,op1,op2,op3

#SRC_HITERROR_EMA,0,0,0,0,0,0,0,0,0,0,0,0,0
#DST_HITERROR_EMA,(NULL),time,x,y,w,h,acc,a,r,g,b,blend,filter,angle,center,loop,timer,op1,op2,op3
```
The SRC commands are very similar to `#SRC_BGA` in that for the most part they are entirely ignored, but I added them for convention, if determined unnecessary they can likely be removed, or potentially a proper support for images could be added. The commands defining the color/dimensions of the hiterror bars could also take a player index but I didn't know if that would be that important, so I haven't added it yet.

## Demonstration

https://github.com/user-attachments/assets/25b55a9a-b835-48c7-b137-b917d08179a0

### DIFF
--- a/LR2/LR2_skinload.cpp
+++ b/LR2/LR2_skinload.cpp
@@ -1850,92 +1850,76 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 					case "#SRC_HITERROR"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						if (CheckIndexRange(csv.val[1], 0, 1, line, pFbuf)) {
-							int player = csv.val[1];
-							ReadSRC(&sk->src_HITERROR[player], &csv, sk);
-							for (int i = 0; i < sk->src_HITERROR[player].graphcount; i++)
-								sk->src_HITERROR[player].grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
-
+							ReadSRC(&sk->src_HITERROR[csv.val[1]], &csv, sk);
 						}
 						break;
 					}
 					case "#DST_HITERROR"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						if (CheckIndexRange(csv.val[1], 0, 1, line, pFbuf)) {
-							ReadDST(&sk->dst_HITERROR[csv.val[1]], &csv, tSkin_num);
+							ReadDST(&sk->dst_HITERROR[csv.val[1]], &csv, tSkin_num, line);
 						}
 						break;
 					}
 					case "#SRC_HITERROR_CENTER"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_CENTER, &csv, sk);
-						for (int i = 0; i < sk->src_HITERROR_CENTER.graphcount; i++)
-							sk->src_HITERROR_CENTER.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_CENTER"_hash:{
 						SplitCSV(fBuf, &csv, ",");
-						ReadDST(&sk->dst_HITERROR_CENTER[0], &csv, tSkin_num);
+						ReadDST(&sk->dst_HITERROR_CENTER[0], &csv, tSkin_num, line);
 						sk->dst_HITERROR_CENTER[1] = sk->dst_HITERROR_CENTER[0];
 						break;
 					}
 					case "#SRC_HITERROR_PGREAT"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_PGREAT, &csv, sk);
-						for (int i = 0; i < sk->src_HITERROR_PGREAT.graphcount; i++)
-							sk->src_HITERROR_PGREAT.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_PGREAT"_hash:{
 						SplitCSV(fBuf, &csv, ",");
-						ReadDST(&sk->dst_HITERROR_PGREAT, &csv, tSkin_num);
+						ReadDST(&sk->dst_HITERROR_PGREAT, &csv, tSkin_num, line);
 						break;
 					}
 					case "#SRC_HITERROR_GREAT"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_GREAT, &csv, sk);
-						for (int i = 0; i < sk->src_HITERROR_GREAT.graphcount; i++)
-							sk->src_HITERROR_GREAT.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_GREAT"_hash:{
 						SplitCSV(fBuf, &csv, ",");
-						ReadDST(&sk->dst_HITERROR_GREAT, &csv, tSkin_num);
+						ReadDST(&sk->dst_HITERROR_GREAT, &csv, tSkin_num, line);
 						break;
 					}
 					case "#SRC_HITERROR_GOOD"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_GOOD, &csv, sk);
-						for (int i = 0; i < sk->src_HITERROR_GOOD.graphcount; i++)
-							sk->src_HITERROR_GOOD.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_GOOD"_hash:{
 						SplitCSV(fBuf, &csv, ",");
-						ReadDST(&sk->dst_HITERROR_GOOD, &csv, tSkin_num);
+						ReadDST(&sk->dst_HITERROR_GOOD, &csv, tSkin_num, line);
 						break;
 					}
 					case "#SRC_HITERROR_BAD"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_BAD, &csv, sk);
-						for (int i = 0; i < sk->src_HITERROR_BAD.graphcount; i++)
-							sk->src_HITERROR_BAD.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_BAD"_hash:{
 						SplitCSV(fBuf, &csv, ",");
-						ReadDST(&sk->dst_HITERROR_BAD, &csv, tSkin_num);
+						ReadDST(&sk->dst_HITERROR_BAD, &csv, tSkin_num, line);
 						break;
 					}
 					case "#SRC_HITERROR_EMA"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_EMA, &csv, sk);
-						for (int i = 0; i < sk->src_HITERROR_EMA.graphcount; i++)
-							sk->src_HITERROR_EMA.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_EMA"_hash:{
 						SplitCSV(fBuf, &csv, ",");
-						ReadDST(&sk->dst_HITERROR_EMA, &csv, tSkin_num);
+						ReadDST(&sk->dst_HITERROR_EMA, &csv, tSkin_num, line);
 						break;
 					}
 				}
--- a/LR2/LR2_skinload.cpp
+++ b/LR2/LR2_skinload.cpp
@@ -421,6 +421,7 @@ int InitSkin(skstruct *sk, int /*unused*/, char font) {
 	InitDST(&sk->dst_HITERROR_POOR);
 	InitSRC(&sk->src_HITERROR_EMA);
 	InitDST(&sk->dst_HITERROR_EMA);
+	sk->HiterrorDSTPool.clear();
 	return 1;
 }
 
@@ -1833,6 +1834,20 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 					case "#SRC_HITERROR"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR, &csv, sk);
+						for (int i = 0; i < sk->src_HITERROR.graphcount; i++)
+							sk->src_HITERROR.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
+						sk->HiterrorDSTPool.resize(sk->src_HITERROR.graphcount);
+						for (auto &entry : sk->HiterrorDSTPool) {
+							if (entry.draw == nullptr) {
+								entry.dataSize = 1;
+								entry.dstCount = 1;
+								entry.draw = (DSTdraw*)malloc(sizeof(DSTdraw));
+							}
+							entry.draw[0] = {};
+							entry.draw[0].acc = 1;
+							entry.draw[0].blend = 1;
+							entry.draw[0].sortID = ++tSkin_num;
+						}
 						break;
 					}
 					case "#DST_HITERROR"_hash:{
@@ -1843,6 +1858,8 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 					case "#SRC_HITERROR_CENTER"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_CENTER, &csv, sk);
+						for (int i = 0; i < sk->src_HITERROR_CENTER.graphcount; i++)
+							sk->src_HITERROR_CENTER.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_CENTER"_hash:{
@@ -1853,6 +1870,8 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 					case "#SRC_HITERROR_PGREAT"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_PGREAT, &csv, sk);
+						for (int i = 0; i < sk->src_HITERROR_PGREAT.graphcount; i++)
+							sk->src_HITERROR_PGREAT.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_PGREAT"_hash:{
@@ -1863,6 +1882,8 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 					case "#SRC_HITERROR_GREAT"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_GREAT, &csv, sk);
+						for (int i = 0; i < sk->src_HITERROR_GREAT.graphcount; i++)
+							sk->src_HITERROR_GREAT.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_GREAT"_hash:{
@@ -1873,6 +1894,8 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 					case "#SRC_HITERROR_GOOD"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_GOOD, &csv, sk);
+						for (int i = 0; i < sk->src_HITERROR_GOOD.graphcount; i++)
+							sk->src_HITERROR_GOOD.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_GOOD"_hash:{
@@ -1883,6 +1906,8 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 					case "#SRC_HITERROR_BAD"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_BAD, &csv, sk);
+						for (int i = 0; i < sk->src_HITERROR_BAD.graphcount; i++)
+							sk->src_HITERROR_BAD.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_BAD"_hash:{
@@ -1893,6 +1918,8 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 					case "#SRC_HITERROR_POOR"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_POOR, &csv, sk);
+						for (int i = 0; i < sk->src_HITERROR_POOR.graphcount; i++)
+							sk->src_HITERROR_POOR.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_POOR"_hash:{
@@ -1903,6 +1930,8 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 					case "#SRC_HITERROR_EMA"_hash:{
 						SplitCSV(fBuf, &csv, ",");
 						ReadSRC(&sk->src_HITERROR_EMA, &csv, sk);
+						for (int i = 0; i < sk->src_HITERROR_EMA.graphcount; i++)
+							sk->src_HITERROR_EMA.grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
 						break;
 					}
 					case "#DST_HITERROR_EMA"_hash:{
--- a/LR2/LR2_skinload.cpp
+++ b/LR2/LR2_skinload.cpp
@@ -975,7 +975,7 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 				if (!fBuf.left(1).isDiff("#")) {
 					std::string_view fBufStringView(fBuf.c_str());
 					switch (hash(fBufStringView.substr(0, fBufStringView.find(',')))) {
-					case "#IMAGE"_hash:{
+					case "#IMAGE"_hash: {
 						if (sk->count == 100) {
 							ErrorLogFmtAdd("スキン読み込みエラー %d行目\n%s\nこれ以上#IMAGEを登録できません。\n", line, fBuf.body);
 						}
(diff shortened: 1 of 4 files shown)


LR2/structure.h
```diff
@@ -1012,6 +1030,20 @@ struct skstruct {
 	int event_FADEOUT[10]{};
 	struct DSTstruct dst_EVENT_LOADINGBG[5]{};
 	int horizontal{};
+	struct SRCstruct src_HITERROR[2] {};
+	struct DSTstruct dst_HITERROR[2] {};
+	struct SRCstruct src_HITERROR_CENTER {};
+	struct DSTstruct dst_HITERROR_CENTER[2] {};
```
tenaibms: This does not need to be doubled since the values are copied when drawn, can just be one definition.

LR2/Scene04_Play.cpp
```diff
@@ -787,6 +787,8 @@ int ProcSinglenote(game *g, int lane, int keypress, int timing, int player) {
 			increment_extended(extendedStatsCourse, isFast, offset);
 			increment_extended(extendedColumnStatsCourse, isFast, offset);
 			lastOffsetColumnIdx = lane;
+			g->gameplay.player[player].hiterror.notes.push({ (size_t)lane, (double)offset, (double)	timing, 5 });
```
chown2: nit: use static_cast instead of C-style casts https://www.youtube.com/watch?v=DAvZ3OG9cNo

LR2/Scene04_Play.cpp
```diff
@@ -1119,6 +1127,90 @@ static void QuickRestart(game& game, bool newRandom) {
...
+					break;
+				default: 
+					continue;
+			}
+
+			if (!parentSRC || parentSRC->graphcount <= 0) continue;
+
+			childDST = CloneDSTstruct(parentDST);
+			
+			for (size_t j = 0; j < childDST.dstCount; ++j) {
```
chown2: nit: use range loops in new code
In this case specifically you can use std::span{childDST.draw, childDST.dstCount}

LR2/Scene04_Play.cpp
```diff
@@ -1119,6 +1127,90 @@ static void QuickRestart(game& game, bool newRandom) {
...
+					parentDST = &sk->dst_HITERROR_BAD;
+					parentSRC = &sk->src_HITERROR_BAD;
+					break;
+				default: 
+					continue;
+			}
+
+			if (!parentSRC || parentSRC->graphcount <= 0) continue;
+
+			childDST = CloneDSTstruct(parentDST);
```
chown2: 1) declare when initialized
2) why not `DSTstruct childDST = parentDST;`?

chown2: 2. ah, because 'draw' is a pointer. Well, instead of doing a temporary allocaiton every frame you can store a second buffer near the original one.

chown2: that would need a note that it has a copy to fade alpha. although it seems like this should've been achieved automatically by timers.

LR2/Scene04_Play.cpp
```diff
@@ -1119,6 +1127,90 @@ static void QuickRestart(game& game, bool newRandom) {
...
+
+		for (int i = 0; i < hiterrorData.notes.size(); i++) {
+			const JudgeData& jd = hiterrorData.notes[i];
+			SRCstruct *parentSRC = nullptr;
+			DSTstruct *parentDST = nullptr;
+
+			DSTstruct childDST = DSTstruct{};
+
+			switch (jd.judge) {
+				case 5:
```
chown2: Do we not have named constants for judges? You can add an enum locally for now.
```c
enum {
\tJUDGEMENT_BAD = 2,
\tJUDGEMENT_GOOD = 3,
...
};
```

chown2: extending on this:
make this an enum class, make JudgeData::judge use that enum class.
to convert into it, add a switch with cases like `case 2: return XXX::BAD` instead of using a static_cast. casting into enum class with values outside of it's range is UB, that will make the conversion explicit.

LR2/structure.h
```diff
@@ -1283,6 +1315,51 @@ struct EXTENDEDPLAYERSTATS {
 	int lastFastSlow = 0;
 };
 
+struct EMA {
+	double value = 0.0;
+	static constexpr double alpha = 0.07;
+	void add(double x) { value = alpha * x + (1.0 - alpha) * value; }
+	void reset() { value = 0.0; }
```
chown2: reset method seems redundant, `asdf.ema = {};` achieves the same thing.

LR2/structure.h
```diff
@@ -1283,6 +1315,51 @@ struct EXTENDEDPLAYERSTATS {
...
+template<typename T>
+class CircularBuffer {
+	std::vector<T> buf;
+	size_t head = 0;
+
+public:
+	CircularBuffer() = default;
+	explicit CircularBuffer(size_t capacity) : buf(capacity) {}
+
+	void push(const T& val) {
```
chown2, line 1314: This class is overly generic for this use case but it's fine.
Gotta test if our compilers support 'deducing this' yet to avoid duplicating this method.

LR2/Scene04_Play.cpp
tenaibms: 1. Will do
2. DSTstruct has `DSTdraw*` members so to clone and change things without messing up the parent we need to do use the copy function.

LR2/Scene04_Play.cpp
tenaibms: I believe in this case it's possible to just use int for offset, lane isn't used and can be removed, and timing can also use an int, so it's possible to avoid casting all-together.

LR2/Scene04_Play.cpp
chown2: Doesn't SetDSTdrawByTime already fade objects?

LR2/LR2_skinload.cpp
```diff
@@ -1829,13 +1847,104 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
 						sk->horizontal = 1;
 						break;
 					}
+					case "#SRC_HITERROR"_hash:{
+						SplitCSV(fBuf, &csv, ",");
+						if (CheckIndexRange(csv.val[1], 0, 1, line, pFbuf)) {
+							int player = csv.val[1];
+							ReadSRC(&sk->src_HITERROR[player], &csv, sk);
+							for (int i = 0; i < sk->src_HITERROR[player].graphcount; i++)
+								sk->src_HITERROR[player].grHandles[i] = sk->GrHandle[GRHTYPE_WHITE];
```
MatVeiQaaa: i'm really not sure if using GRHTYPE_WHITE is a good idea. First of all, you can remove this line at all and it would use a texture defined in .csv already, no code required, there is no reason to not use it. Also, right now you are leaking a texture each scene load if the user does define it. Then, the user can set isDrawBackbox instead of this, probably. Or, if none of isDrawBackbox and gr is set, you can generate a texture yourself by doing MakeScreen -> SetDrawScreen -> DrawBox -> SetDrawScreen(DX_SCREEN_BACK), or keep using GRHTYPE_WHITE if you insist.

MatVeiQaaa: and also, someone could specify the use of GRHTYPE_WHITE in the src itself, using it as gr. There is no need for this code at all. In the end i recommend doing absolutely nothing but simply removing this line.

MatVeiQaaa: besides, we got three SRC pointing to same texture. This can't be good, as there are ownership issues. Has to use DerivationGraph in cases like that. In this case you could modify gr on the input to ReadSRC, which would use DerivationGraph on it. I'm just saying for educational purposes though, what you should do is leave this to .csv definition.

LR2/Scene04_Play.cpp
tenaibms: > Doesn't SetDSTdrawByTime already fade objects?

It seems like it, but I think it might be a bit more complicated to do it with this since we'd need to add an extra draw with 0 alpha for each dst the user added, I feel like just using the lambda is easier since we are only changing alpha anyway



> 2. ah, because 'draw' is a pointer. Well, instead of doing a temporary allocaiton every frame you can store a second buffer near the original one.

I'm not sure if this is the best option, since each judgement's DST may have a different number of DST's defined, maybe a buffer for each of them? Or a single buffer with the max number of draw commands required by all of the judgements I guess if we don't want to do that. Also when you say near do you mean in the skstruct?

MatVeiQaaa:
@tenaibms `git rebase` exists :D

tenaibms:
> @tenaibms `git rebase` exists :D

Is it better to do that than to merge? I am not sure how they differ.

LR2/Scene04_Play.cpp
tenaibms: Added an enum class as per your suggestion.

LR2/structure.h
tenaibms: Fixed

chown2:
@tenaibms different projects want this handled differently.
Personally, I believe it's best if you amend into previous commits and force push instead of adding fixup commits.
Although, since you intend this to be merged into a single commit, we might as well do that ourselves when merging into master.
It starts to matter when you intend to have several commits in a PR, like `refactor: do x to prepare for feature y` and `feat: add feature y`.

tenaibms:
> @tenaibms different projects want this handled differently. Personally, I believe it's best if you amend into previous commits and force push instead of adding fixup commits. Although, since you intend this to be merged into a single commit, we might as well do that ourselves when merging into master. It starts to matter when you intend to have several commits in a PR, like `refactor: do x to prepare for feature y` and `feat: add feature y`.

I see, I'll do that in the future.

LR2/LR2_skinload.cpp
tenaibms: Changed as per your suggestion.

LR2/Scene04_Play.cpp
tenaibms: Can I in this case? I need to iterate over both childDST.draw and parentDST->draw.

chown2: try `for (auto [child, parent] : std::views::zip(std::span(childDST.draw, childDST.dstCount), std::span(parentDST->draw, parentDST->dstCount))) { /* ... */ }`

LR2/structure.h
chown2: 1) deducing this does work on GCC, clang, and MSVC!
2) these functions don't overload on 'this'...

So it's simpler, you probably wanted
```cpp
	void push(T&& val) {
		if (buf.empty()) return;
		buf[head] = std::forward<T>(val);
		head = (head + 1) % buf.size();
	}
```

LR2/structure.h
```diff
@@ -1283,6 +1315,51 @@ struct EXTENDEDPLAYERSTATS {
...
+};
+
+template<typename T>
+class CircularBuffer {
+	std::vector<T> buf;
+	size_t head = 0;
+
+public:
+	CircularBuffer() = default;
+	explicit CircularBuffer(size_t capacity) : buf(capacity) {}
```
chown2: The default constructor in presense of this other constructor makes a crippled instance of this class, it serves no purpose. Maybe we should then only provide a default constructor and have a separate method `void reset(size_t capacity);`? That would also let us reuse capacity.

LR2/Scene04_Play.cpp
chown2: This is ugly but we can flesh out the details later. Let's just get this in for now. This code doesn't run if there aren't matching skin elements anyway, so no impact in the usual case, everybody else opts in.

LR2/structure.h
```diff
@@ -720,6 +720,24 @@ struct DSTstruct { /* 44bytes.4*0x0b */
 	int dstCount;
 };
 
+inline DSTstruct CloneDSTstruct(const DSTstruct* src) {
+	DSTstruct clone = *src;
```
chown2: Change signature on this one, then move it and FreeDSTstructClone to Scene04_Play.cpp. Also add 'static' for all new methods unless they need to be exposed to other modules. DrawHitError wouldn't be in the header then.
```suggestion
DSTstruct CloneDSTstruct(DSTstruct dst) {
```

chown2: github broke the suggestion, good job

LR2/LR2_skinload.cpp
```diff
@@ -1829,13 +1847,88 @@ int ReadSkin(skstruct *sk,CSTR FilePath, int unused, int skin_num, SkinUser* sku
...
+						ReadDST(&sk->dst_HITERROR_EMA, &csv, tSkin_num, line);
+						break;
 					}
 				}
 				tSkin_num++;
 			}
 		}
 		*fBuf.atPos(0) = '\0';
 	}
+	}
```
chown2: Looking at the diff alone, I don't understand where this brace came from

LR2/Scene04_Play.cpp
```diff
@@ -1119,6 +1127,88 @@ static void QuickRestart(game& game, bool newRandom) {
 	StopAllKeysound(&game);
 }
 
+int DrawHitError(game *g, skstruct *sk, Timer *T) {
+	for (int p = 0; p < 2; p++) {
```
chown2: To avoid indentation drift, move this into a helper like DrawHitErrorForPlayer with the same signature but also an `int player` and call it from here.
Call it like DrawHitErrorForPlayer(PLAYER_1) DrawHitErrorForPlayer(PLAYER_2). Makes it easier to step through in a debugger.

chown2: Maybe EMA and 50-last-notes cases can also be split into several helpers, with an additional one for the src_HITERROR and src_HITERROR_CENTER. But let's think about it after moving this out of the loop first.

LR2/LR2_skinload.cpp
chown2: I feel like it was misplaced and has to be after the 'DST_HITERROR_EMA' block instead.

LR2/Scene04_Play.cpp
```diff
@@ -1119,6 +1127,88 @@ static void QuickRestart(game& game, bool newRandom) {
...
+
+		AddDrawingBuffer_Image(&sk->drBuf, &sk->src_HITERROR[p], &sk->dst_HITERROR[p], T);
+
+		if (sk->src_HITERROR_CENTER.graphcount > 0) {
+			center(&sk->dst_HITERROR_CENTER[p]);
+			AddDrawingBuffer_Image(&sk->drBuf, &sk->src_HITERROR_CENTER, &sk->dst_HITERROR_CENTER[p], T);
+		}
+		
+		double fadeTime = 0.75 * 1000;
+		if (hiterrorData.notes.size() > 50) fadeTime = hiterrorData.notes.size() / 50 * 1000;
```
chown2: Looks like integer divison

MatVeiQaaa:
@tenaibms check DP btw. Maybe it could make sense to have separate hiterror for sides, as well as a shared one.

LR2/structure.h
tenaibms: seems reasonable as the reset method will always be called before it is used for storage

LR2/Scene04_Play.cpp
MatVeiQaaa: > > Doesn't SetDSTdrawByTime already fade objects?
> 
> It seems like it, but I think it might be a bit more complicated to do it with this since we'd need to add an extra draw with 0 alpha for each dst the user added, I feel like just using the lambda is easier since we are only changing alpha anyway

You just need to have a double variable field for each node. You then run GetTimeWrap into it once on creation. Then you pass this value to SetDSTdrawByTime. It is not complicated. SetDSTdrawByTime then returns you an entirely new DSTdraw structure which you use for AddDrawingBuf. This means you only need a single DSTstruct for all of your nodes. Also, when SetDSTdrawByTime returns you a DSTdraw with `time` field set to -1, indicates that the animation has finished and the texture will no longer be drawn. You can use it to remove your node from array. So, in other words, it's all extremely simple, and new code for this feature can be really, really minimal, like around 50 lines total.

MatVeiQaaa: >     2. DSTstruct has `DSTdraw*` members so to clone and change things without messing up the parent we need to do use the copy function.

DSTdraw in DSTstruct is not actually used for drawing, its used by SetDSTdrawByTime to sample the values to generate a new DSTdraw. Those in DSTstruct are only meant to define what the beginning and final state are supposed to be.

LR2/LR2_skinload.cpp
tenaibms: fixed this I think

LR2/structure.h
tenaibms: removed the pointer signature, replaced with pass by value

LR2/Scene04_Play.cpp
```diff
@@ -1127,6 +1127,25 @@ static void QuickRestart(game& game, bool newRandom) {
...
+	DSTstruct clone = src;
+	if (src.dstCount > 0 && src.draw) {
+		clone.draw = (DSTdraw *) malloc(src.dstCount * sizeof(DSTdraw));
+		memcpy(clone.draw, src.draw, src.dstCount * sizeof(DSTdraw));
+	}
+	else {
+		clone.draw = nullptr;
+	}
+	return clone;
+}
```
MatVeiQaaa: you shouldn't need this method. Dont use DSTstruct to render.

tenaibms: I'm not really sure what you mean by this, can you elaborate? I thought you have to use DSTstruct to render

MatVeiQaaa: @tenaibms this is talking about the same. You use DSTstruct to render only indirectly.
https://github.com/GOMazk/OpenLR2/pull/238#discussion_r3638772883

tenaibms: Ok, I think I understand

tenaibms: Removed the clone but I'm leaving the alpha fading method because it's significantly cheaper than using SetDSTdrawByTime as it only interpolates the alpha value and not every other value.

LR2/Scene04_Play.cpp
tenaibms: Done with the first comment.

LR2/Scene04_Play.cpp
MatVeiQaaa: > I'm leaving the alpha fading method because it's significantly cheaper than using SetDSTdrawByTime as it only interpolates the alpha value and not every other value.

First of all, the cost is completely irrelevant here, it is still insignificant, and you don't have thousands nodes either. Then, it will only interpolate if the value changes between two states. If someone doesn't make an animation change a value, it will do no processing. Have you looked inside ``ChangeValueByTime``?

But either way, there is a hurdle that you would have to use two timers at once, which we can't do as of now.

LR2/Scene04_Play.cpp
tenaibms: fixed

LR2/Scene04_Play.cpp
tenaibms: Alright, so it's fine to leave it like this for now until we can utilize two timers?

LR2/structure.h
tenaibms: I have applied your suggestion.

chown2: You don't need the const overload now @tenaibms

LR2/structure.h
chown2: Fixed https://github.com/GOMazk/OpenLR2/pull/238/changes/618af3038476bdc1d0c76875e982e5e8389919a5

LR2/LR2_skinload.cpp
```diff
@@ -419,6 +419,23 @@ int InitSkin(skstruct *sk, int /*unused*/, char font) {
 	for (int i = 0; i < 5; i++) {
 		InitDST(&sk->dst_EVENT_LOADINGBG[i]);
 	}
+
+	for (int i = 0; i < 2; i++) {
```
chown2: Here and twice in DrawHitErrorForPlayer and once in ProcS_Play: use range loops instead of raw loops in new code

LR2/Scene04_Play.cpp
```diff
@@ -1119,6 +1127,98 @@ static void QuickRestart(game& game, bool newRandom) {
 	StopAllKeysound(&game);
 }
 
+int DrawHitErrorForPlayer(game *g, skstruct *sk, Timer *T, int player) {
```
chown2: These pointers are dereferenced unconditionally, so they are not meant to be nullable. Use references instead.
Yes, timer is only passed as a pointer, but I'm pretty sure it's not meant to be nullable either.
```suggestion
int DrawHitErrorForPlayer(game &g, skstruct &sk, Timer &T, int player) {
```
Same with DrawHitError. Also, since DrawHitError is only used in this file, you can remove it from the header and make the function 'static'.

LR2/Scene04_Play.cpp
```diff
@@ -1119,6 +1127,98 @@ static void QuickRestart(game& game, bool newRandom) {
...
+	double noteTimer = GetTimeLapse(142, &g->timer1);
+
+	HITERRORDATA &hiterrorData = g->gameplay.player[player].hiterror;
+
+	AddDrawingBuffer_Image(&sk->drBuf, &sk->src_HITERROR[player], &sk->dst_HITERROR[player], T);
+
+	if (sk->src_HITERROR_CENTER.graphcount > 0) {
+		center(&sk->dst_HITERROR_CENTER);
+		AddDrawingBuffer_Image(&sk->drBuf, &sk->src_HITERROR_CENTER, &sk->dst_HITERROR_CENTER, T);
+	}
```
chown2: This block can be in it's own function

LR2/Scene04_Play.cpp
```diff
@@ -1119,6 +1127,98 @@ static void QuickRestart(game& game, bool newRandom) {
...
+
+		parentDSTByTime.x += offset(jd.offset);
+
+		AddDrawingBuffer(&sk->drBuf, grh, &parentDSTByTime);
+	}
+
+	if (sk->src_HITERROR_EMA.graphcount > 0) {
+		center(&sk->dst_HITERROR_EMA);
+		AddDrawingBuffer_Object(&sk->drBuf, &sk->src_HITERROR_EMA, &sk->dst_HITERROR_EMA, T, offset(hiterrorData.ema.value), 0);
+	}
```
chown2: Same, This block can be in it's own function

LR2/Scene04_Play.cpp
```diff
@@ -1119,6 +1127,98 @@ static void QuickRestart(game& game, bool newRandom) {
...
+		case Judgement::GOOD:
+			parentDST = &sk->dst_HITERROR_GOOD;
+			parentSRC = &sk->src_HITERROR_GOOD;
+			break;
+		case Judgement::BAD:
+			parentDST = &sk->dst_HITERROR_BAD;
+			parentSRC = &sk->src_HITERROR_BAD;
+			break;
+		default:
+			continue;
```
chown2: For potential compiler-assisted refactorings. We are unlikely to add new judgements, but it's better to follow a practice everywhere.
```suggestion
		case AIR_POOR:
		case MISS_POOR:
			continue;
```

LR2/Scene04_Play.cpp
MatVeiQaaa: > Alright, so it's fine to leave it like this for now until we can utilize two timers?

yes, sure, im just saying to share a thought.

LR2/Scene04_Play.cpp
tenaibms: Done

LR2/Scene04_Play.cpp
tenaibms: The function is not long enough to warrant this I feel like, and trying to make it its own function feels dumb because there is so much information I have to pass to the helper function.

chown2: Not sure about 'much information', seems like just two variables to me, hiterrorData.ema and hiterrorByTime.
My issue is that conceptually HITERROR_EMA, HITERROR_CENTER, and HITERROR_BAD/etc are all the same kind of object, an image you'd render based on a certain hit-time-off-by-x offset. And they they are all handled differently here. The hiterrorData.notes loop doesn't even resemble how HITERROR_EMA and HITERROR_CENTER are rendered, and has a lot of preparation work not used by them.

tenaibms: Very well then

tenaibms: The problem is `center()` is a lambda that relies on `hiterrorByTime` being available so I guess I could move that out and make `hiterrorByTime` a parameter instead. Would you be ok with the hiterror functions being lambdas instead of functions?

chown2: > The problem is center() is a lambda that relies on hiterrorByTime being available so I guess I could move that out and make hiterrorByTime a parameter instead

Sure! But better move out centerDraw() instead and remove center() altogether, it just copies centerDraw implementation and puts it in a loop.

> Would you be ok with the hiterror functions being lambdas instead of functions?

Them being lambdas won't change anything. If you don't want them as standalone functions, how about making the variables only needed for the `hiterrorData.notes` for loop into it's own block?

```
	for (int i = 0; i < hiterrorData.notes.size(); i++) {
...
	}
```
->
```
	{
		double noteTimer = ...;
		for (int i = 0; i < hiterrorData.notes.size(); i++) {
...
		}
	}
```

LR2/Scene04_Play.cpp
```diff
@@ -1119,6 +1127,100 @@ static void QuickRestart(game& game, bool newRandom) {
...
+
+	constexpr auto fadeAlpha = [](int originalAlpha, int time, int fadeTime) -> int {
+		return originalAlpha - (time * originalAlpha / fadeTime);
+	};
+
+	double noteTimer = GetTimeLapse(142, &g.timer1);
+
+	HITERRORDATA &hiterrorData = g.gameplay.player[player].hiterror;
+
+	AddDrawingBuffer_Image(&sk.drBuf, &sk.src_HITERROR[player], &sk.dst_HITERROR[player], &T);
```
chown2: This isn't ever gonna draw anything if it's used like in the PR's description.
Actually, I have some more thoughts about the current skin commands usage, but I want to try using the PR in it's current form first.

chown2:
I will try using this soon. Thank you for working on this!

LR2/Scene04_Play.cpp
tenaibms: I can upload a working csv for wmix play if you'd like.

tenaibms:
Not sure if this is an appropriate PR comment but I would just like to let y'all know that I still want to work on this I've just been swamped irl with moving house, so I haven't had time. When I have my computer set up I will begin work again.

tenaibms:
> @tenaibms check DP btw. Maybe it could make sense to have separate hiterror for sides, as well as a shared one.

This is added in now, it uses index 2/3 so that one can still use the shared hiterror in dp, depending on what someone wants to do.

PR #240, opened 2026-07-15
Add CRT fullscreen filter option

MatVeiQaaa:
The promised 3rd value of fullscreenfilter.
div for button 400 changed from 2 to 3.

Also removes the fix introduced in https://github.com/GOMazk/OpenLR2/pull/202/changes/e6dffbb0e3c72a2f4536c48edd5172696eba3347, as it seems to no longer be needed. Something else fixed the bug (properly), probably reset code.


dep/DxLib/Windows/DxGraphicsD3D11.cpp
```diff
@@ -4410,6 +4414,11 @@ extern int Graphics_D3D11_Shader_Initialize( void )
 				goto ERR ;
 			}
 		}
+
+		if (Graphics_D3D11_CompilePixelShader("CRT_Geom.hlsl", "CRTGeom_PS", NULL, &Shader->Base.CrtGeomPixelShader) != 0)
+		{
+			DXST_LOGFILEFMT_ADDUTF16LE(("Couldn't compile CRT_Geom pixel shader\n"));
```
chown2: Don't you need to use a non-utf16 logging function here?

MatVeiQaaa: Maybe? Does such function exist?

chown2: DXST_LOGFILE_ADDA probably

chown2 (approved):
DX{9{,EX},11} all work. Pls double check logging.

MatVeiQaaa:
Logging is fixed and confirmed to work.

PR #241, opened 2026-07-15
Fix overlapping text on login progress screen

### LINKED ISSUE #239: "Logging into Internet Rankings" text overlaps with "Connected to..." text on initialization screen
Image

As seen on this screenshot. All the CustomIR DLLs are the latest released versions that could be obtained.

If that matters, the game is also launched through Wine (vanilla one)

MatVeiQaaa:
Closes https://github.com/GOMazk/OpenLR2/issues/239

### DIFF
--- a/LR2/Main.cpp
+++ b/LR2/Main.cpp
@@ -547,6 +547,7 @@ int main(int argc, char** argv) {
 				printfDx("%.*s%s\n", static_cast<int>(name.size()), name.data(), isFutureReady(future) ? " - done!" : ellipsis(i, 3, 20).c_str());
 			ScreenFlip();
 			clsDx();
+			ClsDrawScreen();
 			std::this_thread::sleep_for(std::chrono::milliseconds(10));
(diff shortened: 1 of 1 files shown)


LR2/Main.cpp
```diff
@@ -547,6 +547,7 @@ int main(int argc, char** argv) {
 				printfDx("%.*s%s\n", static_cast<int>(name.size()), name.data(), isFutureReady(future) ? " - done!" : ellipsis(i, 3, 20).c_str());
 			ScreenFlip();
 			clsDx();
+			ClsDrawScreen();
```
chown2, line 550: Since this is otherwise masked by `DrawExtendGraph` above, would it make sense to instead put into an else branch there?

chown2 (approved):

LR2/Main.cpp
MatVeiQaaa: I think this would be overengineering.

PR #242, opened 2026-07-15
feat(customir): add rival support

luxurya111:
This feature will carry out the work of allowing the existing RIVAL folder to be used via Custom IR.

While it follows the legacy logic of the existing LR2IR as is, unlike the previous logic which retrieved data in XML format, anyone implementing a Custom IR can choose to create it in either XML or JSON format. This is highly similar to the existing GetResultRank logic.

~~I have added sqlite3.c to ExampleIR.cpp. Since the actual rival folder reads from the DB, the actual example now requires sqlite3.c as well.~~

### DIFF
--- a/ExampleIR/ExternalIR.cpp
+++ b/ExampleIR/ExternalIR.cpp
@@ -113,6 +113,31 @@ static std::string OLR2_IR_API GetWebRankingUrl(const char* songHash) {
 	return std::format("https://example.com/{}", songHash);
 }
 
+static openlr2::GetStatus OLR2_IR_API GetRivals(openlr2::IRRivalListResult& out) {
+    std::println(std::cout, "GetRivals");
+    // A real module would fetch the rival list from its service (up to 20 rival ids/names).
+    // ExampleIR returns static in-memory data to show the API shape (no sqlite needed).
+    out = {};
+    out.fetched_at = 1;
+    out.rivals = {
+        { .id = 70100, .name = "ExampleRivalA" },
+        { .id = 70101, .name = "ExampleRivalB" },
+    };
+    return openlr2::GetStatus::Ok;
+}
+
+static openlr2::GetStatus OLR2_IR_API SyncRivalScores(int rivalId, uint64_t /*lastUpdateHint*/, std::vector<openlr2::IRRivalScore>& out) {
+    std::println(std::cout, "SyncRivalScores({})", rivalId);
+    // A real module would fetch this rival's complete score snapshot from its service.
+    // ExampleIR returns static in-memory data; OpenLR2 writes the .db and .lr2folder from it.
+    // Replace the hashes with real chart md5s to make the scores show up in-game.
+    out = {
+        { .hash = "e6f2a1c48b7d0359e1a2b3c4d5e6f708", .clear = 4, .notes = 1200, .combo = 1180, .pg = 1000, .gr = 180, .gd = 15, .bd = 3, .pr = 2, .minbp = 5, .option = 0, .lastupdate = 1262304000 },
+        { .hash = "1a2b3c4d5e6f70819aabbccddeeff0011", .clear = 5, .notes = 900, .combo = 900, .pg = 880, .gr = 20, .gd = 0, .bd = 0, .pr = 0, .minbp = 0, .option = 0, .lastupdate = 1262304000 },
+    };
+    return openlr2::GetStatus::Ok;
+}
+
 extern "C" OLR2_IR_EXPORT void OLR2_IR_API GetMethodTable(MethodTable& table) {
     // Fill out the pointers to methods you want to use. Leave them at nullptr if you don't want to use them.
     // As API gets updated, new methods may appear available at MethodTable, but old ones will never be removed or their
@@ -124,6 +149,8 @@ extern "C" OLR2_IR_EXPORT void OLR2_IR_API GetMethodTable(MethodTable& table) {
     table.RestoreCachedRank = &RestoreCachedRank;
     table.GetGhost = &GetGhost;
     table.GetWebRankingUrl = &GetWebRankingUrl;
+    table.GetRivals = &GetRivals;
+    table.SyncRivalScores = &SyncRivalScores;
 }
 
 #ifdef _WIN32
--- a/LR2/LR2_customir.cpp
+++ b/LR2/LR2_customir.cpp
@@ -440,94 +440,117 @@ bool CUSTOMIR_MANAGER::IsDisplayIrOnline() const {
 	return (*irIt)->DidLoginSuccessfully();
 }
 
+bool CUSTOMIR_MANAGER::RivalSyncBatch::HasTasks() const {
+	return std::ranges::any_of(providers, [](const RivalSyncProvider& p) { return !p.tasks.empty(); });
+}
+
 CUSTOMIR_MANAGER::RivalSyncBatch CUSTOMIR_MANAGER::SyncRivals() {
 	RivalSyncBatch batch{};
 	mRivalPaths.clear();
 
-	if (!IsDisplayIrOnline()) {
-		ErrorLogAdd("CustomIR rival sync skipped: display IR offline\n");
-		return batch;
+	for (const auto& ir : mModules) {
+		if (!ir->DidLoginSuccessfully()) continue;
+
+		openlr2::IRRivalListResult list{};
+		switch (ir->GetRivals(list)) {
+		case openlr2::GetStatus::Ok:
+			break;
+		case openlr2::GetStatus::Retry:
+			OverlayNotification("'%s' GetRivals retry ignored\n", ir->Name().c_str());
+			continue;
+		case openlr2::GetStatus::Fail:
+		default:
+			if (ir->Name() == mDisplayIr) {
+				OverlayNotification("'%s' GetRivals failed - falling back to legacy rival\n", ir->Name().c_str());
+			}
+			continue;
+		}
+
+		RivalSyncProvider provider{};
+		provider.providerName = ir->Name();
+		provider.listOk = true;
+		const std::filesystem::path rivalDirectory = CustomIRRivalDirectory(provider.providerName);
+		std::error_code ec;
+		std::filesystem::create_directories(rivalDirectory, ec);
+		provider.tasks.reserve(list.rivals.size());
+		for (const auto& entry : list.rivals) {
+			if (entry.id < 1) continue;
+			const uint64_t lastUpdateHint = CustomIRRivalLastUpdateHint(rivalDirectory, entry.id);
+
+			provider.tasks.push_back(RivalSyncTask{
+				.name = entry.name,
+				.result = std::async(std::launch::async, [ir, id = entry.id, lastUpdateHint]() -> std::optional<std::vector<openlr2::IRRivalScore>> {
+					std::vector<openlr2::IRRivalScore> scores;
+					const auto syncStatus = ir->SyncRivalScores(id, lastUpdateHint, scores);
+					if (syncStatus != openlr2::GetStatus::Ok) {
+						ErrorLogFmtAdd("SyncRivalScores %s for %s id=%d\n",
+							syncStatus == openlr2::GetStatus::Retry ? "retry ignored" : "failed",
+							ir->Name().c_str(),
+							id);
+						return std::nullopt;
+					}
+					return scores;
+				}),
+				.id = entry.id,
+			});
+		}
+		batch.providers.push_back(std::move(provider));
 	}
-	const auto irIt = std::ranges::find(mModules, mDisplayIr, &CustomIR::Name);
-	if (irIt == mModules.end()) return batch;
-	auto ir = *irIt;
-	batch.providerName = ir->Name();
-	const std::filesystem::path rivalDirectory = CustomIRRivalDirectory(batch.providerName);
-	std::error_code ec;
-	std::filesystem::create_directories(rivalDirectory, ec);
 
-	openlr2::IRRivalListResult list{};
-	switch (ir->GetRivals(list)) {
-	case openlr2::GetStatus::Ok:
-		break;
-	case openlr2::GetStatus::Retry:
-		OverlayNotification("'%s' GetRivals retry ignored\n", ir->Name().c_str());
-		return batch;
-	case openlr2::GetStatus::Fail:
-	default:
-		OverlayNotification("'%s' GetRivals failed - falling back to legacy rival\n", ir->Name().c_str());
-		return batch;
+	if (batch.providers.empty()) {
+		ErrorLogAdd("CustomIR rival sync skipped: no logged-in CustomIR with rivals\n");
 	}
-
-	batch.supported = true;
-	batch.tasks.reserve(list.rivals.size());
-	for (const auto& entry : list.rivals) {
-		if (entry.id < 1) continue;
-		const uint64_t lastUpdateHint = CustomIRRivalLastUpdateHint(rivalDirectory, entry.id);
-
-		batch.tasks.push_back(RivalSyncTask{
-			.name = entry.name,
-			.result = std::async(std::launch::async, [ir, id = entry.id, lastUpdateHint]() -> std::optional<std::vector<openlr2::IRRivalScore>> {
-				std::vector<openlr2::IRRivalScore> scores;
-				const auto syncStatus = ir->SyncRivalScores(id, lastUpdateHint, scores);
-				if (syncStatus != openlr2::GetStatus::Ok) {
-					ErrorLogFmtAdd("SyncRivalScores %s for id=%d\n",
-						syncStatus == openlr2::GetStatus::Retry ? "retry ignored" : "failed",
-						id);
-					return std::nullopt;
-				}
-				return scores;
-			}),
-			.id = entry.id,
-		});
-	}
-
 	return batch;
 }
 
 bool CUSTOMIR_MANAGER::ApplyRivalSyncResults(RivalSyncBatch& sync) {
 	mRivalPaths.clear();
-	if (!sync.supported) return false;
-
 	cleanUpOldFutures(mDiscardedRivalSyncFutures);
 
-	const std::filesystem::path rivalDirectory = CustomIRRivalDirectory(sync.providerName);
-	int count = 0;
-	for (auto& task : sync.tasks) {
-		if (!task.result.valid()) continue;
+	bool displayListOk = false;
+	int displayCount = 0;
+	for (auto& provider : sync.providers) {
+		if (!provider.listOk) continue;
 
-		// Soft skip: do not block on unfinished downloads; park like discarded result-IR futures.
-		if (task.result.wait_for(std::chrono::milliseconds(0)) != std::future_status::ready) {
-			mDiscardedRivalSyncFutures.push_back(std::move(task.result));
-			continue;
+		const bool isDisplay = provider.providerName == mDisplayIr;
+		if (isDisplay) displayListOk = true;
+
+		const std::filesystem::path rivalDirectory = CustomIRRivalDirectory(provider.providerName);
+		int providerCount = 0;
+		for (auto& task : provider.tasks) {
+			if (!task.result.valid()) continue;
+
+			// Soft skip: do not block on unfinished downloads; park like discarded result-IR futures.
+			if (task.result.wait_for(std::chrono::milliseconds(0)) != std::future_status::ready) {
+				mDiscardedRivalSyncFutures.push_back(std::move(task.result));
+				continue;
+			}
+
+			std::optional<std::vector<openlr2::IRRivalScore>> scores = task.result.get();
+			if (!scores) continue;
+
+			if (!WriteCustomIRRivalArtifacts(rivalDirectory, task.id, task.name, *scores)) {
+				ErrorLogFmtAdd("CustomIR rival write failed for %s id=%d\n", provider.providerName.c_str(), task.id);
+				continue;
+			}
+
+			const std::filesystem::path rivalPath = rivalDirectory / std::to_string(task.id);
+			if (isDisplay) {
+				mRivalPaths.emplace_back(task.id, rivalPath);
+				++displayCount;
+			}
+			++providerCount;
+			ErrorLogFmtAdd("CustomIR rival ready: %s %d (%s, scores=%zu)\n",
+				provider.providerName.c_str(), task.id, task.name.c_str(), scores->size());
 		}
-
-		std::optional<std::vector<openlr2::IRRivalScore>> scores = task.result.get();
-		if (!scores) continue;
-
-		if (!WriteCustomIRRivalArtifacts(rivalDirectory, task.id, task.name, *scores)) {
-			ErrorLogFmtAdd("CustomIR rival write failed for id=%d\n", task.id);
-			continue;
-		}
-
-		const std::filesystem::path rivalPath = rivalDirectory / std::to_string(task.id);
-		mRivalPaths.emplace_back(task.id, rivalPath);
-		++count;
-		ErrorLogFmtAdd("CustomIR rival ready: %d (%s, scores=%zu)\n", task.id, task.name.c_str(), scores->size());
+		ErrorLogFmtAdd("CustomIR rival sync wrote %d rivals for %s%s\n",
+			providerCount, provider.providerName.c_str(), isDisplay ? " (display)" : "");
 	}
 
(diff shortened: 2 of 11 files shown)


MatVeiQaaa:
lol dont compile sqlite in the example project. Include the existing sqlite project. Make it depend on it. Link the .lib.

ExampleIR/ExternalIR.cpp
```diff
@@ -113,6 +115,41 @@ static std::string OLR2_IR_API GetWebRankingUrl(const char* songHash) {
...
+	const auto dbPath = rivalDir / (std::format("{}.db", rivalId));
+	sqlite3* db = nullptr;
+	if (sqlite3_open(dbPath.string().c_str(), &db) != SQLITE_OK) {
+		return openlr2::GetStatus::Fail;
+	}
+	sqlite3_exec(db,
+		"CREATE TABLE IF NOT EXISTS rival(hash TEXT primary key,r_clear INTEGER,r_totalnotes INTEGER,r_maxcombo INTEGER,r_perfect INTEGER,r_great INTEGER,r_good INTEGER,r_bad INTEGER,r_poor INTEGER,r_minbp INTEGER,r_option INTEGER,r_lastupdate INGEGER);"
+		"DELETE FROM rival;"
+		"INSERT OR REPLACE INTO rival VALUES('00000000000000000000000000000000',3,100,50,40,20,5,1,0,2,0,1);",
+		nullptr, nullptr, nullptr);
```
MatVeiQaaa: are you sure its a good idea to actually modify any data in the live database in the example module? Imo even reading is a stretch and unnecessary, but its understandable somewhat.

MatVeiQaaa:
i dont believe its really necessary for example module to have database access at all, are you sure this example brings enough value to justify this?

LR2/LR2_customir.cpp
```diff
@@ -90,19 +90,23 @@ class CustomIR {
 	openlr2::GetStatus RestoreCachedRank(const char* songHash, openlr2::IRRankResult& out);
 	openlr2::GetStatus GetGhost(const char* songHash, openlr2::GhostMode mode, int targetPlayerId, openlr2::IRGhostResult& out);
 	[[nodiscard]] std::string GetWebRankingUrl(const char* songHash);
+	openlr2::GetStatus GetRivals(openlr2::IRRivalListResult& out);
+	openlr2::GetStatus SyncRivalScores(int rivalId, int lastUpdateHint);
 	[[nodiscard]] const std::string& Name() const { return mName; };
 	[[nodiscard]] bool DidLoginSuccessfully() const { return mDidLoginSuccessfully; };
+	[[nodiscard]] std::filesystem::path RivalDir() const { return mDirectory / "Rival"; };
```
MatVeiQaaa: should OpenLR2 dictate the format of rival files and the directory in any way? I feel this should be entirely within the module implementation.

LR2/LR2_customir.cpp
```diff
@@ -301,6 +301,21 @@ bool CUSTOMIR_MANAGER::SyncRivals(int rivalsOut[20], int* rivalCountOut) {
...
+		if (GetMouseInput()) {
+			ClsDrawScreen();
+			printfDx("CustomIR rival sync skipped by mouse.\n");
+			ScreenFlip();
+			ProcessMessage();
+			return false;
+		}
+		std::this_thread::sleep_for(std::chrono::milliseconds(10));
+		ProcessMessage();
+	}
```
MatVeiQaaa: we just lost loading.bmp. This loop should be at main. Also is clsDx not necessary?

LR2/LR2_customir.cpp
```diff
@@ -317,12 +332,24 @@ bool CUSTOMIR_MANAGER::SyncRivals(int rivalsOut[20], int* rivalCountOut) {
...
+			entry.name.c_str(),
+			entry.id);
+		ScreenFlip();
+		ProcessMessage();
 		if (GetMouseInput()) {
+			ClsDrawScreen();
 			printfDx("CustomIR rival sync skipped by mouse.\n");
+			ScreenFlip();
+			ProcessMessage();
 			break;
```
MatVeiQaaa: do we need the first loop then, if we duplicate it here? The first loop saves no meaningful processing, i think it can be entirely removed.

ExampleIR/ExternalIR.cpp
luxurya111: I will remove all related examples from the current ExampleIR. I will think about whether there are other good alternatives.

LR2/LR2_customir.cpp
luxurya111: Using ClsDrawScreen was the cause. I have fixed it to clsDx.

ExampleIR/ExternalIR.cpp
luxurya111: @MatVeiQaaa IMO need to insert data into the live database to check the example, but if we don't need to make a full example, would it be okay to just create it as an empty function with only comment?

MatVeiQaaa: comment describing the idea would be perfectly enough, i believe. ExampleIR is not supposed to be a functional module in any significant way, or even demonstrate how the features end up looking in the game, it's entirely illustrative. It's only meant to define the skeleton.

luxurya111:
review is ready

chown2:
There's little demand for this feature so I'll be a bit slow to review this
I have a few problems with the current implementation:
- CUSTOMIR_MANAGER::SyncRivals should return a bunch of std::future instead of blocking by itself and rendering with DxLib. main() would then be able to display however it wants. Look for `Logging into internet ranking:` in the code. If you mimic how logging in is done, that will also fix some issues with loading.bmp.
- Since in future we want to be able to switch between different display IRs, we need to store the rivals tied to each custom IR. Probably right inside a class CustomIR, like mName. You can then, after fetching all rivals, copy from there into the buffer in NETWORK, like what NETWORK::GetRivalInfo currently does.
- CustomIR modules shouldn't be required to use sqlite themselves. Rivals themselves and their scores need to go through the our API. So SyncRivalScores should probably get a `std::vector<SomeScoreStruct>& out` parameter, the CustomIR module will fill it, and then the game itself will make appropriate .lr2folder and Rival/\<rival\>.db.

chown2:
Also here's how it's implemented in Raja for Tachi. Hm, doesn't look like it has a list of all scores. Maybe it used to have that in the past but now doesn't. https://github.com/zkldi/Tachi/blob/b3ad99cf1c3fded04b14953546b3470632aab39e/java/beatoraja-ir/src/main/java/bms/player/beatoraja/ir/TachiIR.java#L461

luxurya111:
image

> CUSTOMIR_MANAGER::SyncRivals should return a bunch of std::future instead of blocking by itself and rendering with DxLib. main() would then be able to display however it wants. Look for Logging into internet ranking: in the code. If you mimic how logging in is done, that will also fix some issues with loading.bmp.

I am not sure if the feature you want is what is shown in the image. If we use std::future, does it mean requests will be handled in batches like in the image?

The legacy behavior of the previous LR2IR wouldn't be awkward either, but implementing it with your requested design sounds good as well. For now, I will implement it so that even during a batch request, the request can be interrupted to immediately proceed to the song selection screen, and then I will submit the commit.

chown2:
@luxurya111 I'm good with what is in the picture! My point was about separating the rival fetching logic and displaying the progress. LR2_customir.cpp‎ shouldn't use any DxLib code (other than ErrorLogAdd/ErrorLogFmtAdd). 
Look at how I've solved the same async issue with CustomIR login:
https://github.com/GOMazk/OpenLR2/blob/d6c26d5108a8514bbaee53e5f2295ccfa17cc490/LR2/Main.cpp#L528-L554

luxurya111:
Now, db and lr2folder will be located in CustomIRRival/\<ModuleName\>.
Accordingly, the following requirements will allow the in-game display IR to refer to this folder when switching in the future.

> Since in future we want to be able to switch between different display IRs, we need to store the rivals tied to each custom IR. Probably right inside a class CustomIR, like mName. You can then, after fetching all rivals, copy from there into the buffer in NETWORK, like what NETWORK::GetRivalInfo currently does.

LR2/Main.cpp
```diff
@@ -552,11 +552,50 @@ int main(int argc, char** argv) {
 		}
 	}
 
+	gs.net.getRival = gs.config.network.getRival;
+	if (gs.net.getRival && gs.net.customIR.IsDisplayIrOnline()) {
+		auto rivalSync = gs.net.GetRivalInfo();
+		auto isRivalSyncReady = [&](const CUSTOMIR_MANAGER::RivalSyncTask& task) {
+			return isFutureReady(task.result);
+		};
```
chown2:
```suggestion
```
This helper doesn't add any clarity imo

LR2/LR2_songmanage.cpp
```diff
@@ -2840,12 +2856,31 @@ int LoadLR2CustomFolder(sqlite3 *sql, CONFIG_JUKEBOX *jb, CSTR scoreDBpath, char
...
-				if (jb->rival[i] < 1) break;
-				cstrSprintf(&jb->path[jb->numOfPath], fs::make_preferred("LR2files/Rival/%d.lr2folder").data(), jb->rival[i]);
-				GetFolderDataFromPath(jb->path[jb->numOfPath], sql);
-				jb->numOfPath++;
-				folderAddCount++;
+			if (CUSTOMIR_MANAGER::RivalPath(jb->rival[0])) {
+				for (int i = 0; i < 20; i++) {
+					if (jb->rival[i] < 1) break;
+					const auto rivalPath = CUSTOMIR_MANAGER::RivalPath(jb->rival[i]);
+					if (!rivalPath) continue;
```
chown2: Remove `if (CUSTOMIR_MANAGER::RivalPath(jb->rival[0])) {` and instead branch out here

LR2/LR2_customir.cpp
```diff
@@ -167,10 +173,191 @@ std::string CustomIR::GetWebRankingUrl(const char* songHash) {
...
+std::filesystem::path CustomIRRivalDirectory(const std::string& providerName) {
+	return std::filesystem::path(fs::make_preferred("LR2files/CustomIRRival").data()) / providerName;
+}
+
+int CustomIRRivalLastUpdateHint(const std::filesystem::path& rivalDirectory, int rivalId) {
+	const auto dbPath = rivalDirectory / std::format("{}.db", rivalId);
+	std::error_code ec;
+	if (!std::filesystem::exists(dbPath, ec)) {
+		return 0;
+	}
```
chown2: std::filesystem::exists(dbPath) here and the two new places isn't needed. You should instead pass the path to sqlite3_open like this: `file:<dbPath>?mode=ro`. Then sqlite itself will give an error if th efile doesn't exist.

LR2/LR2_customir.cpp
```diff
@@ -167,10 +173,191 @@ std::string CustomIR::GetWebRankingUrl(const char* songHash) {
...
+openlr2::GetStatus CustomIR::SyncRivalScores(int rivalId, int lastUpdateHint, std::vector<openlr2::IRRivalScore>& out) {
+	if (mMethods.SyncRivalScores == nullptr) return openlr2::GetStatus::Fail;
+	return mMethods.SyncRivalScores(rivalId, lastUpdateHint, out);
+}
+
 // All held std::future are automatically awaited on.
 // Can hang for up to 15 minutes if score sending fails time and time again...
+namespace {
+
+std::vector<std::pair<int, std::filesystem::path>> g_rivalPaths;
```
chown2: No new global variables. Looks like this should be a member of CUSTOMIR_MANAGER.

ExampleIR/ExternalIR.cpp
```diff
@@ -113,6 +113,19 @@ static std::string OLR2_IR_API GetWebRankingUrl(const char* songHash) {
...
+    out = {};
+    return openlr2::GetStatus::Fail;
+}
+
+static openlr2::GetStatus OLR2_IR_API SyncRivalScores(int /*rivalId*/, int /*lastUpdateHint*/, std::vector<openlr2::IRRivalScore>& /*out*/) {
+    // A real module would fetch or refresh one rival here and fill out with a complete score snapshot.
+    // OpenLR2 writes LR2files/CustomIRRival/<provider>/<rivalId>.db and .lr2folder from out.
+    return openlr2::GetStatus::Fail;
+}
+
```
chown2, line 140: I think Mat was calling out sqlite specifically. Now that we have an API with a list of 'openlr2::IRRivalScore', we can easily feed it with some data. Please do that.
The old GetRivals was good too
```
static openlr2::GetStatus GetRivals(openlr2::IRRivalListResult& out) {
	out = {};
	out.fetched_at = 1;
	out.rivals = {
		{ .id = 70100, .name = "ExampleRivalA" },
		{ .id = 70101, .name = "ExampleRivalB" },
	};
}
```

LR2/LR2_customir.cpp
```diff
@@ -268,6 +455,112 @@ bool CUSTOMIR_MANAGER::IsDisplayIrOnline() const {
...
+			scores = task.result.get();
+		}
+		catch (const std::exception& e) {
+			ErrorLogFmtAdd("CustomIR rival sync exception for id=%d: %s\n", task.id, e.what());
+			continue;
+		}
+		catch (...) {
+			ErrorLogFmtAdd("CustomIR rival sync unknown exception for id=%d\n", task.id);
+			continue;
+		}
```
chown2: So far, we haven't used exceptions in OpenLR2. If your code on the IR module side can throw, I suggest using try-catch in your module and returning GetStatus::Fail.
```suggestion
		scores = task.result.get();
```

LR2/LR2_customir.cpp
```diff
@@ -148,10 +154,188 @@ std::string CustomIR::GetWebRankingUrl(const char* songHash) {
...
+	std::string uri = "file:" + dbPath.generic_string() + "?mode=ro";
+	if (sqlite3_open(uri.c_str(), &db) != SQLITE_OK) {
+		return 0;
+	}
+
+	sqlite3_stmt* stmt = nullptr;
+	int lastUpdate = 0;
+	if (sqlite3_prepare_v2(db, "SELECT r_lastupdate FROM rival ORDER BY r_lastupdate DESC LIMIT 1", -1, &stmt, nullptr) == SQLITE_OK) {
+		if (sqlite3_step(stmt) == SQLITE_ROW) {
+			lastUpdate = sqlite3_column_int(stmt, 0);
```
chown2: sqlite3_column_int64 and uint64_t (check that's it's >= 0 first) for the timestamp, please, or this will fail after 2038 :D

LR2/LR2_customir.cpp
```diff
@@ -148,10 +154,188 @@ std::string CustomIR::GetWebRankingUrl(const char* songHash) {
...
+}
+
+openlr2::GetStatus CustomIR::SyncRivalScores(int rivalId, int lastUpdateHint, std::vector<openlr2::IRRivalScore>& out) {
+	if (mMethods.SyncRivalScores == nullptr) return openlr2::GetStatus::Fail;
+	return mMethods.SyncRivalScores(rivalId, lastUpdateHint, out);
+}
+
 // All held std::future are automatically awaited on.
 // Can hang for up to 15 minutes if score sending fails time and time again...
+namespace {
```
chown2: Personally, I prefer 'static' to anonymous namespaces, and it's what this file has been using so far. Can you make these function s'static' instead please?
Also, the comment `All held std::future are automatically awaited on.....` above got moved accidentally, it's supposed to be right above `CUSTOMIR_MANAGER::~CUSTOMIR_MANAGER() = default;`.

LR2/LR2_songmanage.cpp
```diff
@@ -2061,7 +2063,17 @@ int LoadFilteredBmsListFromDB(CSTR query, sqlite3 *sql, SONGSELECT *ss, int *dif
...
 		CSTR str;
-		cstrSprintf(&str, fs::make_preferred("ATTACH \'LR2files/Rival/%d.db\' AS rivaldb").data(), rivalID);
+		if (const auto rivalPath = CUSTOMIR_MANAGER::RivalPath(rivalID)) {
+			std::filesystem::path dbPath = *rivalPath;
+			dbPath += ".db";
+			std::string dbPathStr = dbPath.string();
+			for (char& ch : dbPathStr) {
+				if (ch == '\\') ch = '/';
+			}
+			cstrSprintf(&str, "ATTACH \'file:%s?mode=ro\' AS rivaldb", dbPathStr.c_str());
```
chown2: Can you make this more consistent with CustomIRRivalLastUpdateHint? The whole block can be something like
```cpp
	cstrSprintf(&str, "ATTACH \'file:%s?mode=ro\' AS rivaldb", std::filesystem::path{*rivalPath + / ".db"}.generic_string().c_str());
```

LR2/LR2_ir.cpp
```diff
@@ -691,6 +691,10 @@ int NETWORK::GetRivalInfo(int rivalID) {
 	return 1;
 }
 
+CUSTOMIR_MANAGER::RivalSyncBatch NETWORK::GetRivalInfo() {
+	return customIR.SyncRivals();
+}
+
```
chown2: Can we avoid this helper please
```suggestion
```

LR2/LR2_ir.cpp
```diff
@@ -955,7 +959,7 @@ int NETWORK::LR2IR_Login(int isDirectPlay) {
 
 		this->IR_ID = this->rankingData.myID;
 
-		if (isDirectPlay == 0) {
+		if (isDirectPlay == 0 && !customIR.IsDisplayIrOnline()) {
```
chown2: Why not?
If CustomIR and LR2IR storage for rivals (NETWORK::rivals?) conflict, it indicates a problem in our CustomIR API.
It gets in the way of switching the display IR at run-time too.
How hard would it be to make this new rival support not rely on NETWORK::rivals?
Note that in my view of perfect code we will eventually remove NETWORK completely, it's a horrendous tangled mess that is not fixable incrementally, so in the future we don't want any traces of it left.

LR2/LR2_customir.cpp
```diff
@@ -249,6 +433,101 @@ bool CUSTOMIR_MANAGER::IsDisplayIrOnline() const {
...
+				}
+				return scores;
+			}),
+		});
+	}
+
+	return batch;
+}
+
+bool CUSTOMIR_MANAGER::ApplyRivalSyncResults(RivalSyncBatch& sync, int rivalsOut[20], int* rivalCountOut) {
```
chown2: This line specifically may change after addressing my previous comment, but in future, take std::span instead of out array pointers

```suggestion
bool CUSTOMIR_MANAGER::ApplyRivalSyncResults(RivalSyncBatch& sync, std::span<int> rivalsOut) {
```

Also, we should avoid hard-coding '20' everywhere, we definitely want to support more rivals eventually

LR2/LR2_customir_api.h
```diff
@@ -243,6 +243,33 @@ enum class GetStatus: int {
 	Fail,
 };
 
+struct IRRivalInfo {
+	int id{};
+	std::string name;
+};
+
+// One rival score row. Field names match legacy LR2IR XML / Stellaverse JSON.
+struct IRRivalScore {
```
chown2, line 252: Some food for thought: conceptually, isn't this the same thing as IRRankPlayer, but with 'hash' instead of the player name?

LR2/LR2_ir.cpp
luxurya111: https://github.com/GOMazk/OpenLR2/pull/242/commits/813fe58f609e70fb0a0ddae11a1492235c74eba4

I am not sure if this is the exact part you intended, but I have made the modifications for now. Thank you.

LR2/LR2_customir_api.h
luxurya111: Conceptually agreed; both are essentially one score row, keyed by player vs chart hash. They’re close enough to unify later, but the call sites and legacy rival fields (option, int clear, etc.) make a merge risky in this PR. We’ll keep them separate for now and consider a shared score struct in a follow-up.

chown2:
I do not have the time to review this yet but I didn't forget about this. Being deeply involved with all of CustomIR code so far I would like to understand this code from top to bottom. :smiley: As a reminder, "there's little demand for this feature so I'll be a bit slow to review this".

chown2:
I have rebased this and added some review comments in a separate commit. If you are planning to make any changes, please, git fetch and git reset on the new commit first.

LR2/LR2_customir.h
```diff
@@ -47,28 +47,29 @@ class CUSTOMIR_MANAGER {
 	// \retval nullopt - Fail
 	std::optional<openlr2::IRGhostResult> TryGetTargetInfo(const char* songmd5, int mode, int targetPlayerId);
 
+	// XXX: we should support fetching several IRs at once.
```
luxurya111: Wouldn't this make OpenLR2's initial loading heavy? Or does it perform the initial load with display_ir and then synchronize rivals other than display_ir at an arbitrary point later on?

chown2: 1) updates are ideally incremental (CustomIRRivalLastUpdateHint) so we shouldn't usually be loading a lot of scores
2) each rival is loaded asynchronously (rivalSync.tasks), so it's only going to be as slow as the slowest rival to fetch
3) we can stop rival loading early (the `if (GetMouseInput()) {` branch)

LR2/LR2_customir.cpp
```diff
@@ -148,10 +154,202 @@ std::string CustomIR::GetWebRankingUrl(const char* songHash) {
...
+static std::filesystem::path CustomIRRivalDirectory(const std::string& providerName) {
+	return std::filesystem::path(fs::make_preferred("LR2files/CustomIRRival").data()) / providerName;
+}
+
+static uint64_t CustomIRRivalLastUpdateHint(const std::filesystem::path& rivalDirectory, int rivalId) {
+	const auto dbPath = rivalDirectory / std::format("{}.db", rivalId);
+
+	sqlite3* db = nullptr;
+	if(sqlite3_open_v2(std::format("file:{}?mode=ro", dbPath.generic_string()).c_str(), &db,
+					   SQLITE_OPEN_READONLY | SQLITE_OPEN_URI, nullptr)
```
chown2: probably
```suggestion
	if(sqlite3_open_v2(dbPath.generic_string(), &db, SQLITE_OPEN_READONLY, nullptr)
```

LR2/LR2_customir.cpp
```diff
@@ -336,11 +343,9 @@ std::optional<std::filesystem::path> CUSTOMIR_MANAGER::RivalPath(int rivalId) {
 }
 
 bool CUSTOMIR_MANAGER::CopyRivalIds(std::span<int> rivalsOut) {
-	std::ranges::fill(rivalsOut, 0);
-	if (sRivalPaths.empty() || rivalsOut.empty()) return false;
-	const std::size_t n = std::min(sRivalPaths.size(), rivalsOut.size());
-	for (std::size_t i = 0; i < n; ++i) {
-		rivalsOut[i] = sRivalPaths[i].first;
+	std::ranges::fill(rivalsOut, 0); // Fill the remainder if rivalsOut.size() < sRivalPaths.size()
```
luxurya111: Is that comparison direction intentional? As written, fill zeroes the whole span and the unused tail is when sRivalPaths is shorter than rivalsOut.

luxurya111: \+ It seems that the previously added empty guard has been removed. It looks like true will always be returned in Main.cpp, which uses this function—is this related to that comment?

chown2: 1) good point, a typo
```suggestion
	std::ranges::fill(rivalsOut, 0); // Fill the remainder if sRivalPaths.size() < rivalsOut.size()
```
2) also good point. given that this variable is only used in LoadLR2CustomFolder, which needs to have CUSTOMIR_MANAGER passed into it, we might as well not do this step at all, adding a getter into CUSTOMIR_MANAGER instead.

LR2/LR2_customir_api.h
```diff
@@ -274,6 +300,17 @@ struct MethodTable {
 	// This is called synchronously when F5 or the IR button is pressed in song-select.
 	// \retval "" - error or inapplicable.
 	std::string(OLR2_IR_API* GetWebRankingUrl)(char const* songHash) = nullptr;
+	// XXX: just some food for thought: should we introduce a `size_t MethodTableSize` here?
```
luxurya111, line 303: I don't think we need to consider it in the rival branch right now. If large-scale modifications such as LR2_customir_api.h are required in the future, it might be a good idea to fix it then.

PR #248, opened 2026-07-18
Dynamic skinnable panels

MatVeiQaaa:
WHY:

Assume you want a new feature - you code basic elements for it like `SRC_BUTTON`, some accessors for `SRC_NUMBER`, etc.
Now, you have made 50 new buttons and numbers. Unfortunately, there are only that many pixels on the screen, and that many button binds available, so UI flow must be extended, as you can't cram everything into the scene root. It makes most sense to have basically sub-scenes (panels), which will be able to store those elements. Much like `START` panel in select scene, which allows to draw a bunch of logically connected buttons in a single place, and alter what keybinds are gonna be doing.

HOW:

Adds `#PANEL` skinnable element. It's a non-rendering element, which has parameters 'master' and 'selectable'. 'master' parameter is a panel index which this panel will be assigned to as a sub-panel (slave). Indexes are assigned to panels in the order of their definition, in the range of 10-59. Panels will also have their own timers in the range of 500-599, where 500-549 is open timers, and 550-599 is close timers. Slave panels are not allowed to have slave panels themselves; only master panel ('master' parameter set to 0) index can be used as `master` param. Other skin elements can be bound to the panel through its panel and timer indexes, just like with existing panels.

Custom panel view can be opened by pressing `SELECT` while holding `START` (or the other way around, will see). This brings up the first master panel as well as its slave panels. You don't need to hold `SELECT` for the panel to stay open, so you can close it by pressing `START`. You can change which sub-panel (current master panel and its slaves) is active using scratch. This impacts which panel will be using button input on keys 1-7. Keys 1-7 are naively bound to first seven #SRC_BUTTON which use that panel's index as `panel` param. Panel can have more than 7 buttons, but anything past 7 will only be operable with the mouse. You can switch to next master panel by pressing `SELECT`, and previous by pressing `START` while holding `SELECT`. `#SRC_BUTTON` will likely be extended with new values for `type` param to have on-screen controls for custom panel flow.

USE CASE:

It allows us to only code new primitive skin elements, without having to worry about their GUI, leaving those things to skinners discretion, as well as have nice LR2-like flow to new features. For example, a panel to control unrandomizer could be made, with a sub-panel to select arrangement, and another sub-panel to control its toggles. There could be another panel to use other features of OpenLR2 like GAS, maybe some additional settings for GAS could be introduced too, and i'm sure there will be other gameplay features requiring GUI in the future. Settings menu could be expanded to control CustomIR with sub-sections to change more settings (there are a lot of settings currently not exposed at runtime which make sense to expose). Select and result scene could use a per-column judgement stats panel, which the game tracks the data for already, but lacks GUI. And anything else, really, anything at all. I believe this could be a future-proof solution for all our GUI needs.

Currently this is very early WIP, a draft PR is made already so contributors could follow the progress and make their input as it goes, so the development flow isn't based exclusively on my ideas, and before it grows too big for someone else to properly follow and impact.

### DIFF
--- a/LR2/LR2_panels.cpp
+++ b/LR2/LR2_panels.cpp
@@ -1,18 +1,35 @@
 #include "LR2_panels.h"
+#include "LR2_skindraw.h"
+#include "En_timer.h"
 #include <DxLib.h>
 #include <ranges>
 #include <optional>
 #include <algorithm>
 
-static auto is_pointer_legit = []<typename T>(const T* ptr, const std::vector<T>& arr) {
+static auto is_ptr_legit = []<typename T>(const T* ptr, const std::vector<T>& arr) {
 	if (ptr < arr.data()) return false;
 	if (ptr >= arr.data() + arr.size()) return false;
 	if ((reinterpret_cast<uintptr_t>(ptr) - reinterpret_cast<uintptr_t>(arr.data())) % sizeof(T) != 0) return false;
 	return true;
 };
 
-size_t PanelManager::GetIdx(Panel* ptr) {
-	if (is_pointer_legit(ptr, mPanels)) return ptr - mPanels.data();
+static auto is_idx_legit = []<typename T>(size_t idx, const std::vector<T>&arr) {
+	if (idx >= arr.size()) return false;
+	return true;
+};
+
+void PanelManager::Panel::Open() {
+	mIsActive = true;
+	mTimer = GetTimeWrap();
+}
+
+void PanelManager::Panel::Close() {
+	mIsActive = false;
+	mTimer = -1.;
+}
+
+size_t PanelManager::GetIdx(const Panel* ptr) {
+	if (is_ptr_legit(ptr, mPanels)) return ptr - mPanels.data();
 	return 0;
 }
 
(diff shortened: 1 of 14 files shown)


LR2/LR2_panels.cpp
```diff
@@ -0,0 +1,232 @@
...
+size_t PanelManager::GetIdx(const Panel* ptr) {
+	if (is_ptr_legit(ptr, mPanels)) return ptr - mPanels.data();
+	return 0;
+}
+
+bool PanelManager::IterateMaster(bool plus) {
+	if (mPanels.empty()) return false;
+	auto get_next = []<typename T>(T* now, std::vector<T>&arr, bool plus) {
+		auto iterate_plus = [&](T* now) {
+			if (now == arr.data() + arr.size() - 1) return &arr[0];
```
chown2, line 47: You should avoid pointer arithmetic IMO. Are you sure this is actually any easier to use than just using array indices?

MatVeiQaaa: tbh no, not really anymore, i see that i mostly end up using indices. I don't think the code would change logically if pointer iteration was replaced to index iteration.

PR #249, opened 2026-07-18
feat(main,customir): show IR login result while others are in progress 

chown2:
Show "Logged in" or "Failed to log in" immediately instead of first
displaying "done" and only then on a separate screen showing the status.

std::shared_future won't cut it in the `feat(customir): add rival
support` branch because the std::shared_future::get() method would force
us to copy the results as compared to std::future::get().
So this piece of code will be changed in that branch again. For now,
this is still an improvement.

@MatVeiQaaa look at what C++ standard library has

### DIFF
--- a/LR2/Main.cpp
+++ b/LR2/Main.cpp
@@ -11,7 +11,11 @@
 #include <chrono>
 #include <filesystem>
 #include <fstream>
+#include <future>
+#include <iterator>
+#include <string_view>
 #include <thread>
+#include <utility>
 
 #include <DxLib.h>
 extern "C" {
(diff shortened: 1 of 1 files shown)


MatVeiQaaa:
ooh, it only kills the future content after get() was called on all objects at least once?

chown2:
std::shared_future::get() returns a const reference to the data, while std::future::get() moves it out. It's fine here, but in `add rival support` we want to be able to do both. To be honest, I don't see why these are two separate classes in the standard library. If get() was overloaded on 'this' (const this& and this&&), it could fill all three use-cases described.

PR #254, opened 2026-07-21
fix(keyconfig): stop processing input on fadeout (LR2 bug)

chown2:
Don't let the user assign keys while on fadeout.

### DIFF
--- a/LR2/Scene06_Keyconfig.cpp
+++ b/LR2/Scene06_Keyconfig.cpp
@@ -42,6 +42,14 @@ int isBannedInput(int key) {
 	return 0;
 }
 int ProcI_Keyconfig(game *g) {
+	enum { PHASE_NORMAL = 1, PHASE_FADEOUT = 2 };
+	if (g->procPhase == PHASE_FADEOUT) {
+		if ( g->skstruct.fadeout < GetTimeLapse(2, &g->timer1) || g->skstruct.fadeout == 0 ) {
+			g->procSelecter = 2;
+		}
+		return 1;
+	}
+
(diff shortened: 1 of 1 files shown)


LR2/Scene06_Keyconfig.cpp
```diff
@@ -42,6 +42,14 @@ int isBannedInput(int key) {
 	return 0;
 }
 int ProcI_Keyconfig(game *g) {
+	enum { PHASE_NORMAL = 1, PHASE_FADEOUT = 2 };
```
MatVeiQaaa, line 45: for the love of god just define this in the structure.h. There are also stages 0 and 3. Init and exit.

chown2: It's not clear to me. It would make sense if each scene could have it's own list of phases. This documents phases of keyconfig specifically.

MatVeiQaaa: they are used consistently throughout the code though, as much as i have seen.

chown2: to be sure we better do what I did here to every other scene first

chown2: @MatVeiQaaa is that okay for now? I don't wanna go through every scene this very moment to meticulously verify how each of them handles phases

MatVeiQaaa: > is that okay for now?

If you believe there is a reason, sure

MatVeiQaaa (approved):

PR #257, opened 2026-07-22
feat(songselect): add support for beatoraja preview files

ainlorn:
song preview filenames are saved to db during initial chart parsing. a separate table is used for that. #PREVIEW directive in BMS file is used to determine the filename of the preview, or if it's not present the song directory is scanned for audio files with "preview" in name.
two new options have been added to the config file: one allows to ignore preview files (and always play the full previews), another determines the delay before preview playback.

### DIFF
--- a/LR2/En_fileutil.cpp
+++ b/LR2/En_fileutil.cpp
@@ -380,6 +380,8 @@ bool IsMediaFile(CSTR str) {
 		str.lower();
 		if (str.isSame(".mp3")) return true;
 		if (str.isSame(".wav")) return true;
+		if (str.isSame(".ogg")) return true;
+		if (str.isSame(".flac")) return true;
 		if (str.isSame(".avi")) return true;
 	}
 	return false;
@@ -391,6 +393,8 @@ bool IsSndFile(CSTR str) {
 		str.lower();
 		if (str.isSame(".mp3")) return true;
 		if (str.isSame(".wav")) return true;
+		if (str.isSame(".ogg")) return true;
+		if (str.isSame(".flac")) return true;
 	}
 	return false;
 }
--- a/LR2/LR2_songmanage.cpp
+++ b/LR2/LR2_songmanage.cpp
@@ -669,7 +669,7 @@ int EditTag(SONGDATA *song, sqlite3 *sql) {
 		SQL_Run(query, sql);
 		sqlite3_snprintf(1024, query, "INSERT INTO openlr2_preview (hash,previewpath) VALUES(\'%q\',\'%q\') ON CONFLICT(hash) DO UPDATE SET previewpath=excluded.previewpath",
 			meta.hash.body, meta.previewpath.body);
-		SQL_Run(query, sql);
+		sqlite3_exec(sql, query, nullptr, nullptr, nullptr);
 
 		if (meta.difficulty <= 0 || meta.difficulty > 5) {
 			SetUndefinedDifficulty(sql);
@@ -1377,7 +1377,7 @@ int SearchSongsFromPath(CSTR root, sqlite3 *sql, CSTR path) {
 					SQL_Run(str, sql);
 					sqlite3_snprintf(2048, str, "INSERT INTO openlr2_preview (hash,previewpath) VALUES(\'%q\',\'%q\') ON CONFLICT(hash) DO UPDATE SET previewpath=excluded.previewpath",
 						meta.hash.body, meta.previewpath.body);
-					SQL_Run(str, sql);
+					sqlite3_exec(sql, str, nullptr, nullptr, nullptr);
 					if (g_fullSongPass) g_processedSongPaths.insert(MakePathKey(searchPath));
 				}
 				count++;
@@ -1513,8 +1513,9 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb, ReloadProgr
 						LoadBMSMETAFromDB(&meta, sql);
 						SQL_Run(sqlite3_snprintf(1024, sBuf, "INSERT INTO song (hash,title,subtitle,genre,artist,subartist,level,date,path,folder,stagefile,banner,backbmp,parent,maxbpm,minbpm,random,longnote,judge,mode,bga,difficulty,favorite,type,txt,karinotes,adddate,exlevel) VALUES(\'%q\',\'%q\',\'%q\',\'%q\',\'%q\',\'%q\',%d,%d,\'%q\',\'%q\',\'%q\',\'%q\',\'%q\',\'%q\',%d,%d,%d,%d,%d,%d,%d,%d,0,0,%d,%d,%d,%d)",
 							meta.hash.body, meta.title.body, meta.subtitle.body, meta.genre.body, meta.artist.body, meta.subartist.body, meta.selLevel, newTime, str.body, AssignCRC32(meta.folderpath).body, meta.stagefilepath.body, meta.bannerpath.body, meta.backBMPpath.body,AssignCRC32(meta.parentfolderpath).body,meta.maxbpm,meta.minbpm,meta.random,meta.longnote,meta.judge,meta.keymode,meta.bga,meta.difficulty,(int)meta.hasTxt,meta.notecount,now,meta.exlevel), sql);
-						SQL_Run(sqlite3_snprintf(1024, sBuf, "INSERT INTO openlr2_preview (hash,previewpath) VALUES(\'%q\',\'%q\') ON CONFLICT(hash) DO UPDATE SET previewpath=excluded.previewpath",
-							meta.hash.body, meta.previewpath.body), sql);
+						sqlite3_snprintf(1024, sBuf, "INSERT INTO openlr2_preview (hash,previewpath) VALUES(\'%q\',\'%q\') ON CONFLICT(hash) DO UPDATE SET previewpath=excluded.previewpath",
+							meta.hash.body, meta.previewpath.body);
+						sqlite3_exec(sql, sBuf, nullptr, nullptr, nullptr);
 						if (g_fullSongPass) g_processedSongPaths.insert(MakePathKey(str));
 					}
 					else if (is_lr2folder) {
@@ -2291,8 +2292,8 @@ int LoadFilteredBmsListFromDB(CSTR query, sqlite3 *sql, SONGSELECT *ss, int *dif
 					sqlite3_stmt *pStmtPreview;
 					char previewQuery[512];
 					sqlite3_snprintf(512, previewQuery, "SELECT previewpath FROM openlr2_preview WHERE hash=\'%q\'", song.hash.body);
-					SQL_prepare(previewQuery, sql, &pStmtPreview);
-					if (sqlite3_step(pStmtPreview) == 100) {
+					sqlite3_prepare(sql, previewQuery, -1, &pStmtPreview, nullptr);
+					if (sqlite3_step(pStmtPreview) == SQLITE_ROW) {
 						song.previewfile = SQL_GetColumn(0, pStmtPreview);
 						if (song.previewfile.isDiff("(null)") && song.previewfile.length() > 4) song.isPreviewfile = true;
 					}
@@ -2779,7 +2780,7 @@ int LoadLR2CustomFolder(sqlite3 *sql, CONFIG_JUKEBOX *jb, CSTR scoreDBpath, char
 		SQL_Run("CREATE TABLE song(hash TEXT ,title TEXT ,subtitle TEXT ,genre TEXT,artist TEXT,subartist TEXT,tag TEXT ,path TEXT primary key ,type INTEGER,folder TEXT,stagefile TEXT,banner TEXT,backbmp TEXT,parent TEXT,level INTEGER,difficulty INTEGER,maxbpm INTEGER,minbpm INTEGER,mode INTEGER,judge INTEGER,longnote INTEGER,bga INTEGER,random INTEGER,date INTEGER,favorite INTEGER,txt INTEGER,karinotes INTEGER,adddate INTEGER,exlevel INTEGER)", sql);
 		SQL_Run("CREATE INDEX hashidx ON song (hash)", sql);
 		SQL_Run("CREATE INDEX parentidx ON song (parent)", sql);
-		SQL_Run("CREATE TABLE openlr2_preview(hash TEXT primary key, previewpath TEXT)", sql);
+		sqlite3_exec(sql, "CREATE TABLE openlr2_preview(hash TEXT primary key, previewpath TEXT)", nullptr, nullptr, nullptr);
 		SQL_Run("DROP TABLE course", sql);
 		SQL_Run("CREATE TABLE nonstop(id INTEGER primary key,title TEXT,line INTEGER,hash TEXT,ir INTEGER)", sql);
 		SQL_Run("CREATE TABLE expert(id INTEGER primary key,title TEXT,line INTEGER,hash TEXT,ir INTEGER)", sql);
--- a/LR2/LR2_songmanage.cpp
+++ b/LR2/LR2_songmanage.cpp
@@ -200,7 +200,7 @@ bool IsTextCommand(std::string_view line) {
 		StartsWithICaseAscii(line, "#STAGEFILE") ||
 		StartsWithICaseAscii(line, "#BANNER") ||
 		StartsWithICaseAscii(line, "#BACKBMP") ||
-        StartsWithICaseAscii(line, "#PREVIEW");
+		StartsWithICaseAscii(line, "#PREVIEW");
 }
 
 bool ParseTextCommand(BMSMETA& meta, CSTR& line) {
@@ -221,7 +221,7 @@ bool ParseTextCommand(BMSMETA& meta, CSTR& line) {
 	if (GetStringBodyStr(&line, "#STAGEFILE", &meta.stagefilepath)) return true;
 	if (GetStringBodyStr(&line, "#BANNER", &meta.bannerpath)) return true;
 	if (GetStringBodyStr(&line, "#BACKBMP", &meta.backBMPpath)) return true;
-    if (GetStringBodyStr(&line, "#PREVIEW", &meta.previewpath)) return true;
+	if (GetStringBodyStr(&line, "#PREVIEW", &meta.previewpath)) return true;
 	return false;
 }
 
@@ -337,11 +337,11 @@ void COPY_SONGDATA(SONGDATA *self, SONGDATA *other){
 	self->stagefile = other->stagefile;
 	self->banner = other->banner;
 	self->backBMP = other->backBMP;
-    self->previewfile = other->previewfile;
+	self->previewfile = other->previewfile;
 	self->isStagefile = other->isStagefile;
 	self->isBanner = other->isBanner;
 	self->isBackBMP = other->isBackBMP;
-    self->isPreviewfile = other->isPreviewfile;
+	self->isPreviewfile = other->isPreviewfile;
 	self->difficulty = other->difficulty;
 	self->level = other->level;
 	self->exlevel = other->exlevel;
@@ -396,7 +396,7 @@ int InitSongData(SONGDATA *song){
 	song->stagefile.fillzero();
 	song->banner.fillzero();
 	song->backBMP.fillzero();
-    song->previewfile.fillzero();
+	song->previewfile.fillzero();
 	song->folder.fillzero();
 	song->difficulty = 0;
 	song->folderType = 0;
@@ -411,7 +411,7 @@ int InitSongData(SONGDATA *song){
 	song->isStagefile = false;
 	song->isBanner = false;
 	song->isBackBMP = false;
-    song->isPreviewfile = false;
+	song->isPreviewfile = false;
 	song->replayExist = 0;
 	song->favorite = 0;
 	song->adddate = 0;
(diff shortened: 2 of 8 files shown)


LR2/structure.h
```diff
@@ -609,9 +612,11 @@ struct SONGDATA { /* 712bytes */
 	CSTR stagefile;
 	CSTR banner;
 	CSTR backBMP;
+    CSTR previewfile;
 	bool isStagefile{};
 	bool isBanner{};
 	int isBackBMP{};
+    bool isPreviewfile{};
```
chown2: 1) instead of two variables where one only makes sense in presense of the other, use `std::optional<CSTR> previewfile;`
2) How hard would it be to have it as a std::string instead of CSTR?

LR2/LR2_songmanage.cpp
```diff
@@ -2900,6 +2927,27 @@ int InitBMSMETA(BMSMETA *meta_) {
 	return 1;
 }
 
+static void tryFindPreviewFile(BMSMETA *meta) {
```
chown2: Replace the usage of this whole function with FindAltSound

LR2/LR2_songmanage.cpp
```diff
@@ -1365,6 +1375,9 @@ int SearchSongsFromPath(CSTR root, sqlite3 *sql, CSTR path) {
 					sqlite3_snprintf(2048, str, "INSERT INTO song (hash,title,subtitle,genre,artist,subartist,level,date,path,folder,stagefile,banner,backbmp,parent,maxbpm,minbpm,random,longnote,judge,mode,bga,difficulty,favorite,type,txt,karinotes,adddate,exlevel) VALUES(\'%q\',\'%q\',\'%q\',\'%q\',\'%q\',\'%q\',%d,%d,\'%q\',\'%q\',\'%q\',\'%q\',\'%q\',\'%q\',%d,%d,%d,%d,%d,%d,%d,%d,0,0,%d,%d,%d,%d)",
 						meta.hash.body, meta.title.body, meta.subtitle.body, meta.genre.body, meta.artist.body, meta.subartist.body, meta.selLevel, filetime, searchPath.body, AssignCRC32(meta.folderpath).body, meta.stagefilepath.body, meta.bannerpath.body, meta.backBMPpath.body, AssignCRC32(meta.parentfolderpath).body, meta.maxbpm, meta.minbpm, meta.random, meta.longnote, meta.judge, meta.keymode, meta.bga, meta.difficulty, (int)meta.hasTxt, meta.notecount, now, meta.exlevel);
 					SQL_Run(str, sql);
+                    sqlite3_snprintf(2048, str, "INSERT INTO openlr2_preview (hash,previewpath) VALUES(\'%q\',\'%q\') ON CONFLICT(hash) DO UPDATE SET previewpath=excluded.previewpath",
+                         meta.hash.body, meta.previewpath.body);
+                    SQL_Run(str, sql);
```
chown2: Replace SQL_Run usage with sqlite3_exec and SQL_prepare with sqlite3_prepare

https://github.com/GOMazk/OpenLR2/blob/3d0e42550314ae756414aaa10bb3342392bd7de0/LR2/En_dbio.cpp#L3-L11

LR2/LR2_songmanage.cpp
```diff
@@ -2273,7 +2288,17 @@ int LoadFilteredBmsListFromDB(CSTR query, sqlite3 *sql, SONGSELECT *ss, int *dif
 					thisFolder = SQL_GetColumn(9, pStmt);
 					song.folder = thisFolder;
 
-					if (song.keymode > 0) {
+                    sqlite3_stmt *pStmtPreview;
+                    char previewQuery[512];
+                    sqlite3_snprintf(512, previewQuery, "SELECT previewpath FROM openlr2_preview WHERE hash=\'%q\'", song.hash.body);
+                    SQL_prepare(previewQuery, sql, &pStmtPreview);
+                    if (sqlite3_step(pStmtPreview) == 100) {
```
chown2: Replace 100 with the named constant (probably either SQLITE_OK or SQLITE_ROW)

LR2/LR2_songmanage.cpp
```diff
@@ -1500,7 +1513,9 @@ int ReloadSongsByQuery(CSTR query, sqlite3 *sql, CONFIG_JUKEBOX *jb, ReloadProgr
...
 							meta.hash.body, meta.title.body, meta.subtitle.body, meta.genre.body, meta.artist.body, meta.subartist.body, meta.selLevel, newTime, str.body, AssignCRC32(meta.folderpath).body, meta.stagefilepath.body, meta.bannerpath.body, meta.backBMPpath.body,AssignCRC32(meta.parentfolderpath).body,meta.maxbpm,meta.minbpm,meta.random,meta.longnote,meta.judge,meta.keymode,meta.bga,meta.difficulty,(int)meta.hasTxt,meta.notecount,now,meta.exlevel), sql);
-						if (g_fullSongPass) g_processedSongPaths.insert(MakePathKey(str));
+                        SQL_Run(sqlite3_snprintf(1024, sBuf, "INSERT INTO openlr2_preview (hash,previewpath) VALUES(\'%q\',\'%q\') ON CONFLICT(hash) DO UPDATE SET previewpath=excluded.previewpath",
+                            meta.hash.body, meta.previewpath.body), sql);
+                        if (g_fullSongPass) g_processedSongPaths.insert(MakePathKey(str));
```
chown2: Stray change caused by usage of wrong indentation. All of your edits use spaces, while the whole LR2 folder uses tabs. Your text editor seems to ignore .editorconfig.

chown2 (commented):
Thanks. Personally, I think #PREVIEW promotes bad ideas, and we should instead work on improving soft-preview, but this looks like a simple implementation useful while soft-preview support is not good.

chown2:
Also, add `#PREVIEW` to `new_command(bms).md`, please.

LR2/Scene02_Songselect.cpp
```diff
@@ -1670,9 +1670,42 @@ static void ThreadProc_RankingAutoUpdate(game* g) {
...
+        g->gameplay.previewStatus = 0;
+        g->gameplay.isPreviewLoad = 0;
+    }
+}
+
 static void ThreadProc_LoadPreview(game *g) {
 	BMSMETA meta;
 
+    if (!g->config.select.ignorePreviewFiles) {
+        SONGDATA *song = &g->sSelect.bmsList[g->sSelect.cur_song];
```
```suggestion
        SONGDATA &song = g->sSelect.bmsList[g->sSelect.cur_song];
```
In new code never have variables as pointers if you always dereference them. Use references. Also applies to at least LoadPreviewFile.

LR2/LR2_songmanage.cpp
ainlorn: Sadly, won't work. Preview files can be named anything starting with "preview". For example, BmsPreviewAudioGenerator names files "preview_auto_generator.ogg" by default.
FindAltSound seems to only match files by full name.

ainlorn:
@chown2 i think i fixed everything

ainlorn:
nevermind, found a bug. for some reason it only works when sorting by directory
(fixed now)

LR2/LR2_songmanage.cpp
```diff
@@ -220,6 +222,10 @@ bool ParseTextCommand(BMSMETA& meta, CSTR& line) {
 	if (GetStringBodyStr(&line, "#STAGEFILE", &meta.stagefilepath)) return true;
 	if (GetStringBodyStr(&line, "#BANNER", &meta.bannerpath)) return true;
 	if (GetStringBodyStr(&line, "#BACKBMP", &meta.backBMPpath)) return true;
+	if (GetStringBodyStr(&line, "#PREVIEW", &buf)) {
```
chown2: make a GetStringBodyStr alternative that would take in an std::string. A 'static' function would be fine for now.
`bool GetStringBodyStr(std::string_view str, std::string_view head, std::string& out)`. For the comparison you can use StartsWithICaseAscii from the same file.

LR2/LR2_songmanage.cpp
```diff
@@ -3040,6 +3093,13 @@ int ParseBMSMETA(BMSMETA *meta, CSTR filepath, char flag) {
 	dir.add("*.txt");
 	if (IsFileExist(dir)) meta->hasTxt = true;
 	meta->notecount = notes;
+
+	CSTR previewFullPath;
+	previewFullPath.assign(meta->folderpath).add(meta->previewpath.c_str());
+	if (meta->previewpath.length() == 0 || !IsFileExist(previewFullPath)) {
+		tryFindPreviewFile(meta);
+	}
```
chown2, line 3072: Not a good idea IMO. This means it won't even work unelss you reset your song database completely. I think it's better to move this into ThreadProc_LoadPreview.
EDIT: actually, it probably works because parsebmsmeta is called in select on each chart again and again instead of just using values from the DB. But relying on this seems fragile.

LR2/LR2_songmanage.cpp
```diff
@@ -2273,6 +2291,18 @@ int LoadFilteredBmsListFromDB(CSTR query, sqlite3 *sql, SONGSELECT *ss, int *dif
...
 
+					sqlite3_stmt *pStmtPreview;
+					char previewQuery[512];
+					sqlite3_snprintf(512, previewQuery, "SELECT previewpath FROM openlr2_preview WHERE hash=\'%q\'", song.hash.body);
+					sqlite3_prepare(sql, previewQuery, -1, &pStmtPreview, nullptr);
+					if (sqlite3_step(pStmtPreview) == SQLITE_ROW) {
+						song.previewfile = SQL_GetColumn(0, pStmtPreview);
+						if (song.previewfile.value() == "(null)" || song.previewfile.value().length() <= 4) {
+							song.previewfile = {};
+						}
```
chown2: Better avoid going through a CSTR caused by SQL_GetColumn. The '(null)' check tells exactly why. Then this second if check isn't needed at all too. Hm, we shouldn't want this field nullable in the database.
```suggestion
						song.previewfile = std::string(sqlite3_column_text(pStmtPreview, 0), sqlite3_column_bytes(pStmtPreview, 0));
```

LR2/LR2_songmanage.cpp
```diff
@@ -2754,6 +2784,7 @@ int LoadLR2CustomFolder(sqlite3 *sql, CONFIG_JUKEBOX *jb, CSTR scoreDBpath, char
 		SQL_Run("CREATE TABLE song(hash TEXT ,title TEXT ,subtitle TEXT ,genre TEXT,artist TEXT,subartist TEXT,tag TEXT ,path TEXT primary key ,type INTEGER,folder TEXT,stagefile TEXT,banner TEXT,backbmp TEXT,parent TEXT,level INTEGER,difficulty INTEGER,maxbpm INTEGER,minbpm INTEGER,mode INTEGER,judge INTEGER,longnote INTEGER,bga INTEGER,random INTEGER,date INTEGER,favorite INTEGER,txt INTEGER,karinotes INTEGER,adddate INTEGER,exlevel INTEGER)", sql);
 		SQL_Run("CREATE INDEX hashidx ON song (hash)", sql);
 		SQL_Run("CREATE INDEX parentidx ON song (parent)", sql);
+		sqlite3_exec(sql, "CREATE TABLE openlr2_preview(hash TEXT primary key, previewpath TEXT)", nullptr, nullptr, nullptr);
```
chown2: NOT NULL + STRICT. Before doing this, make a copy of your song.db and make sure STRICT tables don't cause any problems in older SQLite versions, see if LR2 still works. If STRICT doesn't work in older version or requires some PRAGMA just remove it.
```suggestion
		sqlite3_exec(sql, "CREATE TABLE openlr2_preview(hash TEXT PRIMARY KEY NOT NULL, previewpath TEXT NOT NULL) STRICT", nullptr, nullptr, nullptr);
```

LR2/LR2_songmanage.cpp
chown2: This function is too long for what it does and a has a few problems:
- shouldn't take in a fat object like BMSMETA (btw, should have been a reference instead of a pointer). Should have taken what's in meta->folderpath and returned what's written into meta->previewpath.
- std::filesystem::exists should be replaced with std::filesystem::directory_iterator(..., ec). Without a 'ec' this would have killed the game if the the path was a file or if there were no permissions to read the folder.
- the tolower call is UB, welcome to C++! check cppreference. We need a new ToLowerAscii analogous to ToUpperAscii.
- the checks `starts_with("preview") ... and ...ends_with('.mp3')...` are too specific and file types are sprawled around.

Let me think how to make this better.

LR2/Scene02_Songselect.cpp
```diff
@@ -1670,9 +1670,42 @@ static void ThreadProc_RankingAutoUpdate(game* g) {
...
 	BMSMETA meta;
 
+	if (!g->config.select.ignorePreviewFiles) {
+		SONGDATA &song = g->sSelect.bmsList[g->sSelect.cur_song];
+		if (song.previewfile) {
+			const auto previewpath = std::filesystem::path(song.filepath.getDirectory().body) / song.previewfile.value();
+			if (std::filesystem::exists(previewpath)) {
+				LoadPreviewFile(g, previewpath.string());
+				return;
+			}
```
chown2, line 1717: Personally, I think what's this 'if' check does should've been inside LoadPreviewFile. Also, LoadPreviewFile should've probably accepted std::filesystem:path directly instead of a std::string.

ainlorn: better to keep it here, so that if the file is not present we can still fallback to full preview

LR2/LR2_songmanage.cpp
ainlorn: STRICT seems to crash LR2

LR2/LR2_songmanage.cpp
ainlorn: used StartsWithICaseAscii and IsSndFile instead of starts_with, ends_with

LR2/LR2_songmanage.cpp
ainlorn: Added this logic to ThreadProc_LoadPreview. ParseBMSMETA doesn't actually seem to be used when selecting songs so a full db refresh was indeed required prior to this change.

LR2/LR2_songmanage.cpp
ainlorn: done

LR2/En_fileutil.cpp
```diff
@@ -380,6 +380,8 @@ bool IsMediaFile(CSTR str) {
 		str.lower();
 		if (str.isSame(".mp3")) return true;
 		if (str.isSame(".wav")) return true;
+		if (str.isSame(".ogg")) return true;
+		if (str.isSame(".flac")) return true;
```
chown2, line 384: this is unreachable, see `str.lastCut(4);` above.
It's probably best to reimplement this function in terms of std::string
roughly
```cpp
bool IsMediaFile(CSTR str) {
  auto tolower_ascii = [](char c) { if(c >= 'A' && c <= 'Z') c += 'a' - 'A'; return c; };
  std::string s = str.body;
  for (char& c : s) c = tolower_ascii(c);
  if (s.ends_with(".mp3")) return true;
  ...
  return false;
}
```

LR2/En_fileutil.cpp
```diff
@@ -391,6 +393,8 @@ bool IsSndFile(CSTR str) {
...
```
chown2, line 397: same, lastCut(4) makes this unreachable

LR2/LR2_songmanage.cpp
```diff
@@ -199,10 +207,12 @@ bool IsTextCommand(std::string_view line) {
 		StartsWithICaseAscii(line, "#COMMAND") ||
 		StartsWithICaseAscii(line, "#STAGEFILE") ||
 		StartsWithICaseAscii(line, "#BANNER") ||
-		StartsWithICaseAscii(line, "#BACKBMP");
+		StartsWithICaseAscii(line, "#BACKBMP") ||
+		StartsWithICaseAscii(line, "#PREVIEW");
 }
 
 bool ParseTextCommand(BMSMETA& meta, CSTR& line) {
+	CSTR buf;
```
chown2, line 215:
```suggestion
```
Not used

LR2/LR2_songmanage.cpp
chown2: The tryFindPreviewFile block in this function is not needed anymore then. Better keep 'metadata' conceptually clean, only inferred from the .bms file itself. The `meta->hasTxt` right above is not nice and not something we should repeat.

LR2/Scene02_Songselect.cpp
chown2: Then make LoadPreviewFile return a bool, move the fs::exists check inside it, and here check the return value of LoadPreviewFile

LR2/Scene02_Songselect.cpp
```diff
@@ -1672,9 +1672,52 @@ static void ThreadProc_RankingAutoUpdate(game* g) {
...
+		SONGDATA &song = g->sSelect.bmsList[g->sSelect.cur_song];
+		std::filesystem::path folderpath = song.filepath.getDirectory().body;
+
+		if (!song.previewfile) {
+			auto foundFile = openlr2::tryFindPreviewFile(folderpath);
+			if (foundFile) {
+				song.previewfile = foundFile.value();
+				openlr2::updateSongPreview(song.hash.body, song.previewfile.value(), sql);
+			}
+		}
```
chown2, line 1710: Expands on my comment about BMSMETA purity, don't write back to the DB in here

LR2/LR2_songmanage.cpp
```diff
@@ -172,6 +172,14 @@ bool GetRawBodyInt(std::string_view line, std::string_view head, int& out) {
 	return true;
 }
 
+bool GetStringBodyStr(std::string_view str, std::string_view head, std::string& out) {
+	if (StartsWithICaseAscii(str, head) && str.size() > head.size()) {
+		out = str.substr(head.length() + 1);
```
chown2, line 177:
```suggestion
	if (StartsWithICaseAscii(str, head) && str.size() > head.size()) {
		out = str.substr(head.size() + 1);
```
For consistency

LR2/LR2_songmanage.cpp
```diff
@@ -2269,6 +2286,15 @@ int LoadFilteredBmsListFromDB(CSTR query, sqlite3 *sql, SONGSELECT *ss, int *dif
 					thisFolder = SQL_GetColumn(9, pStmt);
 					song.folder = thisFolder;
 
+					sqlite3_stmt *pStmtPreview;
+					char previewQuery[512];
+					sqlite3_snprintf(512, previewQuery, "SELECT previewpath FROM openlr2_preview WHERE hash=\'%q\'", song.hash.body);
+					sqlite3_prepare(sql, previewQuery, -1, &pStmtPreview, nullptr);
+					if (sqlite3_step(pStmtPreview) == SQLITE_ROW && sqlite3_column_bytes(pStmtPreview, 0) > 0) {
+						song.previewfile = reinterpret_cast<const char *>(sqlite3_column_text(pStmtPreview, 0));
+					}
```
chown2, line 2295: Does this work?
1) I don't see why avoid assigning the value if string length is 0, it's essentially a no-op at worst
2) use sqlite3_column_bytes to avoid a strlen
```suggestion
					if (sqlite3_step(pStmtPreview) == SQLITE_ROW) {
						song.previewfile = std::string_view(reinterpret_cast<const char *>(sqlite3_column_text(pStmtPreview, 0)), sqlite3_column_bytes(pStmtPreview, 0));
					}
```

LR2/LR2_songmanage.cpp
```diff
@@ -3048,3 +3083,22 @@ int openlr2::adjustFilterKey(CONFIG_SELECT const& cfg_select, int key) {
...
+		}
+	}
+	return {};
+}
+
+void openlr2::updateSongPreview(const std::string& hash, const std::string& previewpath, sqlite3* sql) {
+	char query[512];
+
+	sqlite3_snprintf(512, query, "INSERT INTO openlr2_preview (hash,previewpath) VALUES(\'%q\',\'%q\') ON CONFLICT(hash) DO UPDATE SET previewpath=excluded.previewpath",
+		hash.c_str(), previewpath.c_str());
```
chown2, line 3102: Does this work? You can then take in 'string_view' as arguments instead. Avoids strlen calls too. 
```suggestion
	sqlite3_snprintf(512, query, "INSERT INTO openlr2_preview (hash,previewpath) VALUES(\'%.*q\',\'%.*q\') ON CONFLICT(hash) DO UPDATE SET previewpath=excluded.previewpath",
		static_cast<int>(hash.size()), hash.data(), static_cast<int>(previewpath.size()), previewpath.data());
```

LR2/LR2_songmanage.cpp
ainlorn: > I don't see why avoid assigning the value if string length is 0, it's essentially a no-op at worst

because it's std::optional

LR2/Scene02_Songselect.cpp
```diff
@@ -1672,9 +1672,52 @@ static void ThreadProc_RankingAutoUpdate(game* g) {
...
+	g->gameplay.flag_closingPhase = 0;
+	StopSound(&g->audio, &g->sSelect.previewSound);
+
+	const auto prevFile = g->sSelect.previewSound.filename.body;
+	if (!g->sSelect.previewSound.load || prevFile == nullptr || previewpath.string() != prevFile) {
+		ReleaseSound(&g->audio, &g->sSelect.previewSound);
+		if (g->gameplay.flag_closingPhase == 0 && g->procSelecter == 2 && g->gameplay.previewStatus == 1) {
+			LoadSound(&g->audio, &g->sSelect.previewSound, previewpath.string().c_str(), true, false, true);
+		}
+	}
```
chown2, line 1685: LoadSound has the same checks inside itself
```suggestion
	if (g->gameplay.flag_closingPhase == 0 && g->procSelecter == 2 && g->gameplay.previewStatus == 1) {
		LoadSound(&g->audio, &g->sSelect.previewSound, previewpath.string().c_str(), true, false, true);
	}
```

chown2: Also, I'd argue `g->gameplay.flag_closingPhase == 0 && g->procSelecter == 2 && g->gameplay.previewStatus == 1` checks, neither here not just below, are actually any useful.
They don't actually avoid any thread unsafety.
And they don't make it just appear to work, either.
Because in here, if two such threads manage to run at once, we will reach the 'else' case where preview is disabled.
So I propose removing both of them in here.

LR2/LR2_songmanage.cpp
chown2:
```suggestion
					if (sqlite3_step(pStmtPreview) == SQLITE_ROW) {
						auto s = std::string_view(reinterpret_cast<const char *>(sqlite3_column_text(pStmtPreview, 0)), sqlite3_column_bytes(pStmtPreview, 0));
						if (s.size() != 0) {
							song.previewfile = s;
						} else {
							ErrorLogFmtAdd("BUG: empty preview path in the DB on song=%s\n", song.hash.body);
						}
					}
```

chown2: > because it's std::optional

It's not a normally expected case then

PR #263, opened 2026-07-25
Use async load functions on bga loading

m8nscape:
This fixes crashes when playing a chart with image BGA on dx11 mode. Details:
- Disable parallel loading (a setting in LR2.exe) : ok
- Enable parallel loading, video bga: ok
- Enable parallel loading, image bga: crash
- Enable parallel loading, image bga, bga off: ok

It won't crash in dx9 but well I just leave it here
I've seen the panels on the screen flickers when loading the chart on vanilla LR2. Happens rarely. I reckon it is affected by this too

ref: https://dxlib.xsrv.jp/cgi/patiobbs/patio.cgi?mode=view&no=3929

also CONFIG_SYSTEM::isablebmsthread is a very misleading typo, pls consider rename

### DIFF
--- a/LR2/LR2_bmsload.cpp
+++ b/LR2/LR2_bmsload.cpp
@@ -788,6 +788,7 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
 			if (!anyLoading) break;
 			std::this_thread::sleep_for(std::chrono::milliseconds(10));
 		}
+	}
 
 	SetTransColor(Rtmp, Gtmp, Btmp);
 #ifdef _WIN32
--- a/LR2/LR2_bmsload.cpp
+++ b/LR2/LR2_bmsload.cpp
@@ -768,33 +768,26 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
 				bgaHandleLoaded[i] = false;
 			}
 		}
-		bool finishedLoading = false;
-		while (!finishedLoading) {
-			finishedLoading = true;
+		while (true) {
+			bool anyLoading = false;
 			for (auto& [i, loaded] : bgaHandleLoaded) {
-				if (!loaded && gp->bgaHandle[i] != -1) {
-					int loadResult = CheckHandleASyncLoad(gp->bgaHandle[i]);
-					if (loadResult == FALSE) {
-						// FALSE: finished
-						loaded = true;
-						gp->loadObject_loaded++;
-					} else if (loadResult == TRUE) {
-						// TRUE: still loading
-						finishedLoading = false;
-						break;
-					} else {
-						// -1: error
-						// Release the handle and do not care about it anymore
-						DeleteGraph(gp->bgaHandle[i]);
-						gp->bgaHandle[i] = -1;
-					}
+				if (loaded) continue;
+				int loadResult = CheckHandleASyncLoad(gp->bgaHandle[i]);
+				if (loadResult == TRUE/*still loading*/) {
+					anyLoading = true;
+					continue;
 				}
+				if (loadResult != FALSE/*finished*/) {
+					// Error
+					DeleteGraph(gp->bgaHandle[i]);
+					gp->bgaHandle[i] = -1;
+				}
+				loaded = true;
+				gp->loadObject_loaded++;
 			}
-			if (!finishedLoading) {
-				std::this_thread::sleep_for(std::chrono::milliseconds(10));
-			}
+			if (!anyLoading) break;
+			std::this_thread::sleep_for(std::chrono::milliseconds(10));
 		}
-	}
 
 	SetTransColor(Rtmp, Gtmp, Btmp);
 #ifdef _WIN32
--- a/LR2/LR2_bmsload.cpp
+++ b/LR2/LR2_bmsload.cpp
@@ -776,7 +776,7 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
 				}
 			}
 			if (!finishedLoading) {
-				std::this_thread::sleep_for(std::chrono::milliseconds(100));
+				std::this_thread::sleep_for(std::chrono::milliseconds(10));
 			}
 		}
 	}
--- a/LR2/LR2_bmsload.cpp
+++ b/LR2/LR2_bmsload.cpp
@@ -742,7 +742,11 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
 			else {
 				gp->bgaHandle[i] = LoadGraph(gp->BMP_filename[i]);	// Note: When using SetUseASyncLoadFlag, LoadGraph returns the handle immediately. Further check is required
 			}
-			gp->loadObject_loaded++;
+
+			// Count only after load finish. Only add if using synchronous load here
+			if (cfg->system.isablebmsthread != 0) {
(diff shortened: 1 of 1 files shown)


LR2/LR2_bmsload.cpp
```diff
@@ -722,34 +722,71 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
...
+					} else {
+						// -1: error
+						// Release the handle and do not care about it anymore
+						DeleteGraph(gp->bgaHandle[i]);
+						gp->bgaHandle[i] = -1;
+					}
+				}
+			}
+			if (!finishedLoading) {
+				std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
chown2: 100ms might be a bit too much. How about making this 10ms?

LR2/LR2_bmsload.cpp
```diff
@@ -722,34 +722,71 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
 	}
 	
 #ifdef _WIN32
-	if (cfg->system.isablebmsthread == 0) CoInitialize(NULL);
+	if (cfg->system.isablebmsthread == 0) {
+		CoInitialize(NULL);
+		SetUseASyncLoadFlag(TRUE);
```
chown2, line 730: Needs a comment like "We are doing this on another thread, but DirectX requires resouces to be loaded from the main thread. This will make DxLib do it itself on the main thread. This fixes crashes in DX11."

chown2:
I like the approach if this works.
I've tried achieving this with LoadSoftImage (which is how you would do it with any sane graphics library), but IIRC it didn't support videos, and there was no matching video API.
Does this cause any stutters while loading? Creating a few thousand loading tasks at once sounds a bit scary.

MatVeiQaaa:
CONFIG_SYSTEM::isablebmsthread should be removed entirely in the future, there is no need for such toggle.

LR2/LR2_bmsload.cpp
m8nscape: this is not a cheap check (SLOTS=38440). I think 10ms is too short

LR2/LR2_bmsload.cpp
m8nscape: I'm not sure about that. There is no documents describe what dxlib is actually doing inside. On the other hand DX11 itself claims it supports resouces to be loaded in multithread...
https://learn.microsoft.com/en-us/windows/win32/direct3d11/overviews-direct3d-11-render-multi-thread-create

m8nscape:
> Does this cause any stutters while loading? Creating a few thousand loading tasks at once sounds a bit scary.

not noticable on my pc, might need more samples

LR2/LR2_bmsload.cpp
MatVeiQaaa: > this is not a cheap check (SLOTS=38440). I think 10ms is too short

Entire LR2 loop is less than 10ms by multitudes. Don't worry about it.

LR2/LR2_bmsload.cpp
MatVeiQaaa: > I'm not sure about that. There is no documents describe what dxlib is actually doing inside. On the other hand DX11 itself claims it supports resouces to be loaded in multithread... https://learn.microsoft.com/en-us/windows/win32/direct3d11/overviews-direct3d-11-render-multi-thread-create

DX11 only supports resources to load async if you use async interfaces to do that. We don't deal with DirectX directly, it's all abstracted away by DxLib. You can't change device state async, as a common rule of any GPU interaction.

DxLib doesn't tell you anything, but we got it's entire source tree in the codebase and we go as far as modifying it already. What it does at any point is known if required.

m8nscape: oh ok, comments added

MatVeiQaaa (approved):

LR2/LR2_bmsload.cpp
```diff
@@ -721,35 +721,75 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
...
 				SetTransColor(0, 255, 0);
-				gp->bgaHandle[i] = LoadGraph(gp->BMP_filename[i]);
+				gp->bgaHandle[i] = LoadGraph(gp->BMP_filename[i]);	// Note: When using SetUseASyncLoadFlag, LoadGraph returns the handle immediately. Further check is required
 				SetTransColor(0, 0, 0);
 			}
 			else {
-				gp->bgaHandle[i] = LoadGraph(gp->BMP_filename[i]);
+				gp->bgaHandle[i] = LoadGraph(gp->BMP_filename[i]);	// Note: When using SetUseASyncLoadFlag, LoadGraph returns the handle immediately. Further check is required
 			}
 			gp->loadObject_loaded++;
```
chown2: This should be moved into the loop below now

m8nscape: updated

LR2/LR2_bmsload.cpp
```diff
@@ -721,35 +721,87 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
...
+						DeleteGraph(gp->bgaHandle[i]);
+						gp->bgaHandle[i] = -1;
+					}
+				}
+			}
+			if (!finishedLoading) {
+				std::this_thread::sleep_for(std::chrono::milliseconds(10));
+			}
+		}
+	}
```
chown2: This is easier to read and fixes two bugs: loadObject_loaded didn't increment if there was at least one preceding in-progress loading, and didn't ever increment on error results.
```suggestion
		while (true) {
			bool anyLoading = false;
			for (auto& [i, loaded] : bgaHandleLoaded) {
				if (loaded) continue;
				int loadResult = CheckHandleASyncLoad(gp->bgaHandle[i]);
				if (loadResult == TRUE/*still loading*/) {
					anyLoading = true;
					continue;
				}
				if (loadResult != FALSE/*finished*/) {
					// Error
					DeleteGraph(gp->bgaHandle[i]);
					gp->bgaHandle[i] = -1;
				}
				loaded = true;
				gp->loadObject_loaded++;
			}
			if (!anyLoading) break;
			std::this_thread::sleep_for(std::chrono::milliseconds(10));
		}
```

chown2 (commented):
> Creating a few thousand loading tasks at once sounds a bit scary.

No stutters detected.
EDIT: it actually does freeze but it's pretty short, like 50-100ms. Good enough.

---

Not a fan of using an unordered_map here, but it's good enough.

LGTM after applying my suggestion! 🚀

LR2/LR2_bmsload.cpp
m8nscape: fair enough

LR2/LR2_bmsload.cpp
```diff
@@ -721,35 +721,80 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
...
+					// Error
+					DeleteGraph(gp->bgaHandle[i]);
+					gp->bgaHandle[i] = -1;
+				}
+				loaded = true;
+				gp->loadObject_loaded++;
+			}
+			if (!anyLoading) break;
+			std::this_thread::sleep_for(std::chrono::milliseconds(10));
+		}
```
chown2, line 790: My previous diff was missing a closing bracket, oops
```suggestion
		}
	}
```

m8nscape: 😂

chown2 (approved):

PR #264, opened 2026-07-28
fix: use posix_spawnp

zkldi:
Discussed on discord. I'm not a C++ programmer so the `std::thread` stuff here might be wrong. I looked up how to detach the spawn without leaving it a a zombie.

I haven't tested this code. I'm decently sure it works.

### DIFF
--- a/LR2/LR2_ir.cpp
+++ b/LR2/LR2_ir.cpp
@@ -1,6 +1,7 @@
 ﻿// LR2IR integration is deprecated. Please help us improve CustomIR instead.
 
 #include <string>
+#include <thread>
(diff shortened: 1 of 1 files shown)


zkldi:
Looks like zed autoformatting hit the file too, I can undo that if you want.

chown2:
Thanks for reporting this! I have actually been working on a fix of my own, where I also added URL protocol validation. Please see https://github.com/GOMazk/OpenLR2/pull/265.
Do note that the xdg-open call, while is a real issue, has no impact, since the Linux build is currently only there for development purposes.

PR #267, opened 2026-07-29
fix: ignore WM_DEVICECHANGE resync during play

### LINKED ISSUE #253: Possible lag caused by WM_DEVICECHANGE
https://youtu.be/hox9aGbIHv8

There is a lag for about one second during gameplay (at 6:02).

The link above leads to the video where the issue occurs, and below is the Log.txt file provided by the player.

Upon checking, I noticed that XInput, DirectInput, etc... is being reinitialized at the exact moment the lag occurred (while playing Blue Shock), which seems quite suspicious.

[Log.txt](https://github.com/user-attachments/files/30225019/Log.txt)

luxurya111:
Auto ReSetupJoypad on device-node changes caused mid-play hitches. Disable that path and resync only when entering song select.

Due to ongoing stuttering issues causing persistent inconvenience for players, I have submitted a PR.

This PR is currently under review and has not been sufficiently tested yet. Furthermore, since it remains unconfirmed whether detecting devices only upon entering Scene02 is the correct approach, I am keeping it in Draft status.

Fix #253

### DIFF
--- a/dep/DxLib/Windows/DxInputWin.cpp
+++ b/dep/DxLib/Windows/DxInputWin.cpp
@@ -1768,6 +1768,7 @@ extern int InitializeInputSystem_PF_Timing0( void )
 	int    KeyExclusiveCooperativeLevelFlag				= InputSysData.PF.KeyExclusiveCooperativeLevelFlag ;
 	int    KeyToJoypadInputInitializeFlag				= InputSysData.KeyToJoypadInputInitializeFlag ;
 	int    NoUseVibrationFlag							= InputSysData.NoUseVibrationFlag ;
+	int    NoUseDeviceChangeJoypadResyncFlag			= InputSysData.NoUseDeviceChangeJoypadResyncFlag ;
 	int    EnablePadDefaultDeadZone						= InputSysData.EnablePadDefaultDeadZone ;
 	DWORD  PadDefaultDeadZone							= InputSysData.PadDefaultDeadZone ;
 	double PadDefaultDeadZoneD							= InputSysData.PadDefaultDeadZoneD ;
@@ -1782,6 +1783,7 @@ extern int InitializeInputSystem_PF_Timing0( void )
 	InputSysData.PF.KeyExclusiveCooperativeLevelFlag	= KeyExclusiveCooperativeLevelFlag ;
 	InputSysData.KeyToJoypadInputInitializeFlag			= KeyToJoypadInputInitializeFlag ;
 	InputSysData.NoUseVibrationFlag						= NoUseVibrationFlag ;
+	InputSysData.NoUseDeviceChangeJoypadResyncFlag		= NoUseDeviceChangeJoypadResyncFlag ;
 	InputSysData.EnablePadDefaultDeadZone				= EnablePadDefaultDeadZone ;
 	InputSysData.PadDefaultDeadZone						= PadDefaultDeadZone ;
 	InputSysData.PadDefaultDeadZoneD					= PadDefaultDeadZoneD ;
@@ -3533,6 +3535,25 @@ extern int NS_SetUseXInputFlag(	int Flag )
 	InputSysData.PF.NoUseXInputFlag = !Flag ;
 
 	// 終了
+	return 0 ;
+}
+
+// Whether to resync joypads on WM_DEVICECHANGE ( TRUE: do it (default)  FALSE: skip )
+// DEVICECHANGE while FALSE is pending and flushed when set back to TRUE
+extern int NS_SetUseJoypadDeviceChangeResyncFlag( int Flag )
+{
+	InputSysData.NoUseDeviceChangeJoypadResyncFlag = !Flag ;
+
+	if( Flag && WinData.PendingJoypadDeviceChangeResyncFlag )
+	{
+		WinData.PendingJoypadDeviceChangeResyncFlag = FALSE ;
+		if( WinData.QuitMessageFlag == FALSE )
+		{
+			NS_ReSetupJoypad() ;
+		}
+	}
+
+	// Finished
 	return 0 ;
 }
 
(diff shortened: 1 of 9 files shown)


dep/DxLib/Windows/DxSystemWin.cpp
```diff
@@ -826,14 +826,12 @@ extern int NS_ProcessMessage( void )
 #endif // DX_NON_SOUND
 
 #ifndef DX_NON_INPUT
-	// WM_DEVICECHANGE メッセージが来ていたらパッドの再セットアップを行う
+	// WM_DEVICECHANGE ではジョイパッドを自動再セットアップしない。
+	// (DBT_DEVNODES_CHANGED はパッド以外のデバイスでも飛ぶため、プレイ中のフリーズの原因になる)
+	// OpenLR2 側で Scene02(選曲) 進入時に ReSetupJoypad() を明示呼び出しする。
```
MatVeiQaaa: can we please have english comments for any new code introduced in OpenLR2 repo lol

dep/DxLib/Windows/DxSystemWin.cpp
```diff
@@ -826,14 +826,12 @@ extern int NS_ProcessMessage( void )
...
 	if( WinData.RecvWM_DEVICECHANGEFlag )
 	{
 		WinData.RecvWM_DEVICECHANGEFlag = FALSE ;
-		if( WinData.QuitMessageFlag == FALSE )
-		{
-			NS_ReSetupJoypad() ;
```
MatVeiQaaa, line 835: do not remove this code entirely, you can introduce a new flag and a method to control it, so we can actually *disable* input resetup during play.

MatVeiQaaa:
the fact the devices are only ever resynced on select entry is very much not ideal. So, please make a new flag so the execution of that method can be conditional within DxLib, that way we can have a proper solution to the problem without introducing new drawbacks.

dep/DxLib/Windows/DxSystemWin.cpp
luxurya111: lmao I translated it into Japanese because the original code comments were in Japanese.
Honestly, I would like to translate all comments into English later on, but for now, I will go ahead and make the corrections. Thank you.

luxurya111:
@MatVeiQaaa Yes, I've thought about this a lot as well, and I agree that managing it with a separate flag seems like the best approach.
Since I am not personally experiencing this issue right now, I cannot confirm with certainty whether the lag is completely resolved. I will have a friend who experiences the stuttering test it out, and once that is complete, I will mark the draft PR as ready for review.

dep/DxLib/Windows/DxSystemWin.cpp
MatVeiQaaa: > lmao I translated it into Japanese because the original code comments were in Japanese. Honestly, I would like to translate all comments into English later on, but for now, I will go ahead and make the corrections. Thank you.

hmm it would be good, but you need to make sure to make it a single commit if you do, in case we will have to rebase against next DxLib update.

dep/DxLib/Windows/DxInputWin.cpp
```diff
@@ -3536,6 +3538,25 @@ extern int NS_SetUseXInputFlag(	int Flag )
...
 
+// Whether to resync joypads on WM_DEVICECHANGE ( TRUE: do it (default)  FALSE: skip )
+// DEVICECHANGE while FALSE is pending and flushed when set back to TRUE
+extern int NS_SetUseJoypadDeviceChangeResyncFlag( int Flag )
+{
+	InputSysData.NoUseDeviceChangeJoypadResyncFlag = !Flag ;
+
+	if( Flag && WinData.PendingJoypadDeviceChangeResyncFlag )
+	{
+		WinData.PendingJoypadDeviceChangeResyncFlag = FALSE ;
```
MatVeiQaaa, line 3549: ooh very nice, i thought that this would be required ideally, but i doubt i myself would bother implementing it lol! Good job.

MatVeiQaaa (approved):

MatVeiQaaa:
this looks perfect now.

chown2 (approved):

chown2:
Can you please apply this diff so that the project can still build on Linux
```diff
--- a/LR2/En_dxlibstub.h
+++ b/LR2/En_dxlibstub.h
@@ -166,4 +166,9 @@ inline int GetUseDirect3DVersion() { return 0; }
 inline int GetWindowActiveFlag() { return 1; }
 // we are windowed
 inline int GetWindowModeFlag(void) { return 1; }
+
+inline int SetUseJoypadDeviceChangeResyncFlag(int Flag) {
+  ErrorLogFmtAdd("STUB!!! SetUseJoypadDeviceChangeResyncFlag(Flag=%d)", Flag);
+  return -1;
+}
 #endif // _WIN32
```

chown2:
I can still see a little hiccup when replugging a game controller, and the controller works when hot plugged, even though SetupJoypad never ends up being called.
This isn't necessarily a bad indicator, as the lag in #253 according to the log was inside SetupJoypad.

luxurya111:
Another user who was experiencing stuttering tested the updated version, and they reported that the issue no longer occurs.
The user previously couldn't use OpenLR2 because stuttering would occur right away when playing their very first song, but it seems to be resolved now.
I will set the PR status to Ready. Thank you

MatVeiQaaa:
this probably means the game just lags for them always, only outside of play now. I wish there was a way to figure out why does WM_DEVICECHANGE occur for them all the time, because it really shouldn't.

MatVeiQaaa:
the flag specifically requires the param to be DBT_DEVNODES_CHANGED, which indicates that a device was added/removed, and nothing else. Probably, something is wrong with their system, or with one of devices connected.

luxurya111:
As a side note, someone who was experiencing lag on the song selection screen as well tested this, and the log pattern looked similar (even more severe than the one reported in the issue—it occurred about 94 times in a single minute).
So I asked them to unplug all USB devices and plug them back in, and surprisingly, the error was resolved. (Of course, it's only been a few hours, so there is a possibility it might occur again later.)
Could this be related somehow?
My guess is that since the user who tested this mentioned it didn't happen in the original LR2, the updated version of DxLib might have altered the relevant device-handling behavior.

PR #271, opened 2026-07-31
fix(bmsload): pump ProcessMessage while waiting for async BGA

luxurya111:
Skin select preview calls LoadBmsResource on the main thread; without pumping ProcessMessage the async BGA wait never completes and hangs.

Issue Log [Log.txt](https://github.com/user-attachments/files/30598661/Log.txt)

### DIFF
--- a/LR2/LR2_bmsload.cpp
+++ b/LR2/LR2_bmsload.cpp
@@ -325,14 +325,6 @@ int ReleaseBGA(game *g){
 	return 1;
 }
 
-void ProcLoadBmsResource(game *g) {
-	g->gameplay.bmsResourceLoaded = 0;
-	g->gameplay.flag_closingPhase = 0;
-	LoadBmsResource(&g->gameplay, g->sSelect.metaSelected.filepath, &g->audio, &g->config, &g->sSelect.metaSelected, g->skstruct.flag_BGA, g->skstruct.flag_flip, 0, false);
-	g->gameplay.bmsResourceLoaded = 1;
-}
-
-
 bool isVisibleNote(int ch){
 	if (CHANNEL_1P_NOTE_SC <= ch && ch <= CHANNEL_2P_NOTE_END) {
 		return true;
--- a/LR2/LR2_bmsload.cpp
+++ b/LR2/LR2_bmsload.cpp
@@ -328,7 +328,7 @@ int ReleaseBGA(game *g){
 void ProcLoadBmsResource(game *g) {
 	g->gameplay.bmsResourceLoaded = 0;
 	g->gameplay.flag_closingPhase = 0;
-	LoadBmsResource(&g->gameplay, g->sSelect.metaSelected.filepath, &g->audio, &g->config, &g->sSelect.metaSelected, g->skstruct.flag_BGA, g->skstruct.flag_flip, 0);
+	LoadBmsResource(&g->gameplay, g->sSelect.metaSelected.filepath, &g->audio, &g->config, &g->sSelect.metaSelected, g->skstruct.flag_BGA, g->skstruct.flag_flip, 0, false);
 	g->gameplay.bmsResourceLoaded = 1;
 }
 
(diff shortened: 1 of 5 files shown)


chown2 (changes requested):
This methods expects to be called from the main thread, so in the normal case this PR breaks what #263 fixed.
I propose adding a bool paramter to the function, something like `bool isMainThread`, and inside the function replace all `cfg->system.isablebmsthread == 0` (NOTE: 'isable' is a typo, should be 'disable') checks with `!isMainThread`.

chown2:
Do you have any steps to reproduce the issue? Skin customize works without problems for me

luxurya111:
@chown2 This was not an issue I experienced personally, but rather one I fixed after receiving a report from another user.
It was reported that the issue did not occur in v260719, but a bug appeared in v260730. By cross-referencing with Log.txt, I identified the cause as an error in the BGA handling section. Indeed, after fixing that part, the bug no longer occurred.

If this change breaks any previous fixes, I will proceed with your suggestion for now. However, I will need to have that user test it once more.

Below is the chat log with the user in question.
---
chatlog

---
User 1:
It keeps crashing when I scroll through skins on the wmix select screen.
I'm not sure why this is happening, but I'm attaching the log file just in case.
Log.txt (33KB)

LUXURY:
May I ask which version you are using?

User 1:
It's the version uploaded on July 30th.
Ah, sorry, to be precise, it's not crashing...

LUXURY:
Did the previous version work fine without crashing?

User 1:
It completely freezes, so I have to close it using Task Manager.
Yes.

LUXURY:
Which version were you using previously?

User 1:
It was also the latest version at the time.

LUXURY:
I see.
Because there was a version released on the 29th
and another one released on the 30th.

User 1:
Ah, hold on a second.
Ah, it was the 0719 version.

LUXURY:
Alright, let me check real quick.

User 1:
When I press BATTLE here, it freezes.

It doesn't happen 100% of the time.
The way to trigger it is to play any song or pause mid-game and exit, and then press it in the skin select window, which causes it to freeze.

chown2:
Reproducing:

- Enable song select song preview in the launcher
- Launch the game
- Having WMix as the select skin, search for 'imazinexo'
- Wait for preview to start
- Open wmix song select's embedded skin customize, press 'BATTLE'
- Observe freeze

TBH, we are not supposed to load BGAs for song preview, but what this PR does may be a decent work-around.

LR2/LR2_bmsload.cpp
```diff
@@ -720,13 +720,11 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
...
-	// Add async load functions from DxLib: SetUseASyncLoadFlag, CheckHandleASyncLoad
-	// We are on a non-main thread here, but graphics APIs, DirectX normally and OpenGL with dxlib-for-linux, require
-	// resources to be loaded from the main thread. 
-	// The async calls will make DxLib do it itself on the main thread.
-	// This fixes crashes in DX11 and fixes BGAs not displaying with dxlib-for-linux.
-	if (cfg->system.isablebmsthread == 0) {
+
+	// Async BGA only when not on the main thread: graphics APIs need main-thread
+	// completion via ProcessMessage (#263). Calling async+wait on the main thread
+	// deadlocks (e.g. skin select preview).
```
chown2: I think such comment would explain it best
```suggestion
	// Use DxLib's async load methods when not on the main thread.
	// Needed because graphics APIs, DirectX normally and OpenGL with dxlib-for-linux, require resources to be
	// loaded from the main thread.
	// These async methods will make DxLib delay uploading resources to the GPU until we call ProcessMessage, which
	// we do on the main thread.
	// Fixes crashes in DX11 and fixes BGAs not displaying with dxlib-for-linux.
```

chown2 (approved):
LGTM. @MatVeiQaaa pls review too

chown2:
Snuck in a little refactor to avoid conflict with my PR

PR #274, opened 2026-08-01
feat: add LIFT

luxurya111:
Demo: https://www.youtube.com/watch?v=oPo27-xxiTE

Raises the judge line and notes visually only. Judgment timing does not change. (Same role as IIDX Lift.)

### Song select
- Turn Lift ON/OFF with the lift button. (If it is not supported in the song selection skin, please change the value of lifttype in openlr2-config.xml to 1.)
- While OFF, the value (liftv) is kept in config, but no offset or lift graphic is applied.
- There is no mid-play ON/OFF toggle. (Independent of the Sudden+ display toggle.)

### Adjusting during play (liftv 0–100)
Lift can be adjusted only when Cover is inactive.
(Cover active = Sudden+ ON and cover display shown in play.)

| Input | Effect |
|------|--------|
| Hold Start or Select + scratch | ±1 (immediate, then every 100ms) |
| Mouse wheel | ±1 |

When Cover is active, Cover / HS work as before.

## SkinObject

| Kind | ID | Name | Notes |
|------|----|------|--------|
| `#SRC_BUTTON` | 402 | `lift` | ON/OFF (same 0/1 pattern as Sudden+ shutter `46`) |
| `#SRC_TEXT` | 302 | `LIFT_ONOFF` | ON/OFF label |
| `#SRC_SLIDER` | 27 / 28 | `LIFT1P` / `LIFT2P` | Lift cover. Not drawn when OFF. Set `range` to lane height (px); usually `muki=0` |
| `#SRC_NUMBER` | 421 / 422 | `LIFT_1P` / `LIFT_2P` | Value display. Not drawn when OFF. Prefer DST op 270 (while Start/Select held) |

Config / replay: `<lifttype>` and `<lift>` in `openlr2-config.xml`; replay initial opcodes `0x6c` (1P) / `0x9e` (2P).


luxurya111:
For now, I believe stabilization has progressed to some extent, but since the only skin tested was my own, the test cases are not yet sufficient. Testing with commonly used skins such as W-MIX is necessary. Therefore, this remains a draft for now.

chown2:
Is LIFT essentally HIDDEN+, like LANE COVER is SUDDEN+?

luxurya111:
@chown2 I wasn't sure what SUDDEN+ and HIDDEN+ meant, so I looked them up, and you were right.

LR2/LR2_skinobject.cpp
```diff
@@ -4047,6 +4060,14 @@ int SetObjectValue_Button(game *g, skstruct *sk, Timer *T, char flag) {
 				}
 				break;
 			}
+			case 402:
```
chown2: Could be `case 331` to match raja.
And please add `Extension: OpenLR2 lift` code comments to all new indexes (other than `case 331` if we proceed with it, this would be `Extension: raja lift`). Same with replay opcodes.
Also, I checked what other skin indexes raja has for lift, and none of them seem applicable for us.

luxurya111:
I have tested and verified it today with various lift settings. (At least regarding W-MIX, I received no specific issues. I mostly tested with my own custom skin.)

There don't seem to be any issues for now, so I will remove the Draft status.

LR2/LR2_gameloop.cpp
```diff
@@ -235,7 +279,20 @@ void ReactInput(game *g) {
 					&& (g->KeyInput.p1_buttonInput[7] == 1)) {
 					g->config.play.hiSpeed[PLAYER_1] -= g->config.play.hsmargin;
 				}
+			}
```
chown2: What does the moving of this block do?

LR2/Scene04_Play.cpp
```diff
@@ -18,6 +18,31 @@ static void MessageBoxA(const char*,const char* title,const char*desc,const char
 }
 #endif // _WIN32
 
+// Active only when song-select lift option is on. Lane height prefers skin LIFT slider
+// range (op2) so notes move 1:1 with the lift cover graphic; falls back to note DST y.
+float GetLiftOffsetY(const skstruct *sk, const CONFIG_PLAY *cfg, int player) {
```
chown2: You dereference these pointers unconditionally so they can be references
Please, default to doing that in future PRs
```suggestion
float GetLiftOffsetY(const skstruct &sk, const CONFIG_PLAY &cfg, int player) {
```

LR2/LR2_gameloop.cpp
luxurya111: I think it can be moved. Thank you for pointing that out.

LR2/LR2_gameloop.cpp
```diff
@@ -161,9 +161,63 @@ void ReactInput(game *g) {
...
 		if (g->is_starter == 0) {
+			const bool coverActiveP2 = (g->config.play.lanecover[PLAYER_2] && g->gameplay.lanecoverDisplayP2 == 1);
+			const bool startOrSelectP1 = (g->KeyInput.p1_buttonInput[12] == 2 || g->KeyInput.p1_buttonInput[13] == 2);
+			const bool startOrSelectP2 = (g->KeyInput.p2_buttonInput[12] == 2 || g->KeyInput.p2_buttonInput[13] == 2);
+
+			// Start/Select + scratch: lift ±1 (edge, then every 100ms)
+			static double liftScratchTickP1 = 0.0;
+			static double liftScratchTickP2 = 0.0;
+			const double now = GetTimeWrap();
+			auto stepLiftFromScratch = [&](const unsigned char *btn, int *liftv, double *tick) {
```
chown2: liftv and tick can be references

LR2/Main.cpp
```diff
@@ -1881,36 +1881,63 @@ int main(int argc, char** argv) {
...
-					if ((gs.skstruct.adjust.note_x[PLAYER_1] || gs.skstruct.adjust.note_y[PLAYER_1] || gs.skstruct.adjust.note_x[PLAYER_2] || gs.skstruct.adjust.note_y[PLAYER_2]) && gs.procSelecter == 4) {
+					const bool hasLift = ((gs.config.play.lift[PLAYER_1] && gs.config.play.liftv[PLAYER_1])
+						|| (gs.config.play.lift[PLAYER_2] && gs.config.play.liftv[PLAYER_2]));
+					if ((gs.skstruct.adjust.note_x[PLAYER_1] || gs.skstruct.adjust.note_y[PLAYER_1] || gs.skstruct.adjust.note_x[PLAYER_2] || gs.skstruct.adjust.note_y[PLAYER_2] || hasLift) && gs.procSelecter == 4) {
 						int t = gs.skstruct.image.dst[i].timer;
 						//refactored
 						if ((50 <= t && t < 60) || (70 <= t && t < 80) || t == 48) {
 							objx = gs.skstruct.adjust.note_x[PLAYER_1];
-							objy = gs.skstruct.adjust.note_y[PLAYER_1];
+							objy = gs.skstruct.adjust.note_y[PLAYER_1] + (int)GetLiftOffsetY(&gs.skstruct, &gs.config.play, PLAYER_1);
```
chown2: No need to do anything here, we can mass fix this later with clang-tidy, but in future PRs use static_cast instead of C-style casts. C-style casts can fall back to const_cast and reinterpret_cast, it's never what you want.

chown2 (commented):
I didn't look deep into it, but it's fine by me. For a skin without lift support, I didn't notice any immediate regressions.

MatVeiQaaa:
apparently lift was already a feature in LR2OOL which defined num 416 and 417 (417 for decimal, why tf does lift need a decimal number?...), but only for P1, because, as always, P2 does not exist. I'm not sure if this is worth considering, as that implementation was clearly flawed.

luxurya111:
@MatVeiQaaa 
Although I haven't checked LR2OOL directly myself, from my perspective, I don't think there's much need to change it significantly at this point. Unless someone is truly sensitive to timing, decimal places probably aren't necessary.

LR2/LR2_gameloop.cpp
```diff
@@ -309,7 +369,24 @@ void ReactInput(game *g) {
 	}
 
 	if(g->procSelecter==4){
-		if ((g->KeyInput.p1_buttonInput[12] == 1 || g->KeyInput.p1_buttonInput[13] == 1) || ((g->KeyInput.p2_buttonInput[12] == 1 || g->KeyInput.p2_buttonInput[13] == 1) && g->config.play.battle != OPTION_BATTLE_BATTLE)) {
+		// Cover display: Start or Select edge only (not both)
+		const bool p1StartEdge = (g->KeyInput.p1_buttonInput[12] == 1);
```
chown2: You don't have to act on this, it's good enough for me, but if you want to make it even better:
All of these bools, other than, p1SingleEdge and p2SingleEdge, look like an anti-pattern.
Their usage could be replaced with named constants for the button input array and it's value.
It could something like this and there will be much less logical lines of code:
```cpp
enum {
 BUTTON_START = 12,
 BUTTON_SELECT = 13,
};
enum {
 INPUT_INACTIVE = 0,
 INPUT_PRESS = 1,
 INPUT_HOLD = 2,
 INPUT_RELEASE = 3,
};

auto& p1 = g->KeyInput.p1_buttonInput;
const bool p1SingleEdge = (p1[BUTTON_START] == INPUT_PRESS && !p1[BUTTON_SELECT] == INPUT_PRESS && !p1[BUTTON_SELECT] == INPUT_HOLD)
 || (...);

auto& p2 = ...;
... // same thing
```
And obviously make the first three lines of the `if (g->is_starter == 0) {` hunk higher the same way.

chown2:
Would be nice to have info about this in `new_command(skin).md`. It duplicates the `Extension` comments in the code, but it makes sense for the end users, no skin maker is going to go through code to find the skinning indices.

chown2:
@GOMazk YOLO merge? Squash?

GOMazk:
> @GOMazk YOLO merge? Squash?

done?

GOMazk:
there are many refactoring commits, I merge it

chown2:
> there are many refactoring commits, I merge it

They fix things introduced in this PR to begin with, I think there is very little value in keeping such intermediary commits, while they inconveniece bisecting, blaming, and even simply observing the change history. 

---

@luxurya111 how about using 'fixup' commits for "fixing" changes of previous commits?
That way you avoid having to explain intermediary changes, and it makes it clear you intend to keep the git history clean.
This article explains the workflow decently well https://dev.to/koffeinfrei/the-git-fixup-workflow-386d.

And because your PRs usually focus on a specific topic, you may not even need to do the `git rebase --autosquash` step, you can just keep adding fixup commits on top. 
But not if 'main' gets updates in the mean time and there appears a conflict with the PR, please don't add merge commits in your PRs, do an actual rebase then.

luxurya111:
@chown2 I will check the next PR. Thank you

PR #276, opened 2026-08-01
perf(play): don't delay loading BMS resources on #LOADSTART

chown2:
Please, see commit messages for details.

### DIFF
--- a/LR2/LR2_bmsload.cpp
+++ b/LR2/LR2_bmsload.cpp
@@ -771,6 +771,19 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
 			}
 		}
 		while (true) {
+			if (gp->flag_closingPhase) {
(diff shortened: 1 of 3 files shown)


LR2/LR2_bmsload.cpp
```diff
@@ -771,6 +771,19 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
 			}
 		}
 		while (true) {
+			if (gp->flag_closingPhase) {
+				for (auto& [i, _loaded] : bgaHandleLoaded) {
+					// NOTE: SetASyncLoadFinishDeleteFlag is the same as DeleteGraph for already loaded images
+					SetASyncLoadFinishDeleteFlag(gp->bgaHandle[i]);
```
chown2, line 777: It's still not perfect, but this seems to be as good as we can reach with DxLib.
`on the sequence #3` is a good song for testing, you can even see some stutters while the BGA is loading (might be a DxLib bug).

PR #289, opened 2026-08-15
Add opcode(274, 275) to check lanecover status

MellDa1024:
This PR adds opcodes 274 and 275 to check the current status of Lanecover (Sudden).
implemented while adding the Lift feature to skins.

### Background / Motivation
1. To adjust the Lift, lanecover must be turned off.
2. However, when pressing `START/SELECT` to adjust Lift while lanecover is off, the lanecover value is still rendered on screen (if skin renders lanecover value, [video](https://youtu.be/1Rdtvlr7YlA?t=11)).
3. To prevent rendering the lanecover value when lanecover is inactive, an opcode was needed to query whether lanecover is currently enabled. Since I couldn't find an existing opcode for this, I implemented opcodes 274 and 275.
4. Separating opcode 270 into two distinct opcodes for lanecover and Lift could be another option, but I chose this method to offer better compatibility for skin creators who may want to display both values simultaneously.

image

Opcodes 274(p1) and 275(p2) were selected based on the unofficial LR2 skin documentation, where the Lunatic Vibes project uses these opcode numbers for similar functionality.
However, since I haven't inspected the actual source code, I cannot guarantee that the internal implementation is identical.

Because of this, I'm still considering whether it would be better to assign different opcode numbers, and how to properly write the extension comments.

### DIFF
--- a/LR2/LR2_skinobject.cpp
+++ b/LR2/LR2_skinobject.cpp
@@ -943,6 +943,17 @@ bool GetOptionFlag_dst(game *gs, int option) {
 			}
 			break;
 
+		case 274: // Extension: is lanecoverP1 enabled
+			if (gs->procSelecter == 4) {
+				if (gs->gameplay.lanecoverDisplayP1) return ret;
+			}
+			break;
+		case 275: // Extension: is lanecoverP2 enabled
+			if (gs->procSelecter == 4) {
+				if (gs->gameplay.lanecoverDisplayP2) return ret;
+			}
+			break;
+
 		case 280:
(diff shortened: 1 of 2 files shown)


chown2 (approved):
LGTM!
Would be nice to mention this in `new_command(skin).md` too. Please, squash the commmits if you decide to do it.
BTW, what's in the PR description should have ideally been in the commit description.

chown2 (approved):

MellDa1024:
I don't have much experience with PRs, so I'm not sure if I did everything correctly, but I added the related content to new_command(skin).md and included the PR description in the commit message, like you mentioned (with some help from an LLM).

MellDa1024:
There was a small typo fix in the docs. (P1/P2 -> 1P/2P)

chown2:
No worries, it's perfect!

Issue #219, opened 2026-07-09
G-Battle replays don't save correct play options (gauge, random option, rseed)


hirosiHiLowPsi:
Sorry for opening an issue.

If possible, I would like to report an issue where replays can become corrupted after borrowing a chart from the leaderboard and playing it.

This is not urgent at all, so please feel free to treat it as low priority.

chown2:
Please give out exact steps to reproduce this and see the 'corrupted' outcome.
Is the leaderboard from LR2IR or some CustomIR?
"Borrowing" - does this mean G-Battle?
Replay - the actual replay or ghost?
Corrupted - corrupted how exactly? Incorrect random layout?

hirosiHiLowPsi:
Yes, that is exactly right.

When I cleared the chart using G-Battle on Stellaverse IR, the random pattern played back in the replay was significantly different from the one I actually played.

chown2:
Some of the places to fix:
https://github.com/GOMazk/OpenLR2/blob/56f700aaa0f6ea90f9615e68722f1569a7bb373c/LR2/Main.cpp#L1152
https://github.com/GOMazk/OpenLR2/blob/56f700aaa0f6ea90f9615e68722f1569a7bb373c/LR2/Main.cpp#L1170
https://github.com/GOMazk/OpenLR2/blob/56f700aaa0f6ea90f9615e68722f1569a7bb373c/LR2/Scene04_Play.cpp#L1817
https://github.com/GOMazk/OpenLR2/blob/56f700aaa0f6ea90f9615e68722f1569a7bb373c/LR2/LR2_bmsload.cpp#L2063

chown2:
This is a known bug in LR2
@hirosiHiLowPsi If you really want to restore the replay file, there is a tutorial to fix them manually https://www.youtube.com/watch?v=Qj1W_-3llXs

chown2:
This is probably fixed in v260825 by #290.

PR #290, opened 2026-08-16
fix: correct G-Battle replay recording and playback

Skalim112:
## Summary

Related to #219 

G-Battle replays are intended to capture the current player's gameplay as a solo play, using the target chart's seed.

## Known problems

Previously, the replay could store an incorrect seed (the known one), ghost input, and `G-BATTLE` in its header (shown on the video below), causing incorrect notes during playback or the gameplay to initialize twice.

Video showing the actual problem (video is too big): https://gofile.io/d/DCED359x

## What does this fix

- Preserves the generated RANDSEED in recorded replays.
- Records only Player 1 input and judgments during G-Battle.
- Saves Player 1's gauge and random options in the replay header.
- Stores `BATTLE_OFF` so new G-Battle replays are played back as solo play.

Tested with regular Battle, G-Battle, retries, and different gauge settings.

Existing replays are still compatible.

### DIFF
--- a/LR2/structure.h
+++ b/LR2/structure.h
@@ -1477,6 +1477,8 @@ struct gameplay {
 	int bpmChangedBmstime; /* bpm change timing */
 	char ghostBattle;
 	std::optional<CONFIG_PLAY> playConfigBackupBeforeTargetSomething;
+	// Play config including temporary options such as those for g-battle
+	std::optional<CONFIG_PLAY> actualPlayConfigCopyForResultIr;
 	int delayDetectedCount;
 	int delayCheckCount;
 	char isCourse; 
--- a/LR2/LR2_bmsload.cpp
+++ b/LR2/LR2_bmsload.cpp
@@ -2119,7 +2119,6 @@ int ParseBmsFile(gameplay *gp, CSTR filename, AUDIO *aud, ConfigStruct* cfg, BMS
 
 	oldSpeedMultiplier = gp->freqSpeedMultiplier;
 
-	//TOFIX : seed is not putted into replaydata, when use ghostbattle. (retry puts seed) (see also ProcS_Play())
 	//TOFIX: 0 is a valid seed. Note that LR2IR also returned randomseed=0 on missing ghosts.
 	if (gp->randomseed != 0) {
 		ErrorLogFmtAdd("RANDSEEDを引き継ぎます\n");
@@ -2140,9 +2139,9 @@ int ParseBmsFile(gameplay *gp, CSTR filename, AUDIO *aud, ConfigStruct* cfg, BMS
 		if (gp->randomseed == 0xFFFF) {
 			gp->randomseed = GetRand(0x7ffe);
 		}
-		if (gp->replay.status == 1) {
-			AddReplayData(&gp->replay, 0, 200, (short)gp->randomseed);
-		}
+	}
+	if (gp->replay.status == 1) {
+		AddReplayData(&gp->replay, 0, 200, static_cast<short>(gp->randomseed));
 	}
 	ErrorLogFmtAdd("RANDOMSEEDは%dです。\n", gp->randomseed);
 	SRand(gp->randomseed);
--- a/LR2/LR2_customir.cpp
+++ b/LR2/LR2_customir.cpp
@@ -526,7 +526,8 @@ IRScoreInternal::IRScoreInternal(game& game, sqlite3* sql, int _player, std::str
 		song.courseType = curSong.courseType;
 		songPlayLevel = curSong.level;
 	}
-	CONFIG_PLAY& cfg = game.config.play;
+	gameplay& gameplay = game.gameplay;
+	const CONFIG_PLAY& cfg = gameplay.actualPlayConfigCopyForResultIr ? *gameplay.actualPlayConfigCopyForResultIr : game.config.play;
 	settings.gaugeType = cfg.gaugeType[_player];
 	settings.random[PLAYER_1] = cfg.random[PLAYER_1];
 	settings.random[PLAYER_2] = cfg.random[PLAYER_2];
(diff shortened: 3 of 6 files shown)


chown2:
Everything but `fix(g-battle): save g-battle replays as solo play` makes sense for me.
1) as a player, I see no reason not to expect battle replays to just work
2) 2P inputs use separate op codes, they shouldn't be causing any issues, so if there are any, we should figure out where they actually come from
3) figuring out 'the issues' will also help with watching old replays, since the commit only affects new replays

Skalim112:
> Everything but `fix(g-battle): save g-battle replays as solo play` makes sense for me.
> 
>     1. as a player, I see no reason not to expect battle replays to just work
> 
>     2. 2P inputs use separate op codes, they shouldn't be causing any issues, so if there are any, we should figure out where they actually come from
> 
>     3. figuring out 'the issues' will also help with watching old replays, since the commit only affects new replays

1. So, it is not a priority for the moment?
2. At the moment there´s no utility on saving p2 options, i saw no problems on keeping it out of the register or making a rollback of it and save the p2 options anyway (at the moment, its saving something called "judgment" which is the sum of P-GREATS, GREAT, GOOD, BAD and POOR on both p1 and p2, and it only happens on g-battle on the first play, resetting it doesn´t save it)
3. The real issue about corrupted replays is that there´s not a way to save the seed of the replay other than check if THAT REPLAY was your best score, in that case, the program should assume that the seed of that replay is the one in score.db, which is not always true.
imagen
and, based on this point, the replay file saves a seed managed by your options, not by the seed´s rival.

I made a tool for this purpose btw:

## new .exe
- This is the seed i have registered on my score.db
imagen
- This is the seed registered on my last play with g-battle and no restarting it with the solution
imagen

## old .exe
- This is the replay made with the ...19 compiled solution, it doesn´t save the seed at all
imagen
- This is the replay made on the same exe, with a restart, it captures the seed, but g-battle is ON, which makes the game gone crazy
imagen
imagen

And the seed on my score.db was not altered at all.

So, if someone has a new CLEAR, but with worse score, the seed is not going to be saved on score.db so, there´s not a known way to "handle all the cases", only the ones who has registered as "battle = g-battle", because this is the one that makes the replay to start two times and blackscreens it.

The only way to handle this particular issue (and still not find a way to fix the older replays) is to save the seed when your score gets updated in every way (again, it only updated when your pb on exscore).

So, do i rollback the p2 saving or i keep the solution for the moment?

Skalim112:
Rollbacked the player 2 saving events, tested the new resultant replays, 0 issues at the moment

chown2:
> Rollbacked the player 2 saving events, tested the new resultant replays, 0 issues at the moment

Please just use interactive rebase to remove the commit you reverted, keep the history clean. That would also get rid of the unnecessary merge commit. Most open source projects that don't merge with squashing prefer that too. I have written a little guide, but lost it by accidentally closing the browser tab, sorry.

---

If 'g-battle' battle mode causes issues, how about treating it as just 'battle'? To implement it most easily that would mean replacing OPTION_BATTLE_GBATTLE value to OPTION_BATTLE_BATTLE when interpreting 0xc9 op code. Ignoring the rseed issue, that would also help with some LR2 replays.

---

> I made a tool for this purpose btw

Very cool :D

Skalim112:
> Please just use interactive rebase to remove the commit you reverted, keep the history clean. That would also get rid of the unnecessary merge commit. Most open source projects that don't merge with squashing prefer that too. I have written a little guide, but lost it by accidentally closing the browser tab, sorry.
> 
My bad, i will take it into consideration for the next one.
> If 'g-battle' battle mode causes issues, how about treating it as just 'battle'? To implement it most easily that would mean replacing OPTION_BATTLE_GBATTLE value to OPTION_BATTLE_BATTLE when interpreting 0xc9 op code. Ignoring the rseed issue, that would also help with some LR2 replays.
> 
Turns out i made a way to edit that byte on the tool so, this makes the testing more convenient
imagen
LR2 2026-08-18 21-00-18

so, i modified the header for Battle and tested it, it seems like LR2 forces the replay to initiate on g-battle, and at the same time, both sides have errors on seed

The video: https://gofile.io/d/tbYaazmT

The replay if you want to see (the lost delaycated (distant memory) dl3: [b457288fb855e43b2e5f8a9b5d0df1b0.zip](https://github.com/user-attachments/files/31203242/b457288fb855e43b2e5f8a9b5d0df1b0.zip)

Skalim112:
Rebase done, is this what was supposed to be done?

Skalim112:
Along with that i found a bug related to replay submission (This was happening on older builds too)

- When g-battle = ON on non-ran option, and if you PB, the game submits your actual settings to the IR, saving the seed, but making it non-ran so, when you g-battle again vs your pb, you end playing on an overwrited non-ran setting, when you are supposed to play on RANDOM and with the saved seed.

This behavior is fixed on the last commit

LR2/structure.h
```diff
@@ -1477,6 +1477,7 @@ struct gameplay {
 	int bpmChangedBmstime; /* bpm change timing */
 	char ghostBattle;
 	std::optional<CONFIG_PLAY> playConfigBackupBeforeTargetSomething;
+	std::optional<CONFIG_PLAY> playConfigForResultIr;
```
```suggestion
	// Play config including temporary options such as those for g-battle
	std::optional<CONFIG_PLAY> actualPlayConfigCopyForResultIr;
```

chown2:
> Rebase done, is this what was supposed to be done?

Yes! Please use rebase when applying the suggestion too.
You can use `git commit --fixup TARGET_HASH` to create a temporary commit, and then automatically merge it with `git rebase --autosquash --interactive main`.
EDIT: or, well, just use git commit --amend, it's already the last commit.

---

> When g-battle = ON on non-ran option, and if you PB, the game submits your actual settings to the IR

This must be happening since 38f59d5f9b46fb1da233f8130191bc2863540965.
Having to make a config copy is unfortunate but I guess it makes sense as an easier solution.
Other solution is splitting BeginResultIr/ResultIrAsync into two separate things, one for actual score submission, one for leaderboard fetching, with the goal of moving the score submission to be near SaveResult again.

chown2 (commented):
Code changes make sense but I didn't test anything.

LR2/structure.h
Skalim112: Applied with --amend

GOMazk:
I did't test also but looks good for me. merging it

Issue #102, opened 2026-06-06
Restarting from the result screen by pressing the white+blue keys gives a new random


FSRS:
In classic LR2, using the restart feature from the result screen (white+blue keys) preserves the same random, including any #RANDOM processing within the chart.

However, in the current openLR2 implementation, restarting gives a new random, including reprocessing #RANDOM in the chart.

A good reference case is: Set a HS-FIX in MIN for a dramatic change.

Body: https://fether.exblog.jp/#BMS_fether
Chart: https://darksabun.club/event/renewal/sabun/Append_19_twinklesky_DS.zip

MatVeiQaaa:
this is a feature. white key + 2 restarts with new random now. white key + 4/6 restarts with same random.

PR #292, opened 2026-08-18
feat(config): add enable_newrandomrestart to Result retry

luxurya111:
related: https://github.com/GOMazk/OpenLR2/issues/102

This feature will be added to the config, allowing users to optionally enable or disable the function that starts with a different random selection using the 2 key.

### DIFF
--- a/LR2/LR2_configsave.cpp
+++ b/LR2/LR2_configsave.cpp
@@ -579,6 +579,7 @@ int WriteOpenLr2ConfigXml(game *g, const char *filename){
 
 	fputs("\t<play>\n", pFile);
 	WriteXML_Tab2BoolAsInt(pFile, "gaugeautoshift", (g->config).play.m_gas);
+	WriteXML_Tab2BoolAsInt(pFile, "enable_newrandomrestart", (g->config).play.enableNewRandomRestart);
 	WriteXML_Tab2Int(pFile, "lifttype", (g->config).play.lift[PLAYER_1]);
 	WriteXML_Tab2Int(pFile, "lift", (g->config).play.liftv[PLAYER_1]);
 	fputs("\t</play>\n", pFile);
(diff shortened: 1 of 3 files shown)


chown2 (commented):
Personally, I'm against knobs for every little thing. At worst, you just had to hold a different set of keys. Do we really need an option for that?
Code-wise LGTM, though 'enable_newrandomrestart' should probably just be 'newrandomrestart'.

MatVeiQaaa:
+1 for that this toggle doesn't need to and even shouldn't exist. @GOMazk what do you think?

luxurya111:
@chown2 @MatVeiQaaa The reason I wanted to add this feature in the first place is that I’ve occasionally seen people like me who don’t want the 2 key to switch to a different random selection. Since people who use OpenLR2 are, after all, people who have used LR2 and are moving over to it, I think this could help address some of the inconveniences that existing users have experienced.

The toggle button is merely one suggestion, so it doesn’t necessarily have to be provided as a toggle. I would just like to see it offered as some kind of option, even if it’s implemented differently.

Especially for people who have been using it for several years, I think changing an established habit would be quite difficult.. (not for me)

MatVeiQaaa:
@luxurya111 and yet OpenLR2 is not exactly LR2 and some things are inevitable to be different. Why would anyone use OpenLR2 if they want it to be exactly like LR2? Maybe we should make one global toggle to disable every slight difference it created at once? It's incredibly silly to say that it's difficult for someone to realize holding one specific button will trigger some feature, so we must have an *option* to remove that feature, already confirming that this feature is useful by not considering removing it completely. I'm certain even many people who find it "uncomfortable" would end up using it if they didn't retaliate immediately against unfamiliar, and made the minuscule effort to give it a thought. The only justification for such toggle i can see is "why not", i personally can't get behind this.

GOMazk:
> +1 for that this toggle doesn't need to and even shouldn't exist. @GOMazk what do you think?

well, it would sound like off-topic, I think 2+white button as a restart with new randomseed itself not so good. imo start+select on result screen would be good because it's the same in play.

MatVeiQaaa:
> imo start+select on result screen would be good because it's the same in play.

1. It's different in play because you can't have it same way as in select, as it would interfere with playing
2. Why would we have new- and same- random restart in two completely different ways on the same scene
3. It's specifically white+2 because it's on the left side, like it is on the left side in play (start)
4. If start+select is occupied in result scene there will never be any binds that could make sense to open additional menu. Sub-menu is the only way I see to potentially expand the UI.

GOMazk:
> 3. It's specifically white+2 because it's on the left side, like it is on the left side in play (start)
> 4. If start+select is occupied in result scene there will never be any binds that could make sense to open additional menu. Sub-menu is the only way I see to potentially expand the UI.

these make sense, thank you.

I'm the one who thinks "why not". I think they wouldn't want to use different random restart at all.

MatVeiQaaa:
> I'm the one who thinks "why not". I think they wouldn't want to use different random restart at all.

Then merge it - you are the boss.

luxurya111:
@MatVeiQaaa Sorry for the slightly delayed response. From what I can tell, I don’t think we’ll be adding a separate config for every individual feature. The only one that affects gameplay is this particular setting, and I don’t expect any additional separate configs to be added in the future.

PR #296, opened 2026-08-25
Fix DP Fast/Slow and hit offset display

### LINKED ISSUE #250: Skin Number 211 (2P Fast/Slow) Broken
I believe the 2P Fast/Slow indicator has never worked properly in LR2, as I cannot find any evidence of any playskin that has ever had it. The relevant skin number is 211, but using it in a skin yields nothing. Instead, it seems number 210 (1P Fast/Slow) tracks both 1P and 2P judgements, though it's hard to tell exactly what's going on in live gameplay.

I can provide my skin implementation if needed.

Note: tested only in DP, not battle

yuyumina14keys:
Fixes #250

Problem

In DP, OpenLR2 stores judgments from both play sides in PLAYER_1 extended stats. Skin numbers 210/201 therefore receive judgments from both sides, while 211/213 do not receive the correct 2P-side values.

Changes

Add separate Fast/Slow display state for each play side.
Add separate hit-offset display state for each play side.
Update the display state according to the judged lane.
Reset the added state when gameplay is initialized or Quick Restart is used.
Keep the existing empty-POOR / missed-POOR hit-offset behavior unchanged.

Tested

5KEY / 7KEY / 9KEY / 10KEY / 14KEY
SP / DP / SP-to-DP / DP-to-SP / PMS-to-SP
D-Battle / Battle
Normal notes / LN / scratch / simultaneous DP inputs
Replay
Quick Restart with random changed and preserved
Song and mode changes

1P and 2P Fast/Slow and hit-offset displays remained independent in all tested DP/Battle cases.

### DIFF
--- a/LR2/LR2_bmsload.cpp
+++ b/LR2/LR2_bmsload.cpp
@@ -454,6 +454,8 @@ int PlayerCheckAndSwap(gameplay *gp){
 int InitGameplay(gameplay *gp, CONFIG_PLAY *cfg) {
 
 	PlayerCheckAndSwap(gp);
+	gp->lastFastSlowBySide.fill(0);
+	gp->lastHitOffsetBySide.fill(0);
 	gp->bpmt_start = 1;
 	gp->isPreviewLoad = 0;
 	if (gp->bmsobj.count == 0) InitNoteBuffer(&gp->bmsobj, 1000);
@@ -852,6 +854,8 @@ int LoadBmsResource(gameplay *gp, CSTR /*BMSfilepath*/, AUDIO *aud, ConfigStruct
 int InitGameplay_retry(gameplay *gp, AUDIO *snd, game *g) {
 	
 	PlayerCheckAndSwap(gp);
+	gp->lastFastSlowBySide.fill(0);
+	gp->lastHitOffsetBySide.fill(0);
 	if (gp->replay.status != 2 && gp->replay.status == 1) {
 		AddReplayData(&gp->replay, 0, 200, gp->randomseed);
 	}
--- a/LR2/Scene04_Play.cpp
+++ b/LR2/Scene04_Play.cpp
@@ -738,6 +738,19 @@ int JudgeToScore(int judge, game *g, int player, int lane, char isReplay) {
 	return ApplyJudgeNote(judge, g, player, lane, &g->timer1, isReplay);
 }
 
+
+static void UpdateLastFastSlowBySide(game* g, int player, int lane) {
+	// In DP, both sides belong to PLAYER_1, so use the lane to determine the display side.
+	const int side = (lane >= 10) ? PLAYER_2 : PLAYER_1;
+	g->gameplay.lastFastSlowBySide[side] = g->gameplay.player[player].extendedColumnStats[lane].lastFastSlow;
+}
+
+static void UpdateLastHitOffsetBySide(game* g, int player, int lane) {
+	// In DP, both sides belong to PLAYER_1, so use the lane to determine the display side.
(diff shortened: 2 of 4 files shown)


LR2/structure.h
```diff
@@ -1503,6 +1503,8 @@ struct gameplay {
 	CSTR previewBMShash;
 	CSTR previewBMSfilepath;
 	std::mutex criticalSection;
+	std::array<int, PLAYER_MAX> lastFastSlowBySide = {};
```
MatVeiQaaa: why in gameplay? This belongs to player struct.

LR2/LR2_bmsload.cpp
```diff
@@ -454,6 +454,8 @@ int PlayerCheckAndSwap(gameplay *gp){
 int InitGameplay(gameplay *gp, CONFIG_PLAY *cfg) {
 
 	PlayerCheckAndSwap(gp);
+	gp->lastFastSlowBySide.fill(0);
```
MatVeiQaaa, line 457: you can just do gp->lastFastSlowBySide{}, although whatever, i suppose.

LR2/Scene04_Play.cpp
```diff
@@ -738,6 +738,19 @@ int JudgeToScore(int judge, game *g, int player, int lane, char isReplay) {
...
+static void UpdateLastFastSlowBySide(game* g, int player, int lane) {
+	// In DP, both sides belong to PLAYER_1, so use the lane to determine the display side.
+	const int side = (lane >= 10) ? PLAYER_2 : PLAYER_1;
+	g->gameplay.lastFastSlowBySide[side] = g->gameplay.player[player].extendedColumnStats[lane].lastFastSlow;
+}
+
+static void UpdateLastHitOffsetBySide(game* g, int player, int lane) {
+	// In DP, both sides belong to PLAYER_1, so use the lane to determine the display side.
+	const int side = (lane >= 10) ? PLAYER_2 : PLAYER_1;
+	g->gameplay.lastHitOffsetBySide[side] = g->gameplay.player[player].extendedColumnStats[lane].lastHitOffset;
```
MatVeiQaaa: since this depends on prior execution of increment_extended, it should be within that lambda. It would look better there either way, there is no need for those two methods.

LR2/Scene04_Play.cpp
```diff
@@ -738,6 +738,19 @@ int JudgeToScore(int judge, game *g, int player, int lane, char isReplay) {
 	return ApplyJudgeNote(judge, g, player, lane, &g->timer1, isReplay);
 }
 
+
+static void UpdateLastFastSlowBySide(game* g, int player, int lane) {
```
MatVeiQaaa, line 742: i don't think there needs to be two separate methods for offset and f/s, you should merge them into one.

MatVeiQaaa:
was this tested to work with g-battle?

LR2/Scene04_Play.cpp
MatVeiQaaa, line 742: also, pass `gameplay` struct by reference, instead of `game` by pointer

chown2 (commented):
Thanks for looking into this. Judging by a similar fix in #298, maybe this could be implemented without additional storage? Otherwise, I agree with mat's comments.

MatVeiQaaa:
> maybe this could be implemented without additional storage?

I thought about this too and realised this storage is necessary, as extendedStats specifically represents player 1 and 2. Here, we don't want fast/slow state be connected to player entity at all, and instead specifically mean just the side (left/right), which isn't the same thing at all.

Sorry, I'm talking outside of context you referenced, but still I believe this is relevant.

PR #299, opened 2026-08-29
"play_log" in scoreDB

GOMazk:
log every play in DB, for further use
it saves everything in existing score table, and more - time stamp, wasBest, mashpoor, fast, slow, cb, isMod, ModName, ModHash
any suggestion of columns are welcome


MatVeiQaaa:
Doesn't need playcount, clearcount and failcount. "minbp" should probably be renamed to just "bp", as it's no longer related to other plays. Same for "clear_ex" and other similar fields. What do "mod" and other related fields mean to represent? What use case do you imagine for "wasBest"?

MatVeiQaaa:
Consider the name "score_log" for the table instead. Should be able to save all replays to go along with those scores too. Maybe it even makes sense to save them as binary data in the database instead of a file.

GOMazk:
> Doesn't need playcount, clearcount and failcount.

 just copying from score, to match column numbers

>  What do "mod" and other related fields mean to represent?

preparing for score-affecting mods, this can be done with new tables but for an accident

> What use case do you imagine for "wasBest"?

didn't imagine yet, just an idea now

imo even if there are no UI for these, recording from early would be good

> Consider the name "score_log" for the table instead. Should be able to save all replays to go along with those scores too. Maybe it even makes sense to save them as binary data in the database instead of a file.

nice idea, I forgot replays

MatVeiQaaa:
>  just copying from score, to match column numbers

Why would you want to match the indices? Accessing table data by index is a poor practice and you shouldn't be accounting for it.

> preparing for score-affecting mods, this can be done with new tables but for an accident

I don't think you need a boolean to indicate mods presence, as the list not being empty would already do that. What would ModHash hash? And maybe it makes sense to have "opEx" instead then to entirely supersede existing "op_best" field and also store potential new properties. The format doesn't have to follow the idea of "op_best", it could be a named list. 

> didn't imagine yet, just an idea now

I can't see any use in this.

MatVeiQaaa:
OpenLR2 could use this as the main score data entry and have "score" table for backwards-compatibility. That way we can have full control over the score data structure.

chown2:
This could be implemented as a CustomIR module :clown_face:

GOMazk:
current status: considering use CustomIR instead score.db, I'm still reading customIR.cpp but busy

Issue #244, opened 2026-07-15
Allow F8 (folder refresh) to work in song search


AYhaz:
The hotkey used to refresh folders ingame (F8) does not do anything when browsing through search results on the song select screen. I think it would be great if the game could refresh the folder that contains the song that is currently highlighted while searching.

If this feature were to be added together with #243, it would allow for well-needed QoL regarding operations like very quickly installing sabuns while the game is opened.

MatVeiQaaa:
IIRC F8 in general doesn't do anything when a song is highlighted? And the desired behaviour is so the folder of that song is reloaded.

PR #300, opened 2026-08-30
Extend song reload to F8, autoreload=1 and metadata repair

Skalim112:
## What is this

There is a known issue provoked by tools like bemusicseeker (which is the most used tool for importing songs to the game) that makes the metadata of the charts imported display their BPM or Total Notes like NONE or 0, and we don´t actually have a way to fix this more than accessing to the song.db and edit them manually (and nobody will do that).

I decided to extend the actual autoreload=2 behavior to the other options, and improve it for finally complete the scan tandem.

Related to: #244

## Summary

- Add a reusable scoped reload for the selected song folder.
- Make F8 refresh the selected chart or folder, independently of the `autoreload` setting.
- Make `autoreload=1` use the same scoped refresh when entering a folder.
- Check charts directly inside the selected folder, plus immediate child folder content.
- Repair incomplete `song.db` metadata (`minbpm`, `maxbpm`, `karinotes`, and `adddate`) after full or targeted reloads.
- Add delayed progress feedback for targeted reloads. It appears after one second, or immediately when an update is detected so, it doesn´t dull the experience of the player.

## Behavior

- `autoreload=0`: F8 performs a manual scoped refresh; folder or chart, its your choice now.
- `autoreload=1`: entering a folder performs a scoped refresh; F8 remains available.

## Testing

Tested F8 reloads, `autoreload=1`, full reloads, metadata repair, and targeted reload progress across folders with and without updates.

## Video

I had FIREBALL with metadata error due to bemusicseeker´s way to import songs, with F8 i fixed it instantly (and reloaded the folder for check new charts)
https://github.com/user-attachments/assets/06163838-6f90-4c1a-bee2-abd5d3a753c7

## Image Summary

song table with metadata problems

how it looks in-game
LR2 2026-08-22 19-03-39

how it should look
LR2 2026-08-22 19-04-56

how it should look part 2


chown2:
`fix(song-db): repair incomplete chart metadata during reload`
I think it's improper to have work-arounds for outside bugs in the game code. BeMusicSeeker should have never touched song.db in the first place. and as a user, I would expect F8 to just reload the song data comletely, therefore there should be no special care for this case, as F8, a song refresh, should already fix the metadata, by definition.

MatVeiQaaa:
To my knowledge song import is not a widely used feature of bmseeker. Isn't there an open source repo for it though? Consider implementing a fix there. I also don't think this fix has a place in the game code.

chown2 (commented):
`feat(song-reload): extend scanning to selected songs and folders`
`feat(song-reload): use scoped reload for F8 and autoreload=1`
looks like these should have been a single commit? on it's own, the first commits adds an used function.

`fix(song-db): repair incomplete chart metadata during reload`
the whole idea makes 0 sense in my view, per the above comment.

`feat(song-reload): show progress for targeted refreshes`
please avoid any additional state, it's complicated enough as it is. the way I see it, any kind of reload should be displayed in the same way. if all of this is necessary just for some little eye candy, I think it would be better to omit it completely.

chown2:
> Isn't there an open source repo for it though?

@MatVeiQaaa there are two closed source reverse engineered forks. they are just redistributed on github, there is no source code. https://github.com/Neeted/bemusicseeker-unofficial-fork https://github.com/SayakaIsBaka/BeMusicSeeker-installer

Skalim112:
> `feat(song-reload): extend scanning to selected songs and folders` `feat(song-reload): use scoped reload for F8 and autoreload=1` looks like these should have been a single commit? on it's own, the first commits adds an used function.
> 
I will check that
> `fix(song-db): repair incomplete chart metadata during reload` the whole idea makes 0 sense in my view, per the above comment.
> 
The old LR2 was able to handle this kind of errors, that´s why i brought it back, if even with that is not contemplated on this project, i can took it out again (the feature was on autoreload=2)
> `feat(song-reload): show progress for targeted refreshes` please avoid any additional state, it's complicated enough as it is. the way I see it, any kind of reload should be displayed in the same way. if all of this is necessary just for some little eye candy, I think it would be better to omit it completely.
>
agree, the thing is that, when i load it without it, it turns to the black-screen immediately, which can be a little annoying. I will check what i can do

Skalim112:
So, i took off the metadata fix, maybe i will make a tool for incidents like that in the future, tell me if it needs something more or if there´s something pending around the same topic that can be added here.

PR #304, opened 2026-09-13
fix(course): wrong total notes on fail/retry

luxurya111:
It seems that the total combo count is not being reset when restarting a course. Since it appeared to be sent with the count accumulating, I fixed it.


chown2:
Does this by any chance fix bogus accuracy value on course result? I've never narrowed it down to be 100% reproducible, but it is something I've seen almost each time I tried courses in OpenLR2

LR2/LR2_bmsload.cpp
```diff
@@ -903,6 +903,7 @@ int InitGameplay_retry(gameplay *gp, AUDIO *snd, game *g) {
 		gp->player[p].judge_damage = tempDmg;
 		memcpy(&gp->player[p].judgetime, tempTime, sizeof(tempTime));
 		gp->player[p].totalnotes = tempCount;
```
chown2, line 906: If we do need to restore 'total_note', we better actually restore 'total_note' instead of writing 'totalnotes' into it

So first something like
```cpp
		int tempCount2 = gp->player[p].total_note;
```
Then
```cpp
		gp->player[p].total_note = tempCount2;
```

I wonder what is the difference between total_note and totalnotes?

By the way, this line wasn't in the original game, also makes me wonder, why it's needed now. You can check it with Ghidra. https://github.com/GOMazk/OpenLR2#ghidra-archive-decompiler-file

luxurya111: @chown2
totalnotes = current chart's note count
total_note = running sum across loaded course stages (`ParseBmsFile` does `total_note += totalnotes`). Course result then copies total_note into totalnotes.

OpenLR2 retry does `player = PLAYERSTATUS()`, which zeroes total_note, then only restores totalnotes / judgetime / judge_damage. That's why total_note is lost here. On this path they happen to be equal (course same-random retry is stage 1).

However, I agree—the code above explains it more clearly, so I think it would be better to make the changes according to that code. Thank you for the feedback.

luxurya111: +To be honest, I haven’t checked Ghidra yet. I only know that this code is the source of the problem, so if an actual reference to Ghidra is necessary, I can put this PR on hold.

chown2: > totalnotes = current chart's note count
> total_note = running sum across loaded course stages

Thanks! Can you add it this as comments for the fields in PLAYERSTATUS, please?

> if an actual reference to Ghidra is necessary, I can put this PR on hold

Nah, it's fine by me. Of course, it would be perfect to find out why exactly have we ended up having different behavior, but realistically it's unnecessary effort.

